PR #1, opened 2021-02-08
Set up the lambda handler for compiling and executing Java

jmkulwik:
This PR handles two things
1. It creates the initial gradle project via `gradle init` - The first option selected was `3: Library` and the rest of the options used the default.
2. It sets up the lambda handler to compile and execute student Java code.

The handler currently only accepts a single file. 

All files are auto-generated by gradle except the following:
- All of CodeExecutor.java
- The section of build.gradle starting with `// AWS Lambda dependencies`
- The last line in the .gitignore

Example lambda: https://console.aws.amazon.com/lambda/home?region=us-east-1#/functions/jessie-test-java-lambda-function?tab=configuration 

This can be run using a Java11 AWS Lambda function ([instructions](https://docs.aws.amazon.com/lambda/latest/dg/lambda-java.html)]

Once the lambda is created, follow these instructions:
- In the AWS console, change the handler to `org.code.javaide.codebuilder.CodeExecutor::handleRequest`
- In your local code branch run `./gradlew -q buildZip`
- Upload the zip file from ./java-ide/org-code-javaide-codebuilder/lib/build/distributions/lib.zip to the lambda on the AWS console
- Add a test event that maps a file name to Java code. Example:
```
{
  "Foo.java": "public class Foo {\n  public static void main(String[] args) {\n    System.out.println(\"Hello Molly!\");\n  }\n}\n"
}
```
- Run the test event
- If all goes well, you should see the output from your program


org-code-javaide-codebuilder/lib/src/main/java/org/code/javaide/codebuilder/CodeExecutor.java
```diff
@@ -0,0 +1,97 @@
...
+        e.printStackTrace();
+        // TODO: CSA-42 Improve error handling
+        return "An error occurred creating files.";
+      }
+    }
+
+    ProcessBuilder pb = new ProcessBuilder().command("bash", compileRunScript.toString());
+
+    Process process = null;
+    try {
```
molly-moen, line 68: I'm finding when I run code that should fail on compile (either a compiler error or mismatched filename and class), I get an empty string--expected behavior?
I also get an empty string if I try to throw an exception from my code.

org-code-javaide-codebuilder/lib/src/main/java/org/code/javaide/codebuilder/CodeExecutor.java
```diff
@@ -0,0 +1,97 @@
...
+      String className = FilenameUtils.removeExtension(projectFile.getKey());
+      String userCode = projectFile.getValue();
+      try {
+        // Build user's java file
+        File userFile = new File(Paths.get(tempDir, projectFile.getKey()).toString());
+        FileWriter writer = new FileWriter(userFile);
+        writer.write(userCode);
+        writer.close();
+
+        // Create the compile & run script
```
molly-moen, line 51: appreciate the comments! I think we should err on the side of over-commenting when it comes to any Java we write.

org-code-javaide-codebuilder/lib/src/main/java/org/code/javaide/codebuilder/CodeExecutor.java
```diff
@@ -0,0 +1,97 @@
...
+        scriptWriter.write("java -cp " + tempDir + " " + className);
+        scriptWriter.close();
+      } catch (IOException e) {
+        e.printStackTrace();
+        // TODO: CSA-42 Improve error handling
+        return "An error occurred creating files.";
+      }
+    }
+
+    ProcessBuilder pb = new ProcessBuilder().command("bash", compileRunScript.toString());
```
sureshc, line 65: I believe the Lambda execution environment is designed for the main “work” or functionality of the Lambda to be carried out by the RequestHandler and any code it invokes within that process. While other processes can be started within the Lambda container, the Lambda execution environment may not be able to ensure those other processes are cleaned up at the end of the RequestHandler’s execution. If those other processes hang or are still executing when the RequestHandler completes, they may persist and resume executing the next time the Lambda is invoked and the same container is re-used, causing unintended side effects. From the [docs](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-context.html):

```
Background processes or callbacks that were initiated by your Lambda function and did not complete when the function ended resume if Lambda reuses the execution environment. Make sure that any background processes or callbacks in your code are complete before the code exits.
```

It may be worth investigating whether we can use the [Java compiler Module](https://docs.oracle.com/en/java/javase/11/docs/api/java.compiler/module-summary.html) to compile the student’s source code, at which point we could possibly load/execute the student’s compiled classes within the main Lambda RequestHandler execution environment. Sand boxing the student’s Java code would be a challenge to ensure it doesn’t invoke/manipulate our own RequestHandler’s functionality.

Found some outdated examples of using `java.compiler`, but those might be useful in prototyping:

https://javabeat.net/the-java-6-0-compiler-api/
https://www.javacodegeeks.com/2015/09/java-compiler-api.html

org-code-codebuilderapp/.gitignore
```diff
@@ -0,0 +1,5 @@
+# Ignore Gradle project-specific cache directory
```
molly-moen, line 1: should probably add .class files to the gitignore and remove them from this pr :)

org-code-codebuilderapp/app/bin/main/static/app.js
```diff
@@ -0,0 +1,50 @@
+var stompClient = null;
```
molly-moen, line 1: should this websocket code be split into a separate PR?

org-code-codebuilderapp/.gitignore
jmkulwik: Oops! I didn't realize I was still on the same branch as this PR. I just kept pushing my work-in-progress.

jmkulwik:
Sorry for the force-push. I had accidentally been using this as a WIP branch after opening the PR.

PR #2, opened 2021-02-17
Set up code builder app with websockets to individual anonymous users.

jmkulwik:
This sets up the initial gradle project for the app that will compile/run java and send the output on a websocket connection.

To test: 
`gradle bootRun`
`open localhost:8080`

**This PR does the following**
* Sets up the initial gradle project - I used `gradle init` with the first option selected as `2: App`. All other options were default
* Sets up a Spring Boot app 
* Sets up websocket connections with responses sent to individual sessions
* Sets up a basic UI for demo purposes only - no form of this will be used in production.

**Expect the following follow-up PRs**
* Project "cleanliness" - adding a linter and updating the .gitignore files
* Dockerfile - setting up the basic dockerfile that will allow this to be deployed to fargate
* Compile/run - updating the app to compile and run user-supplied Java

### Details
Some items to consider while reviewing
- The package name will be `org.code.codebuilderapp` - the org.code portion is convention. Anything after that is our choice.
- Currently, this is set up to allow anonymous users to access the service. This will be limited to verified users prior to adding the ability to run arbitrary Java.
- I chose gradle over maven or ant. This was because it was the more modern option and because it was the first one suggested by AWS when I was initially working with lambdas. If there are opinions around using a different build automation tool, please share :) 

### Background
**STOMP**
This is set up using the [STOMP](https://en.wikipedia.org/wiki/Streaming_Text_Oriented_Messaging_Protocol) protocol. The basic workflow looks like this
* Client CONNECTs to the websocket
* Client SUBSCRIBEs to a specific channel
* Client SENDs a request to the channel
* Responses are sent to subscribers on the channel
* Client DISCONNECTs from the channel

**SockJS**
This is currently using [SockJS](https://github.com/sockjs/sockjs-node/tree/v0.3.19#what-is-sockjs) - "a JavaScript library (for browsers) that provides a WebSocket-like object" - it attempts to set up a websocket, and if unsuccessful, attempts multiple fallback options. 
This may be replaced or updated when we integrate with AWS API Gateway.

**SpringBoot**
This handles much of the app setup for webapps. 
[SpringBoot starting guide for websockets](https://spring.io/guides/gs/messaging-stomp-websocket/)
[Using SpringBoot websockets with messages to individual anonymous sessions](https://www.mokkapps.de/blog/sending-message-to-specific-anonymous-user-on-spring-websocket/)


org-code-codebuilderapp/app/src/main/java/org/code/codebuilderapp/AnonymousUser.java
```diff
@@ -0,0 +1,20 @@
+package org.code.codebuilderapp;
+
+import java.security.Principal;
+
+/**
+ * A principal associated with a single anonymous session.
```
jmkulwik, line 6: Placeholder until we have access control figured out. 
Some discussion on this here: https://codedotorg.slack.com/archives/C03CK49G9/p1613502389142700?thread_ts=1613425248.140900&cid=C03CK49G9

org-code-codebuilderapp/app/src/main/java/org/code/codebuilderapp/AnonymousHandshakeHandler.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.UUID;
+import org.springframework.http.server.ServerHttpRequest;
+import org.springframework.web.socket.WebSocketHandler;
+import org.springframework.web.socket.server.support.DefaultHandshakeHandler;
+
+/**
+ * Assignes a unique id to each handshake to allow messages to be directed
+ * to specific anonymous users.
+ */
+public class AnonymousHandshakeHandler extends DefaultHandshakeHandler {
```
jmkulwik, line 14: Placeholder until we have access control figured out.
Some discussion on this here: https://codedotorg.slack.com/archives/C03CK49G9/p1613502389142700?thread_ts=1613425248.140900&cid=C03CK49G9

org-code-codebuilderapp/app/src/main/resources/static/app.js
```diff
@@ -0,0 +1,50 @@
+var stompClient = null;
```
jmkulwik, line 1: UI for test purposes only.

org-code-codebuilderapp/app/src/main/resources/static/index.html
```diff
@@ -0,0 +1,56 @@
+<!DOCTYPE html>
```
jmkulwik, line 1: UI for test purposes only.

org-code-codebuilderapp/gradlew
```diff
@@ -0,0 +1,185 @@
+#!/usr/bin/env sh
```
jmkulwik, line 1: Autogenerated by gradle. This, the gradle-wrapper.jar file, and the gradlew.bat file are all auto-generated and should all be checked into version control according to the gradle docs: https://docs.gradle.org/current/userguide/gradle_wrapper.html

molly-moen (approved):
LGTM!

org-code-codebuilderapp/app/build.gradle
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+plugins {
+  id 'org.springframework.boot' version '2.4.2'
+  id 'io.spring.dependency-management' version '1.0.11.RELEASE'
+  id 'java'
+}
+
+group = 'org.code'
+version = '0.0.1-SNAPSHOT'
```
jamescodeorg, line 24: My memory is a little fuzzy, but I think there are some implications of using a SNAPSHOT version and also having the version string not change from build to build.  Maybe just double-check that the resulting behavior is what you want?

org-code-codebuilderapp/app/src/main/java/org/code/codebuilderapp/CodeBuildController.java
```diff
@@ -0,0 +1,33 @@
...
+ * channel.
+ */
+@Controller
+public class CodeBuildController {
+
+  private final CompileRunService compileRunService;
+
+  CodeBuildController(CompileRunService compileRunService) {
+    this.compileRunService = compileRunService;
+  }
```
jamescodeorg, line 20: (minor) In case you didn't already know, I think there's some Spring annotation that will automatically inject this dependency.

org-code-codebuilderapp/app/build.gradle
```diff
@@ -0,0 +1,45 @@
...
+sourceCompatibility = '1.8'
+
+repositories {
+  mavenCentral()
+}
+
+dependencies {
+  implementation 'org.springframework.boot:spring-boot-starter-websocket'
+  implementation 'org.webjars:webjars-locator-core'
+  implementation 'org.webjars:sockjs-client:1.0.2'
```
jamescodeorg, line 34: Is there a way to have the demo UI be more separate from the rest of the app?  I'm thinking that we'll ultimately want to ship the service without the UI (right?) and it would be good to keep the js dependencies out of the main list here.  Would it be too much work to build the js project completely separately from the java code?

org-code-codebuilderapp/app/build.gradle
```diff
@@ -0,0 +1,45 @@
+/*
+ * Usage:
+ * gradle build -> builds the project
+ * gradle bootRun -> sets up a local server -> can be accessed at localhost:8080
+ *
+ * To get your changes into the docker container, do the following
+ * gradle build
+ * cp app/build/libs/app-0.0.1-SNAPSHOT.jar dockerContainer
```
jamescodeorg, line 8: Do we want to name our jar something a little more descriptive than 'app'?  Btw, what are calling this service?

org-code-codebuilderapp/app/src/main/java/org/code/codebuilderapp/WebSocketConfig.java
```diff
@@ -0,0 +1,43 @@
...
+
+  /**
+   * Sets up the channels that a client can subscribe to. Specifically, a client
+   * would subscribe to /topic/destination and would send messages to
+   * /app/destination.
+   */
+  @Override
+  public void configureMessageBroker(MessageBrokerRegistry config) {
+    config.enableSimpleBroker("/topic");
+    config.setApplicationDestinationPrefixes("/app");
```
jamescodeorg, line 24: Were the names "/topic" and "/app" picked because they're conventional?  (I'm wondering if we want to pick something that's a bit more descriptive.)

jamescodeorg (commented):
My Java's rustier than I thought so please ignore anything that doesn't make sense.  :-)

org-code-codebuilderapp/app/src/main/java/org/code/codebuilderapp/WebSocketConfig.java
jmkulwik: Hah they're actually configuration settings (not sure why they're strings rather than enums...). 

From the [spring docs](https://docs.spring.io/spring-framework/docs/current/reference/html/web.html#websocket-stomp-message-flow): "For example, STOMP messages whose destination headers start with /app may be routed to @MessageMapping methods in annotated controllers, while /topic and /queue messages may be routed directly to the message broker."

org-code-codebuilderapp/app/build.gradle
jmkulwik: Good callout! We haven't chosen a name for the service. I'm currently using `codebuilder`, but I'd love suggestions for something more descriptive but still short.
(Oh also, I'm taking your suggestion and updating it to not use app 🙂)

org-code-codebuilderapp/app/build.gradle
jmkulwik: Hmm... I'm planning to remove the demo UI as soon as we integrate with code studio (i.e. over the next sprint or two). Probably wouldn't be worth it to invest in separating it for such a short period of time.

jmkulwik: Although now I'm thinking of how nice it was to have a minimal client application that I could use for testing. May be nice to keep that long-term.

org-code-codebuilderapp/app/src/main/java/org/code/codebuilderapp/CodeBuildController.java
jmkulwik: Thank you! Keep the minor tips coming! I'm so new to Spring that a lot of this still feels like magic :)

org-code-codebuilderapp/app/build.gradle
jmkulwik: Ooohhh thanks. I'll look into that.

org-code-codebuilderapp/app/src/main/java/org/code/codebuilderapp/CodeBuildController.java
jamescodeorg: Heh, at some point, I knew more about Spring autowiring than I wanted to and it's definitely black magic.  Which reminds me that the annotation that I used was `@Autowired` (although there may be newer, more specific annotations).  IIRC, for a class to be a target for autowiring, it also has to be annotated with @Service, @Controller, etc.  (Let me know if you have questions and I can look at the docs to give you more precise answers.)

org-code-codebuilderapp/app/build.gradle
jamescodeorg: I think something with the word Java (to indicate that it processes Java, not necessarily that it's written in Java) would be nice?  JavaRunner, JavaBuilder, JavaCompiler, JavaProcessor, ...?

org-code-codebuilderapp/app/src/main/java/org/code/codebuilderapp/WebSocketConfig.java
jamescodeorg: If we have control over these strings, it might be nice to pick something a bit more descriptive?

org-code-codebuilderapp/app/src/main/java/org/code/codebuilderapp/CodeBuildController.java
jmkulwik: Actually, a little help here would be nice. The docs around autowiring aren't the most descriptive. I'm having trouble understanding how it works or what the use case is for it. (most of the docs say something along the lines of "the annotation @Autowired allows you to autowire a constructor" 🙄)

org-code-codebuilderapp/app/build.gradle
jmkulwik: ✔️

org-code-codebuilderapp/app/src/main/java/org/code/codebuilderapp/WebSocketConfig.java
jmkulwik: ✔️

org-code-codebuilderapp/app/build.gradle
jmkulwik: JavaCodeBuilder?

jmkulwik: Ok. I pitched it to the team and we ended up on `JavaBuilder`. Thank you!

org-code-codebuilderapp/app/src/main/java/org/code/codebuilderapp/Destinations.java
```diff
@@ -0,0 +1,31 @@
+package org.code.codebuilderapp;
+
+public final class Destinations {
```
jamescodeorg, line 3: (minor) Probably helpful to have a class comment here.  (We'll be able to generate documentation from the header comments and they also show up in IDEs.)

jamescodeorg (approved):
LGTM!

org-code-codebuilderapp/app/src/main/java/org/code/codebuilderapp/Destinations.java
jmkulwik: ✔️

PR #5, opened 2021-02-19
Added docker integration

jmkulwik:
The command `gradle buildDocker` will create a docker container with the JavaBuilder app installed.


molly-moen (approved):
🎉

PR #7, opened 2021-02-25
Add Java Compile and Run Functionality

molly-moen:
Replace placeholder java compile/run code with real compile/run functionality. The compile/run uses [JavaCompiler[(https://docs.oracle.com/javase/8/docs/api/javax/tools/JavaCompiler.html) to create a .class file from arbitrary Java code. The .class file gets saved to a temporary directory, which we load with [URLClassLoader](https://docs.oracle.com/javase/8/docs/api/java/net/URLClassLoader.html).

We are able to run the class by utilizing Java's Class object's [getDeclaredMethod](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getDeclaredMethod-java.lang.String-java.lang.Class...-), which then lets us [invoke](https://docs.oracle.com/javase/8/docs/api/java/lang/reflect/Method.html#invoke-java.lang.Object-java.lang.Object...-) a method. The code currently behaves like Java and searches for a main method, giving an error if it doesn't find one. However, if we needed to call a specific student method for a lab that would be possible.

It takes around half a second for the backend to compile and run a Hello World program and communicate back to the frontend (all running locally).

This PR also includes a short README for run instructions and a few additions to .gitignores.
## Future Work
Currently this code only runs single file, non-interactive programs. We will need to expand to multi-file and interactive programs. There are also no limits on how much time or memory the program would use, which we will need to add. We will also need to investigate how much the given code can interact with the system and prevent any malicious use.


org-code-javabuilder/.gitignore
```diff
@@ -10,3 +10,5 @@ app/.project
 *.classpath
 .settings
 .project
+
+.vscode/*
```
jamescodeorg, line 14: (very minor) Not sure if you care of not, but this file doesn't end in a newline.

Btw, do you have checkstyle configured?  It probably wouldn't have caught this but it'd be good to get it set up before adding lots more Java code.

org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaSourceFromString.java
```diff
@@ -0,0 +1,26 @@
+package org.code.javabuilder;
+
+import java.net.URI;
+import javax.tools.SimpleJavaFileObject;
+
+/** A file object used to represent source coming from a string. */
```
jamescodeorg, line 6: (very minor) These early check-ins tend to have a way of establishing precedent so just pointing out some very minor things for your team to consider.

I prefer javadoc comments always being multi-line even when they're short.  It makes it easier for me to parse them out from the code and quickly get a sense of what a class does.

jamescodeorg (commented):
I have some suggestions about code organization.

A pattern that I've found to work well is to keep the controller focused mainly on the logic that "translates" between your transport layer and your business logic.  As an example, if we decided to switch from websockets to HTTP long-polling or a direct TCP connection, none of the business logic classes should have to change.  (This might be more of a thought exercise if we're not planning to support alternate transports but I find that it's useful to put logic in the appropriate layers.)

The compiler and executor functionality can probably be put in two separate classes.  Segregating this code is useful because it clarifies the error handling and hides the... complicated interactions with the java apis for compilation and execution.

Hopefully, this will keep the file fairly coherent in its dependencies and functionality and make it easier to unit test each individual piece (e.g. by mocking out the compiler or the executor).

org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaBuildController.java
```diff
@@ -21,9 +38,122 @@
   /** Executes the user code and sends the output of that code across the established websocket. */
   @MessageMapping(Destinations.EXECUTE_CODE)
   @SendToUser(Destinations.PTP_PREFIX + Destinations.OUTPUT_CHANNEL)
-  public UserProgramOutput execute(UserProgram userProgram, Principal principal) {
-    // Send fake output
-    compileRunService.sendMessages(principal.getName(), "Compiling...");
-    return new UserProgramOutput("Hello World!");
+  public UserProgramOutput execute(UserProgram userProgram, Principal principal) throws Exception {
```
jmkulwik, line 41: Should the exception type be more specific than `Exception`?

org-code-javabuilder/app/src/main/java/org/code/javabuilder/UserProgram.java
```diff
@@ -21,7 +21,7 @@ public String getCode() {
     return code;
   }
 
-  public void getFileName(String fileName) {
+  public void setFileName(String fileName) {
```
jmkulwik, line 24: Oops. Thanks :)

org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaBuildController.java
```diff
@@ -21,9 +38,122 @@
...
+
+  // Given a user program, create a compilation task that will save the .class file to the given
+  // temp folder and output any compilation messages to diagnostics.
+  public CompilationTask getCompilationTask(
+      UserProgram userProgram, File tempFolder, DiagnosticCollector<JavaFileObject> diagnostics)
+      throws IOException {
+    JavaCompiler compiler = ToolProvider.getSystemJavaCompiler();
+
+    // set output of compilation to be a temporary folder
+    StandardJavaFileManager fileManager = compiler.getStandardFileManager(null, null, null);
```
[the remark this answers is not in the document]
jamescodeorg: FWIW, IntelliJ will do this kind of annotation automatically when reading code in the IDE.

jmkulwik:
Some high level feedback:
- I ran this through Fargate, and it's _way_ faster. It takes basically the same amount of time to run in Fargate as to run locally. 🙂 
- I ran some longer-running programs from multiple sessions. They can all run at the same time without affecting performance of the others 🎉, but the output gets mixed up between the sessions. Not something we need to fix now (infra decisions may end up fixing this), but we should track it.
- Adding a +1 to James's thoughts about splitting classes & moving the logic out of the controller

org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaBuildController.java
```diff
@@ -21,9 +38,122 @@
...
+        new JavaSourceFromString(userProgram.getFileName(), userProgram.getCode());
+    Iterable<? extends JavaFileObject> compilationUnits = Arrays.asList(file);
+
+    // create compilation task
+    CompilationTask task =
+        compiler.getTask(null, fileManager, diagnostics, null, null, compilationUnits);
+    return task;
+  }
+
+  public void runClass(URL filePath, String className, Principal principal) {
```
jmkulwik, line 127: Possible convention: Do we want to javadoc comment all public methods?

molly-moen: I think that is a good call, if we don't start with it we will never add them later. Will update this PR with javadoc comments + refactoring out logic

org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaBuildController.java
```diff
@@ -21,9 +38,122 @@
...
+        PrintStream out = new PrintStream(outputStream);
+        System.setOut(out);
+
+        compileRunService.sendMessages(principal.getName(), "Compiled!");
+        compileRunService.sendMessages(principal.getName(), "Running your program...");
+        runClass(tempFolder.toURI().toURL(), userProgram.getFileName(), principal);
+
+        // outputStream should now contain output of userProgram
+        outputStream.flush();
+        String result = outputStream.toString();
```
jmkulwik, line 80: Is it possible to log more often? If I run something like this, the output doesn't show up until the end.
```
while(x < 10) {
  x++;
  Thread.sleep(1000);
  System.out.println("Hello World!");
}
```
Also not something that has to be addressed now (but should probably track).

org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaBuildController.java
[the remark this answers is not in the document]
molly-moen: A lot of these Java library functions have a bunch of parameters that are often unnecessary, so I am in favor of not over-commenting and relying on IDE capabilities to inform us of what the parameters are. It's also harder to mess up parameters in Java than in a language like Javascript or Ruby because everything is typed (this is a bad example because everything is null).

org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaSourceFromString.java
molly-moen: our linter enforces a single-line comment takes up one line. Do you think it's preferable to split this text into two lines or just have the start/end /* */ on their own lines? We may need to update our linter to allow that. I don't have a strong opinion either way.

jamescodeorg:
@jmkulwik Just curious, in your tests with longer-running programs, were there any programs in a tight loop (e.g. a for loop incrementing a counter, no I/O, no sleeps)?

org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaSourceFromString.java
jamescodeorg: I _think_ javadoc comments have different rules from regular comments are always allowed to be multiline?
https://google.github.io/styleguide/javaguide.html#s7.1.1-javadoc-multi-line

I prefer the /** and */ to be on their own lines so this block would be:
```
/** 
 * A file object used to represent source coming from a string.
 */
```

jmkulwik: Linter details if it's helpful:
Style Guide: https://google.github.io/styleguide/javaguide.html
Details: https://github.com/sherter/google-java-format-gradle-plugin

org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaBuildController.java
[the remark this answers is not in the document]
jmkulwik: 👍

jmkulwik:
> @jmkulwik Just curious, in your tests with longer-running programs, were there any programs in a tight loop (e.g. a for loop incrementing a counter, no I/O, no sleeps)?

Not that I remember. I did accidentally write a `while(true) print "hello world"` style program.

org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaSourceFromString.java
molly-moen: For now I'm going to leave as-is because we would have to change the linter to do the comment as above (it blocks check-in if I try a multi-line comment with one line of text).

org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaBuildController.java
molly-moen: added a [task](https://codedotorg.atlassian.net/browse/CSA-137)!

molly-moen:
@jamescodeorg @jmkulwik I tried a loop that prints out an incrementing counter 3 million times and it works, but interestingly, "Done!" gets sent before the program output. Something else to follow up on.

org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaRunner.java
```diff
@@ -0,0 +1,169 @@
...
+
+      if (compileSuccess) {
+        // set System.out to be a specific output stream in order to capture output of the
+        // program and send it back to the user
+        ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+        PrintStream out = new PrintStream(outputStream);
+        System.setOut(out);
+
+        compileRunService.sendMessages(principal.getName(), "Compiled!");
+        compileRunService.sendMessages(principal.getName(), "Running your program...");
```
jmkulwik, line 56: Definitely out of scope for this PR, but I wanted to mention it: We'll want to eventually consider how we internationalize strings from the JavaBuilder service. Or whether we can find a way to avoid using strings in the JavaBuilder service.

jmkulwik (approved):
Looks great! Thanks for making this happen! One final thought on the JavaRunner class - I think we'll likely want to split it up further as time goes on (and we add I/O, infinite loop detection, etc.) but probably not needed yet. Deferring to your judgement there.

org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaBuildController.java
jamescodeorg: This is something that the linter that I've used in the past (Checkstyle) can enforce.  Might be worth seeing if there is an equivalent feature in our linter (not necessarily as part of this commit).

jamescodeorg (commented):
+1 to Jessie's comment about splitting up JavaRunner in a class that compiles and a class that runs.

There are still a few things I'd like to see addressed here, but I'm ok with merging this and tackling further updates in a separate commit if you prefer.

org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaRunner.java
```diff
@@ -0,0 +1,169 @@
...
+
+  /**
+   * Compile and Run the given user program and output results to the given compileRunService for
+   * the given user (principal)
+   *
+   * @param userProgram
+   * @param principal
+   * @param compileRunService
+   */
+  public static void compileAndRunUserProgram(
```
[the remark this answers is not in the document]
molly-moen: if you have any resources on this and the autowired bean setup mentioned below that would be great!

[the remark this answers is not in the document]
jamescodeorg: Hmm, it was surprisingly hard to find something an article that informative and concise.  I found [this](https://medium.com/the-resonant-web/spring-boot-2-0-project-structure-and-best-practices-part-2-7137bdcba7d3) but will also start a thread in Slack.

org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaRunner.java
```diff
@@ -0,0 +1,169 @@
...
+      UserProgram userProgram, Principal principal, CompileRunService compileRunService) {
+    String filename = userProgram.getFileName();
+    // We expect the filename to have no .java suffix, remove it if necessary.
+    if (filename.endsWith(".java")) {
```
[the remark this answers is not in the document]
molly-moen: done!

jamescodeorg (approved):
Thanks for all the updates and good to merge!  I left a few more comments to be considered for future work.

org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaRunner.java
```diff
@@ -0,0 +1,182 @@
...
+          principal.getName(), "There was an issue trying to run your program, please try again.");
+      e.printStackTrace();
+    }
+
+    if (tempFolder != null) {
+      tempFolder.delete();
+    }
+
+    // ensure System.out is reset
+    System.setOut(System.out);
```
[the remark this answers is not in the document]
molly-moen: I am catching all possible exceptions (the function itself doesn't throw any exceptions--shouldn't that mean I always get to the end of the function?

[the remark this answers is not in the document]
jamescodeorg: The method can still throw unchecked exceptions.

IIRC, the whole checked vs. unchecked exceptions thing was really confusing in Java.  Some libraries (including Spring?) routinely throw unchecked exceptions because they viewed checked exceptions to be a language design flaw in Java.  Because of that, I think it is not generally safe to assume that all interesting exceptions are handled when your method signature does not have a throws clause.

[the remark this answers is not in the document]
molly-moen: good context, thanks!

PR #9, opened 2021-03-03
Allow console input

jmkulwik:
This PR enables console input in the following way:

- Add an input controller & route & model
- Replace System.in with a PipedInputStream - this allows us to connect a PipedOutputStream in order to write user input to the stream.
- Move Java execution to its own thread - this allows us to log output to the client while their program is running

Future work:
Enable thread-safe execution. Currently output from any user gets streamed to the most recent user.

### DIFF
--- a/org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaRunner.java
+++ b/org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaRunner.java
@@ -4,6 +4,7 @@ import java.io.*;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.security.Principal;
+import java.util.ArrayList;
 import java.util.Arrays;
 import javax.tools.Diagnostic;
 import javax.tools.DiagnosticCollector;
@@ -68,13 +69,17 @@ public class JavaRunner {
           JavaExecutorThread userRuntime = new JavaExecutorThread(tempFolder.toURI().toURL(), userProgram, principal, this.compileRunService);
           userRuntime.start();
           String result = null;
-          byte[] streamBytes = new byte[1024];
+
+          // todo find a better way of reading this stream
+          byte[] streamBytes = new byte[64];
           try {
-            while (systemOutputReader.read(streamBytes, 0, 1024) != -1) {
+            while (systemOutputReader.read(streamBytes, 0, 64) != -1) {
               result = new String(streamBytes, StandardCharsets.UTF_8);
-              streamBytes = new byte[1024];
+              streamBytes = new byte[64];
               if (result.length() > 0) {
                 this.compileRunService.sendMessages(principal.getName(), result);
+              } else {
+                systemOutputStream.flush();
               }
               try {
                 Thread.sleep(100);
@@ -109,9 +114,10 @@ public class JavaRunner {
   }
 
   public void passInputToRuntime(UserInput userInput, Principal principal) {
-    byte[] input = userInput.getInput().getBytes(StandardCharsets.UTF_8);
+    byte[] input = (userInput.getInput() + "\n").getBytes(StandardCharsets.UTF_8);
     try {
       this.systemInputWriter.write(input, 0, input.length);
+      this.systemInputWriter.flush();
     } catch (IOException e) {
       // Figure out what to do here
       e.printStackTrace();
--- a/org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaRunner.java
+++ b/org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaRunner.java
@@ -74,7 +74,7 @@ public class JavaRunner {
           String result = null;
 
           try {
-            while ((result = systemOutputReader.readLine()) != null && userRuntime.isAlive()) {
+            while ((result = systemOutputReader.readLine()) != null) {
               if (result.length() > 0) {
                 this.compileRunService.sendMessages(principal.getName(), result);
               } else {
--- a/org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaRunner.java
+++ b/org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaRunner.java
@@ -1,12 +1,7 @@
 package org.code.javabuilder;
 
-import java.io.ByteArrayOutputStream;
-import java.io.File;
-import java.io.IOException;
-import java.io.PrintStream;
-import java.lang.reflect.InvocationTargetException;
-import java.net.URL;
-import java.net.URLClassLoader;
+import java.io.*;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.security.Principal;
 import java.util.Arrays;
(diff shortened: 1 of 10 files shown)


org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaRunner.java
```diff
@@ -50,40 +47,63 @@ public void compileAndRunUserProgram(UserProgram userProgram, Principal principa
...
+    if (compileSuccess) {
+      this.compileRunService.sendMessages(principal.getName(), "Compiled!");
+      // We use a PipedInput and PipedOutput stream for I/O so we can control input to the user's
+      // program and output from the user's program.
+      // This try statement will close the streams automatically.
+      try (PipedInputStream inputStream = new PipedInputStream();
+          BufferedReader systemOutputReader =
+              new BufferedReader(new InputStreamReader(inputStream));
+          PipedOutputStream systemOutputStream = new PipedOutputStream(inputStream);
+          PrintStream out = new PrintStream(systemOutputStream)) {
```
jmkulwik, line 68: Note: IO listening/logging should be refactored into its own component. That will be done as part of the next task to enable thread safe logging.

org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaRunner.java
```diff
@@ -50,40 +47,63 @@ public void compileAndRunUserProgram(UserProgram userProgram, Principal principa
...
+        this.systemInputWriter = new PipedOutputStream(this.systemInputStream);
+        System.setIn(this.systemInputStream);
+
+        this.compileRunService.sendMessages(principal.getName(), "Running your program...");
+        JavaExecutorThread userRuntime =
+            new JavaExecutorThread(
+                tempFolder.toURI().toURL(), userProgram, principal, this.compileRunService);
+        userRuntime.start();
+
+        while (true) {
```
molly-moen, line 82: could this condition be `userRuntime.isAlive()` or some variation on that? If not, can we put a comment above explaining why we have `while(true)` here? It would also be good to add a todo to kill the program if it runs for too long.

jmkulwik: Hah right. Fixed. Thanks! 
Here's the todo for infinite loops/long running programs: [CSA-151](https://codedotorg.atlassian.net/browse/CSA-151)

molly-moen (approved):
🎉 very exciting!

org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaRunner.java
```diff
@@ -92,6 +111,26 @@ public void compileAndRunUserProgram(UserProgram userProgram, Principal principa
...
+  /**
+   * Write the given userInput to the user's (principal's) input stream
+   *
+   * @param userInput the given input directed to System.in
+   * @param principal the user's identification
+   */
+  public void passInputToRuntime(UserInput userInput, Principal principal) {
+    // System.in expects input to end with a lineSeparator. We add that to allow the user program to
+    // continue.
+    byte[] input = (userInput.getInput() + System.lineSeparator()).getBytes(StandardCharsets.UTF_8);
```
jamescodeorg, line 126: For my education, is the plan for the client to send a UserInput message when a newline is encountered (and the newline will be stripped before sending the message)?

org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaRunner.java
```diff
@@ -50,40 +47,62 @@ public void compileAndRunUserProgram(UserProgram userProgram, Principal principa
...
+        // TODO we'll need to stop execution after a certain time limit otherwise we'll hang here
+        // forever if a user writes infinite loops
+        while (userRuntime.isAlive() || systemOutputReader.ready()) {
+          if (systemOutputReader.ready()) {
+            // The user's program has produced output. Read it and pass it to the client console.
+            this.compileRunService.sendMessages(principal.getName(), systemOutputReader.readLine());
+          } else {
+            // The user's program is running, but there's no output. Let it continue running.
+            try {
+              Thread.sleep(200);
```
jamescodeorg, line 91: Not sure if this is part of what will be refactored, but polling loops like this are generally not great for scalability.  Not sure what the best design here is off the top of my head, let me know if you want to brainstorm together.

jamescodeorg (commented):
Two general comments:
1) I think we're starting to get to some more interesting logic that should probably have test coverage.  Is the plan to backfill that later?
2) If the server crashes, all open sessions will be lost.  Is that something that we're planning to live with to reduce complexity or is it something that we plan to address at some point?

org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaRunner.java
jmkulwik: Some of this will be dependent on our AWS infrastructure, so I didn't want to sink too much time into it just yet. One possibility I've thinking about (curious for your thoughts): Rather than overriding System.in and System.out with a generic stream, override them with a thread-specific custom-written stream that logs directly to the user's channel. It would be set up similar to this: https://stackoverflow.com/a/15526391

org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaRunner.java
jmkulwik: Hmm yeah, the way this is written is a little odd. Initially, I just added this in because the program kept hanging when I forgot to hit "enter" before hitting the button to send the message. To answer your question: we haven't planned out the client message format yet. A couple things that make me think we should keep some version of this line:
- We can't be sure that all clients platform & browser combinations will pass a line separator that this Linux server will recognize
- It's likely we'll find some creative way of passing input from the client in the future where it wouldn't logically make sense to add a line separator (i.e. from a form or similar). It would be difficult to always remember to add a line separator in those cases.

Perhaps a better way to write this would be to check for a system-recognized line separator and add one if none exists.

jmkulwik:
@jamescodeorg 
> 1. I think we're starting to get to some more interesting logic that should probably have test coverage.  Is the plan to backfill that later?

^ Good call on testing. I would say we definitely could start adding test coverage at this point. At the same time, I'm getting the feeling that we're going to end up with a significant refactor/rewrite once we get the AWS infra worked out (next couple weeks). For now, I would treat this like a prototype still. If we do too much more work here, though, it'll be more than a prototype and we won't be able to keep using that excuse.

> 2. If the server crashes, all open sessions will be lost.  Is that something that we're planning to live with to reduce complexity or is it something that we plan to address at some point?

^ Interesting question. I'm guessing there may be some discussion from this. Mind asking on the [infra dev doc](https://docs.google.com/document/d/196aKj947BYZXZH3nGvzHprgWPOoy2Kw1x3sKgziUaSo/edit#)?

org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaRunner.java
jamescodeorg: That sounds reasonable.  I think part of my question was whether the protocol for sending the input stream is line-based (i.e. input is sent line by line).  I think this simplifies things, but does impose some constraints (e.g. can't submit input without hitting enter).  If the protocol is line-based, we should be clear about whether the newline should or should not be sent by the client (or perhaps that we want to be lenient and accept both but I would not recommend that, at least to start).

org-code-javabuilder/app/src/main/java/org/code/javabuilder/JavaRunner.java
jamescodeorg: Interesting.  I'd need to spend more time with this to form an opinion but I think it's likely that we'll need to replace System.in and System.out with something custom.  I guess a general question is that, if System.out is shared across the process, how can we identify which student program output is coming from?  It sounds like we're thinking by-thread which is reasonable (but very strongly ties a build/run session with a thread).  (Also, it'd be good to double check on the thread safety guarantees for the various print calls to console.)

PR #10, opened 2021-03-26
Create lambda function for JavaBuilder

jmkulwik:
This will be the basic structure for JavaBuilder. It is basically a big function that gets invoked once per student. The entry point is `LambdaRequestHandler.handleRequest`

input would look something like this:
```
{
  "queueUrl": <urlString>,
  "apiEndpoint": <urlString>,
  "connectionId": <idString>
}
```

I'm looking for feedback on the structure of the Java code (*.java files). Some questions:
 - What are the best ways to prevent the busy waiting? 
 - Is the program structured correctly?
 - Should there be more use of singletons/static functions?
 - General feedback?

Instructions for trying out the lambda function are here: https://codedotorg.slack.com/archives/C01EF4GJ9GE/p1616805551042500


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaBuilder.java
```diff
@@ -0,0 +1,66 @@
...
+
+    // Create output poller
+    this.outputPoller = new OutputPoller(this.javaRunner, this.outputHandler, runtimeIO, this.outputSemaphore);
+  }
+
+  public void runUserCode() {
+    this.javaRunner.start();
+    this.inputPoller.start();
+    this.outputPoller.start();
+    while(javaRunner.isAlive()) {
```
jmkulwik, line 50: I'm unsure how to get around this instance of busy-waiting. We can't let LambdaRequestHandler.handleRequest return until the student's code is finished running. Tips? Ideas?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/OutputPoller.java
```diff
@@ -0,0 +1,39 @@
...
+        message = runtimeIO.pollForOutput();
+      } catch (IOException e) {
+        outputHandler.sendMessage("There was an error reading output from your program. Try running it again." + e.toString());
+      }
+      if (message != null) {
+        outputHandler.sendMessage(message);
+      }
+
+      try {
+        Thread.sleep(400);
```
jmkulwik, line 33: I think we can prevent this busy waiting by calling the outputHandler any time system.out.print* is invoked. If anyone has tips here, they're definitely welcome!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/InputPoller.java
```diff
@@ -0,0 +1,44 @@
...
+      for (Message message : messages) {
+        try {
+          runtimeIO.passInputToProgram(message.getBody());
+        } catch (IOException e) {
+          outputHandler.sendMessage("There was an error passing input to your program." + e.getStackTrace());
+        }
+        sqsClient.deleteMessage(queueUrl, message.getReceiptHandle());
+      }
+      try {
+        Thread.sleep(400);
```
jmkulwik, line 38: I think we can prevent this busy waiting by polling the SQS client any time system.in.* is invoked. If anyone has tips on overriding system.in to add this functionality, please share!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaBuilder.java
```diff
@@ -0,0 +1,66 @@
...
+import java.io.IOException;
+
+public class JavaBuilder {
+  private final OutputHandler outputHandler;
+  private final InputPoller inputPoller;
+  private final OutputPoller outputPoller;
+  private final JavaRunner javaRunner;
+  private final OutputSemaphore outputSemaphore;
+
+  public JavaBuilder(String connectionId, String apiEndpoint, String queueUrl) {
```
jmkulwik, line 18: Is there a better way to structure this? Here, we're just instantiating a bunch of local variables to inject them into the other classes. Is that the right way to do it?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/OutputHandler.java
```diff
@@ -0,0 +1,30 @@
+package org.code.javabuilder;
+
+import java.nio.ByteBuffer;
+import com.amazonaws.services.apigatewaymanagementapi.AmazonApiGatewayManagementApi;
+import com.amazonaws.services.apigatewaymanagementapi.model.PostToConnectionRequest;
+
+public class OutputHandler extends Thread {
```
jmkulwik, line 7: Should this be a singleton? That seems somewhat discouraged yet common in the Java community.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaBuilder.java
```diff
@@ -0,0 +1,71 @@
...
+    // Create output poller
+    this.outputPoller =
+        new OutputPoller(this.javaRunner, this.outputHandler, runtimeIO, this.outputSemaphore);
+  }
+
+  public void runUserCode() {
+    this.javaRunner.start();
+    this.inputPoller.start();
+    this.outputPoller.start();
+    while (javaRunner.isAlive()) {
```
jmkulwik, line 53: I'm unsure how to get around this instance of busy-waiting. We can't let LambdaRequestHandler.handleRequest return until the student's code is finished running. Tips? Ideas?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/InputPoller.java
molly-moen: You can reset System.in to an InputStream with [System.setIn](https://docs.oracle.com/javase/7/docs/api/java/lang/System.html#setIn(java.io.InputStream). I think you could define a custom input stream that does some fancy stuff here, although I'm not sure how good of a pattern that would be. Another thing to think about is would there be a case when user input is provided in a way other than System.in? And will it always look like System.in when we are running student-provided code?

molly-moen: I did something somewhat similar in my initial try at running student code for System out. [link](https://github.com/code-dot-org/java-ide/pull/7/files#diff-17458634177104ff6ee8dba2da91fd742a5e681ab48ca3d0c36b188e673481d1R63)

PR #11, opened 2021-03-31
Create long-running lambda that accepts input and output to students' programs

jmkulwik:
This is the lambda function that will execute student code. The entry point is `LambdaRequestHandler.handleRequest`. 

This particular PR sets up the console input/output for student programs. 

Console input will be generated by the client when a user presses the "Enter" key when focused is on the console `<input>` element. This input will be passed to Amazon API Gateway which will forward the input to AWS SQS (Simple Queue Service) via a lambda function. When the student code invokes "Read" on the System.in stream (via `Scanner.readLine()` or otherwise), the `AWSInputAdapter` will query AWS SQS to get the user input to the program. To stay consistent with Java console input, we will block until the user has provided input to the program.

Console output will be generated by the program when the program invokes `System.out.print()` or similar functions. This will cause output to be collected and passed to the `AWSOutputAdapter` which will issue a POST request to API Gateway via the `@connections` api. API Gateway will then direct the program's output to the correct client which will be displayed in the client's console element.

See the [JavaBuilder Dev Doc](https://docs.google.com/document/d/196aKj947BYZXZH3nGvzHprgWPOoy2Kw1x3sKgziUaSo/edit#heading=h.6l2rrk1zc9m5) for details regarding how the various microservices will interact.

Instructions for trying out the lambda function are here: https://codedotorg.slack.com/archives/C01EF4GJ9GE/p1616805551042500


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSInputAdapter.java
```diff
@@ -0,0 +1,47 @@
...
+   * Attempts to access the first element in the fifo queue. If none exist, queries available input
+   * from Amazon SQS and stores each input in the `messages` fifo queue. Messages are deleted after
+   * they are retrieved. This is a blocking call.
+   *
+   * @return the first message in the fifo queue.
+   */
+  public String getNextMessage() {
+    ReceiveMessageRequest request = new ReceiveMessageRequest();
+    request.setQueueUrl(queueUrl);
+    request.setWaitTimeSeconds(20);
```
molly-moen, line 32: for my information: how does this wait time work? Does it wait up to 20 seconds on each call to getMessages()? Something else? (may be worth adding a comment)

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/InputRedirectionStream.java
```diff
@@ -0,0 +1,98 @@
...
+      }
+    }
+
+    return (int) queue.remove();
+  }
+
+  /** See: https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#read-byte:A- */
+  @Override
+  public int read(byte[] b) {
+    return read(b, 0, b.length);
```
molly-moen, line 46: style question: when we call methods within a class, should we preface with `this.`? I personally find it easier to read because I know we are calling another method within this class, but am not sure what the preferred style is.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/InputRedirectionStream.java
```diff
@@ -0,0 +1,98 @@
...
+    if (b == null) {
+      throw new NullPointerException();
+    }
+    if (off < 0 || len < 0 || len > b.length - off) {
+      throw new IndexOutOfBoundsException();
+    }
+
+    int k = 0;
+    while (k < len) {
+      b[off + k++] = (byte) read();
```
molly-moen, line 63: nit: could we split `b[off + k++]` into a couple lines? At least one to increment `k`, maybe another to calculate the index we are inserting and give it a name, although `off + k` could be fine too.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/OutputRedirectionStream.java
```diff
@@ -0,0 +1,74 @@
...
+    this.buffer = new StringBuilder();
+  }
+  /**
+   * See: https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#read-- Checks the queue
+   * for existing bytes. If the queue is empty, polls the inputAdapter for new data. This is a
+   * blocking call.
+   *
+   * @return the first byte in the queue
+   */
+  //
```
molly-moen, line 27: missing a function here? Or do we need to remove this comment?

molly-moen (commented):
This looks awesome! Just a couple nits :)

jamescodeorg (commented):
Sorry, this isn't done yet, but I have to run to some meetings but wanted to get some quick thoughts out.

There's some conversion here between messages, lines, and characters.  Could you elaborate on what you think the browser <--> server protocol will be?  When/how is the client expected to send input?  When/how is the server expected to send output?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSInputAdapter.java
jmkulwik: Yes, that's what it does - it could return much faster, but 20s is the max. I'll add a comment.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/InputRedirectionStream.java
jmkulwik: That sounds good to me. I also find that easier to read.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/InputRedirectionStream.java
jmkulwik: Good call. I'll do that.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/OutputRedirectionStream.java
jmkulwik: Haha oops. Thank you.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/InputRedirectionStream.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.LinkedList;
+import java.util.Queue;
+
+/**
+ * An InputStream that queries an InputAdapter for new bytes. This is intended to redirect system.in
+ * to use the InputAdapter rather than the server's console. See
+ * https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html for full implementation
+ * details & contract.
+ */
+public class InputRedirectionStream extends InputStream {
```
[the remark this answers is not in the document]
jmkulwik: That was the [original model](https://github.com/code-dot-org/java-ide/pull/10/files#diff-0fb41f59205338bc121571090988be2d73294a07927aaac2201eacd426642876) with the piped input and output streams. It was very confusing, and took up to 5 seconds to set up each time it ran. It's not a common pattern and therefore doesn't have a ton of documentation and the documentation is mostly on stack overflow. This method, however, was very clearly documented on the oracle docs. The implementation contract is precisely documented - it basically tells you the code to write line by line and when to throw various types of exceptions.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/InputRedirectionStream.java
```diff
@@ -0,0 +1,98 @@
...
+  private final Queue<Byte> queue;
```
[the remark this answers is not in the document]
jmkulwik: Are you talking about `Queue` or `Byte`? Any recommendations for a replacement?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/OutputRedirectionStream.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.io.OutputStream;
+
+/**
+ * An OutputStream that passes output to an OutputAdapter. It is intended to redirect output from
+ * the user program that is intended for the console. See
+ * https://docs.oracle.com/javase/8/docs/api/java/io/OutputStream.html for full implementation
+ * details and contract.
+ */
+public class OutputRedirectionStream extends OutputStream {
```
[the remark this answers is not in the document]
jmkulwik: Were you intending to ask this about the InputStream?

jmkulwik:
> Sorry, this isn't done yet, but I have to run to some meetings but wanted to get some quick thoughts out.
> 
> There's some conversion here between messages, lines, and characters. Could you elaborate on what you think the browser <--> server protocol will be? When/how is the client expected to send input? When/how is the server expected to send output?

Updated the PR description with these details. Let me know if more/other detail would be appropriate!

sureshc:
In the future we may want to define a protocol for the messages exchanged between the IDE and the JavaBuilder service over the websocket. Some messages exchanged between the two might be something other than end user input to the student project or output from the student project returned to the end user (interactive debugging, heart beat to keep session alive, terminate execution, terminate & restart, recompile, etc.).

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/OutputRedirectionStream.java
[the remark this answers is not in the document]
jamescodeorg: Oh yes, this comment should have been for the input side.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaBuilder.java
```diff
@@ -0,0 +1,25 @@
...
+    this.outputAdapter = outputAdapter;
+    this.inputAdapter = inputAdapter;
+  }
+
+  /**
+   * Sets replaces System.in and System.out with our custom implementation and executes the user's
+   * code.
+   */
+  public void runUserCode() {
+    System.setOut(new PrintStream(new OutputRedirectionStream(this.outputAdapter), true));
```
jamescodeorg, line 20: Maybe consider having the OutputRedirectionStream extend PrintStream (so we don't need the call to PrintStream here)?  In particular, the `true` parameter here is very important to this working the way you expect so it might be good to encapsulate that in the OutputRedirectionStream?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/OutputRedirectionStream.java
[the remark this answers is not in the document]
jamescodeorg: Ignore this comment.

jamescodeorg (approved):
Approving as this is a solid framework that we can iterate on.

It would be great to clarify whether the input/output protocol from the client to the service (via websocket) is binary, character-based, or line-based.  Currently, it looks like there is a bit of an asymmetry between input and output which could be confusing?  I think my recommendation would be either binary or character-based (depending on whether you think you'll need to send arbitrary data) in both directions but with strict rules around expected encoding and line endings.  Also, +1 to Suresh's comment about designing the protocol in a way that we can send other types of messages in the future.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/InputRedirectionStream.java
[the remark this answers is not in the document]
jmkulwik: I went ahead and tested this out with a 32KB message (max size of a websocket frame), and it took 5ms to box and add all the Bytes to the LinkedList. Interestingly, it took 25ms to create the byte array (line 34) and copy all the bytes into it. Given the small message size we're expecting, I would suggest we stay with Queue<Byte> as it simplifies the code complexity.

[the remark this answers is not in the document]
jmkulwik: Realized I was calling `toString` in the middle of one of the measurements. Updated for more accurate times.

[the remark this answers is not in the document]
jamescodeorg: Ok, there's also additional objects and memory allocated (and the GC pressure that can cause), but I'm ok with starting with this implementation and deferring changes until this becomes an issues.

PR #13, opened 2021-04-06
Add lambda authorizer for API Gateway

molly-moen:
Code for the [lambda authorizer function](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-websocket-api-lambda-auth.html) for out API Gateway websocket connect handler. The authorizer verifies that the request contains a header `Authorization: Bearer <token>`. This token must be generated by our dashboard `javabuilder_session_controller`, and the dashboard and API Gateway environments must match (development->development, staging->staging, etc.). The authorizer also verifies the token has not expired and the issued at time is not in the future.

The authorizer passes on the entire decoded contents of the token to API Gateway for now as `context`, so the downstream lambdas can use any part of the token they need. See details on token generation in [this PR](https://github.com/code-dot-org/code-dot-org/pull/39837). See [this doc](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-lambda-authorizer-output.html) for details on how to use the `context`.

See [slack](https://codedotorg.slack.com/archives/C01EF4GJ9GE/p1617725553019400) for details on how to test this PR. 

To deploy, you can run the script `zip_for_aws` and upload the `function.zip` file generated to a lambda.


javabuilder-authorizer/lambda_function.rb
```diff
@@ -0,0 +1,78 @@
...
+#
+# method_arn is in format
+# arn:aws:execute-api:region:account-id:api-id/stage-name/$connect
+# We only care about stage--that tells us the environment (development, staging, etc.)
+def get_public_key(method_arn)
+  tmp = method_arn.split(':')
+  api_gateway_arn = tmp[5].split('/')
+  stage_id = api_gateway_arn[1]
+
+  key_file = File.open("./public_keys/javabuilder_rsa_#{stage_id}_pub.pem")
```
sureshc, line 75: Could we use Lambda environment variables to store the public keys as configuration settings separate from the code? There seem to be a [few different options](https://www.concurrencylabs.com/blog/configure-your-lambda-function-like-a-champ-sail-smoothly/) for establishing configuration settings, each with their own pros & cons

javabuilder-authorizer/lambda_function.rb
```diff
@@ -0,0 +1,78 @@
...
+        verify_iat: true, # verify issued at time is valid
+        algorithm: 'RS256'
+      )
+    rescue JWT::ExpiredSignature, JWT::InvalidIatError
+      return generate_deny(nil, method_arn)
+    end
+
+    return generate_deny(nil, method_arn) unless decoded_token
+
+    generate_allow('me', method_arn, decoded_token[0])
```
jmkulwik, line 30: Could you leave a comment for `'me'` here? i.e. I'm unsure if it's a placeholder or a value with meaning.

javabuilder-authorizer/lambda_function.rb
```diff
@@ -0,0 +1,78 @@
...
+# Load the JWT token from public_keys based on the environment the request
+# was sent from.
+#
+# method_arn is in format
+# arn:aws:execute-api:region:account-id:api-id/stage-name/$connect
+# We only care about stage--that tells us the environment (development, staging, etc.)
+def get_public_key(method_arn)
+  tmp = method_arn.split(':')
+  api_gateway_arn = tmp[5].split('/')
+  stage_id = api_gateway_arn[1]
```
jmkulwik, line 73: Minor NIT: `stage_id` is used here, but `stage-name` is used in the comment. It took me a bit to realize they were the same thing.

javabuilder-authorizer/lambda_function.rb
molly-moen: that's a good call, didn't think of that. Only downside is when we move the lambda from the dev environment we'll have to copy over the environment variables, but that's probably worth it.

javabuilder-authorizer/lambda_function.rb
molly-moen: good callout. I had copied that from the sample code and hadn't thought about it since. I tested and it seems we can put anything here, so I'm thinking I will change this to user id.

javabuilder-authorizer/lambda_function.rb
sureshc: I will look at the Dance Party Video Export Lambda tomorrow to see how we manage configuration settings across its staging and production deployments.

javabuilder-authorizer/lambda_function.rb
molly-moen: updated to use environment variables. It's slightly less ideal because the newlines aren't preserved in an environment variable, so I have to find/replace the escaped newline with a real one.

javabuilder-authorizer/lambda_function.rb
molly-moen: unified around stage_name, which makes more sense.

javabuilder-authorizer/lambda_function.rb
molly-moen: updated to use environment variables for now (put keys for all envs, with suffix for which env it is). It's slightly less ideal because the newlines aren't preserved in an environment variable, so I have to find/replace the escaped newline with a real one.

javabuilder-authorizer/lambda_function.rb
```diff
@@ -0,0 +1,88 @@
...
+        algorithm: 'RS256'
+      )
+    rescue JWT::ExpiredSignature, JWT::InvalidIatError
+      return generate_deny(nil, method_arn)
+    end
+
+    return generate_deny(nil, method_arn) unless decoded_token
+
+    user_id = decoded_token[0]['uid']
+    generate_allow(user_id, method_arn, decoded_token[0])
```
sureshc, line 31: Since the `generate_allow` method invokes the `generate_policy` method, passing the `user_id` as the `principal_id` in the policy document, I think we would need to configure every Code Studio user in AWS IAM as a `Principal` before we receive the HTTP request? I think we may need to statically provision an IAM Role for a javalab client via CloudFormation and then have the authorizer allow the request to assume that role? I’ll research tomorrow as well.

molly-moen: ah ok good to know. I couldn't find any information on what the principal id needed to be, and it just worked with both 'me' and a random user id so I didn't look much further. Let me know what is recommended here.

jmkulwik: How could this have worked for Molly's local testing? I don't think any of our Code Studio users are set up in AWS IAM.

jmkulwik (approved):
Looks good from my perspective :)

javabuilder-authorizer/lambda_function.rb
sureshc: I'm still researching this morning on the best way to set the generated policy (and the principal_id specified in the policy).

sureshc: I'm used to a Principal identifying a specific User or Role defined in AWS IAM or an AWS Service that needs to invoke other AWS APIs on our behalf, but [reading discussions online](https://forums.aws.amazon.com/thread.jspa?threadID=225671), it's clear that setting this to the Code Studio user id is correct and we don't need to provision those users ahead of time anywhere in AWS. The user id will get logged as the Principal invoking the API Gateway route in some places, so might be useful to prefix/namespace the user ids: `studio.code.org/user_id/####`, `staging-studio.code.org/user_id/####`, etc.

molly-moen: awesome, thank you! Should the user id we pass in the token be namespaced or should just the principal id be namespaced?

sureshc: prefixing the user_id with the `issuer` from the token might work. We'd have to update the Code Studio controller that issues the token to set the `issuer` to the Code Studio dashboard domain name (`CDO.dashboard_domain` ?)

molly-moen: added issuer [here](https://github.com/code-dot-org/code-dot-org/pull/39992) and updated the principal id to be namespaced with the issuer.

PR #14, opened 2021-04-06
Integrate with S3

jmkulwik:


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/ProjectFileLoader.java
```diff
@@ -0,0 +1,66 @@
...
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.util.concurrent.Callable;
+
+/**
+ * Fetches an individual file from a student's project. This should be used with the
+ * {dashboardHost}/v3/files/{channelId}/{fileName} API. It has no permissions to use AWS S3. This
+ * should be used asynchronously so multiple files can be retrieved at once.
```
jamescodeorg, line 14: Maybe clarify this last sentence to say that this class is intended to be called by an executor?  (The code here doesn't really have a requirement to be called asynchronously.)

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/ProjectFileLoader.java
```diff
@@ -0,0 +1,66 @@
...
+    this.projectFile = projectFile;
+    this.fileUrl = fileUrl;
+  }
+
+  /**
+   * @return true if the load succeeds
+   * @throws UserFacingException If any errors occur while fetching the file.
+   */
+  @Override
+  public Boolean call() throws UserFacingException {
```
jamescodeorg, line 34: I'm not sure if this makes the code more or less readable, but it's possible to make this a function in UserProjectFileManager and then call it using an anonymous lambda.  (I suggest this because this class seems very tightly coupled with UserProjectFileManager and that would be a way of making that more explicit.)

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/ProjectFileLoader.java
```diff
@@ -0,0 +1,66 @@
...
+    try {
+      URL url = new URL(fileUrl);
```
jamescodeorg, line 36: This seems really low-level, perhaps [java.net.http.HttpClient](https://docs.oracle.com/en/java/javase/11/docs/api/java.net.http/java/net/http/HttpClient.html) or Apache's [HttpClient](https://hc.apache.org/httpcomponents-client-5.0.x/index.html) would be a better choice?  I think both of these also support async call patterns so you might be able to get rid of this class entirely.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileManager.java
```diff
@@ -1,15 +1,65 @@
...
 
+  /**
+   * Asynchronously loads all user project files.
+   *
+   * @throws UserFacingException If the load times out or fails.
+   */
   public void loadFiles() throws UserFacingException {
-    throw new UserFacingException("We couldn't load your files");
+    ExecutorService executor = Executors.newCachedThreadPool();
+    List<Callable<Boolean>> callables = new ArrayList<>();
```
jamescodeorg, line 26: (very minor) Consider renaming `callables` to something that indicates that it's the list of requests that will fetch files (e.g. `fileLoaders`, `loadRequests`, etc.)

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileManager.java
```diff
@@ -1,15 +1,65 @@
...
+    }
+    List<Future<Boolean>> futures;
+    try {
+      futures = executor.invokeAll(callables);
+      executor.shutdown();
+      boolean finished;
+      finished = executor.awaitTermination(10, TimeUnit.SECONDS);
+
+      if (!finished) {
+        throw new UserFacingException(
```
jamescodeorg, line 40: I kind of expected UserFacingException to only be thrown for errors that are caused by the students code, but now I see that it's more for errors we want to report to the user.  I wonder if it would be useful to add another exception class to separate errors related to student code and reportable errors related to our code.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileManager.java
```diff
@@ -1,15 +1,65 @@
...
+            "We couldn't fetch your files before our loader timed out. Try again.");
```
jamescodeorg, line 41: (minor) I'm guessing we'll revisit all of these error strings at some point, but perhaps "the server" would be better than "our" or "our side"?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileManager.java
```diff
@@ -1,15 +1,65 @@
...
 
-  public UserProjectFileManager(String baseUrl, String[] fileNames) {}
+  public UserProjectFileManager(String baseUrl, String[] fileNames) {
+    this.baseUrl = baseUrl;
+    this.fileNames = fileNames;
+    this.fileList = new ArrayList<>();
+  }
 
+  /**
+   * Asynchronously loads all user project files.
```
jamescodeorg, line 20: This header comment is slightly confusing because the function itself is blocking, right?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/ProjectFileLoader.java
```diff
@@ -0,0 +1,66 @@
...
+      HttpURLConnection con = (HttpURLConnection) url.openConnection();
```
jamescodeorg, line 37: How are we ensuring that a student can only access their own files?  (I was expecting to see us pass a security token of some sort here?)

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileManager.java
```diff
@@ -1,15 +1,65 @@
...
   }
 
+  /**
+   * Returns the user project file. Currently, we only allow a single user project file. TODO:
+   * Enable multi-file programs
+   *
+   * @return the user project file.
+   */
   @Override
   public ProjectFile getFile() {
```
jamescodeorg, line 62: Side comment: I'm not sure if you anticipate adding more functionality to this class, but I think the existing functionality could be handled by a singleton component where loadFiles() just returns the files loaded.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileManager.java
```diff
@@ -1,15 +1,65 @@
...
+  }
 
+  /**
+   * Asynchronously loads all user project files.
+   *
+   * @throws UserFacingException If the load times out or fails.
+   */
   public void loadFiles() throws UserFacingException {
-    throw new UserFacingException("We couldn't load your files");
+    ExecutorService executor = Executors.newCachedThreadPool();
```
jamescodeorg, line 25: Did you want to bound the number of concurrent requests to dashboard?  If so, I think using newFixedThreadPool() might be more appropriate.

molly-moen (approved):
LGTM!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/ProjectFileLoader.java
jmkulwik: ✔️

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileManager.java
jmkulwik: ✔️

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileManager.java
jmkulwik: Yeah, these are placeholders for now.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileManager.java
jmkulwik: ✔️

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/ProjectFileLoader.java
jmkulwik: We'll need to do that for the `PUT` when we implement it later, but for now, we're just using the public `GET` endpoint.

jamescodeorg: Oh, is it by design that we allow anyone to retrieve any student's code?  That seems...  unexpected?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileManager.java
jmkulwik: Hmm what's the concern you're addressing here?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileManager.java
jmkulwik: Yeah, we'd likely add more functionality. i.e. getMainClassFile is one example.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileManager.java
jamescodeorg: If you have lots of files, that would result in a potentially unbounded number of parallel requests to dashboard.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileManager.java
jmkulwik: ✔️

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/ProjectFileLoader.java
jmkulwik: Hmm. Double checking with the infra team.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileManager.java
jmkulwik: Looks like this may end up being a moot point. We just started looking into how we'd handle multiple files, and it's looking like we'll store all files for an individual user project in a single json blob. This will end up being a single request (or possibly two depending on our final design).

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/ProjectFileLoader.java
jmkulwik: Good call. Also, with the way it's looking like we'll change things for multi-file, we may only need to do one HTTP request too.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/ProjectFileLoader.java
jmkulwik: ✔️

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/ProjectFileLoader.java
jmkulwik: Sounds like it's by design. I believe it's a requirement of our project sharing feature (i.e. share between students, share to phone, etc.). More details here: https://codedotorg.slack.com/archives/C03CK49G9/p1617841767304900

PR #15, opened 2021-04-09
Added test framework and initial tests

jmkulwik:
Brings us to ~30% code coverage

There are three classes that don't yet have coverage that definitely need coverage: `UserCodeCompiler`, `UserProjectFileManager`, and `JavaRunner`. There is a follow-up task to set up tests for these classes: https://codedotorg.atlassian.net/browse/CSA-237

image


org-code-javabuilder/lib/src/test/java/org/code/javabuilder/AWSInputAdapterTest.java
```diff
@@ -0,0 +1,63 @@
...
+    sqsMock = mock(AmazonSQS.class);
+    when(sqsMock.receiveMessage(any(ReceiveMessageRequest.class))).thenReturn(result);
+
+    inputAdapter = new AWSInputAdapter(sqsMock, "url");
+  }
+
+  @Test
+  void addsNewlineToMessage() {
+    assertEquals(inputAdapter.getNextMessage(), "hello" + System.lineSeparator());
+    verify(sqsMock, times(2)).deleteMessage("url", null);
```
jamescodeorg, line 42: This verification looks unrelated the main test?  (I see that it has its own test in deletesAllReadMessages() below.)

org-code-javabuilder/lib/src/test/java/org/code/javabuilder/AWSInputAdapterTest.java
```diff
@@ -0,0 +1,63 @@
...
+    when(firstMessage.getBody()).thenReturn("hello");
+    when(secondMessage.getBody()).thenReturn("world");
+    List<Message> messages = new ArrayList<>();
+    messages.add(firstMessage);
+    messages.add(secondMessage);
+
+    ReceiveMessageResult result = mock(ReceiveMessageResult.class);
+    when(result.getMessages()).thenReturn(messages);
+    sqsMock = mock(AmazonSQS.class);
+    when(sqsMock.receiveMessage(any(ReceiveMessageRequest.class))).thenReturn(result);
```
jamescodeorg, line 34: Some of this setup looks test-specific.  I recommend moving anything that is test-specific out to the individual test themselves, adding helper functions to reduce code duplication if needed.

org-code-javabuilder/lib/src/test/java/org/code/javabuilder/AWSInputAdapterTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+    inputAdapter = new AWSInputAdapter(sqsMock, "url");
+  }
+
+  @Test
+  void addsNewlineToMessage() {
+    assertEquals(inputAdapter.getNextMessage(), "hello" + System.lineSeparator());
```
jamescodeorg, line 41: As an example, the setup logic that adds a "hello" message should be in this test method (and not in setUp) because the test logic depends on the message body being "hello".

org-code-javabuilder/lib/src/test/java/org/code/javabuilder/InputRedirectionStreamTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+public class InputRedirectionStreamTest {
+  private InputRedirectionStream stream;
+  private InputAdapter inputAdapter;
+
+  @BeforeEach
+  public void setup() {
+    inputAdapter = mock(InputAdapter.class);
+    when(inputAdapter.getNextMessage()).thenReturn("hello world");
+    stream = new InputRedirectionStream(inputAdapter);
```
jamescodeorg, line 18: As another example, I would consider moving the setup code here into each function (or create a helper) so the arrange-act-assert pattern in each unit test clear.  Alternatively, you can instantiate `stream` and `inputAdapter` here in setup but at least the `when` line should go in each test (and be tailored to be only what is needed by the test).

org-code-javabuilder/lib/src/test/java/org/code/javabuilder/InputRedirectionStreamTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class InputRedirectionStreamTest {
+  private InputRedirectionStream stream;
+  private InputAdapter inputAdapter;
+
+  @BeforeEach
+  public void setup() {
```
jamescodeorg, line 15: Heh, one of the major decisions that you have to make when writing the first unit tests -- is "setup" one word or two?  (It's spelled "setup" here but "setUp" above.)

jamescodeorg (commented):
Sorry for the delay, I'm still catching up on code reviews.

jmkulwik:
@jamescodeorg Thank you for the review! Some other folks are starting to do dev work on Javabuilder - I'm going to merge the portions of this that have no comments in order to unblock test writing on in-process work.

jamescodeorg:
Yes, please feel free to merge some or all of this change.  I know this is all work-in-progress so we don't need to get things perfect before merging.

org-code-javabuilder/lib/src/test/java/org/code/javabuilder/AWSInputAdapterTest.java
jmkulwik: Fixed!

org-code-javabuilder/lib/src/test/java/org/code/javabuilder/AWSInputAdapterTest.java
jmkulwik: Thanks! That cleaned things up a bunch!

org-code-javabuilder/lib/src/test/java/org/code/javabuilder/AWSInputAdapterTest.java
jmkulwik: Fixed!

org-code-javabuilder/lib/src/test/java/org/code/javabuilder/InputRedirectionStreamTest.java
jmkulwik: Thanks! That cleaned things up a bunch!

org-code-javabuilder/lib/src/test/java/org/code/javabuilder/InputRedirectionStreamTest.java
jmkulwik: Haha ok. I standardized :)

org-code-javabuilder/lib/src/test/java/org/code/javabuilder/AWSInputAdapterTest.java
```diff
@@ -0,0 +1,66 @@
...
+  private AWSInputAdapter inputAdapter;
+  private AmazonSQS sqsMock;
+
+  @BeforeEach
+  public void setUp() {
+    sqsMock = mock(AmazonSQS.class);
+    inputAdapter = new AWSInputAdapter(sqsMock, "url");
+  }
+
+  private void messageSetUp(String[] messages) {
```
jamescodeorg, line 26: (minor) A header comment for this method might be useful.  Also, since you are setting precedents here, do you generally want to put helper functions before or after the tests?

jamescodeorg (approved):

molly-moen (approved):
Just a couple places I could use clarifying comments, otherwise looks good!

org-code-javabuilder/lib/src/test/java/org/code/javabuilder/AWSInputAdapterTest.java
jmkulwik: I'm used to putting them before because that's generally what our JS code does. Looking at our Ruby code, it has helpers more rarely, but when they are there, there seems to be a bit of a mix of before and after, possibly more after. Any reasons to favor one way over the other? I think I prefer before so I don't make any assumptions about built-in methods when I'm reading the code, but it's not a strong preference.

jamescodeorg: I think I'm more used to seeing helper methods after the tests but don't know if that's a best practice or just a quirk of codebases I've worked on in the past.  As the person who's blazing the trail here, you get to choose!  :-)

PR #16, opened 2021-04-10
Add a lambda for the API Gateway routes

jmkulwik:
This sets up the lambda for API Gateway. 

All three routes are handled by the same lambda to facilitate sharing common code (this seems to be a common pattern as most of the AWS examples also share one lambda).

Follow these [instructions](https://codedotorg.slack.com/archives/C01EF4GJ9GE/p1616805551042500) on the internal slack to test the lambda.

It is currently hard-coded to run a specific project. This will be updated when we integrate with the Client-side Java Lab.


molly-moen (approved):
looks good to me!

PR #19, opened 2021-04-14
Javabuilder: Load java files from main.json

molly-moen:
As of [this PR](https://github.com/code-dot-org/code-dot-org/pull/40049) we switched from storing Javalab code in individual java files to storing it all in `main.json` under `sources`. This PR updates our Javabuilder logic to pull and parse code from `main.json`. `main.json` is in this format:
```
{
  source: {
     'HelloWorld.java': {
         text: 'public class HelloWorld{...}',
         visible: true
     },...
  }
}
```

### DIFF
--- a/org-code-javabuilder/lib/build.gradle
+++ b/org-code-javabuilder/lib/build.gradle
@@ -35,8 +35,8 @@ task packageSmall(type: Zip) {
 }
 
 java {
-    sourceCompatibility = JavaVersion.VERSION_1_8
-    targetCompatibility = JavaVersion.VERSION_1_8
+    sourceCompatibility = JavaVersion.VERSION_11
+    targetCompatibility = JavaVersion.VERSION_11
 }
 
 build.dependsOn packageSmall
--- a/org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalMain.java
+++ b/org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalMain.java
@@ -11,9 +11,7 @@ public class LocalMain {
   public static void main(String[] args) {
     final LocalInputAdapter inputAdapter = new LocalInputAdapter();
     final LocalOutputAdapter outputAdapter = new LocalOutputAdapter(System.out);
-    // replace with your testing url.
-    String url = "http://localhost-studio.code.org:3000/v3/sources/2sZQjPkr7-vE6zRhbEGTPg";
-    final UserProjectFileManager fileManager = new UserProjectFileManager(url);
+    final LocalProjectFileManager fileManager = new LocalProjectFileManager();
     // Create and invoke the code execution environment
     try (CodeBuilder codeBuilder = new CodeBuilder(inputAdapter, outputAdapter, fileManager)) {
       codeBuilder.compileUserCode();
--- a/org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalMain.java
+++ b/org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalMain.java
@@ -1,9 +1,6 @@
 package dev.javabuilder;
 
-import org.code.javabuilder.CodeBuilder;
-import org.code.javabuilder.InternalFacingException;
-import org.code.javabuilder.UserFacingException;
-import org.code.javabuilder.UserInitiatedException;
+import org.code.javabuilder.*;
 
 /**
  * Intended for local testing only. This is a local version of the Javabuilder lambda function. The
@@ -14,7 +11,9 @@ public class LocalMain {
   public static void main(String[] args) {
     final LocalInputAdapter inputAdapter = new LocalInputAdapter();
     final LocalOutputAdapter outputAdapter = new LocalOutputAdapter(System.out);
-    final LocalProjectFileManager fileManager = new LocalProjectFileManager();
+    // replace with your testing url.
+    String url = "http://localhost-studio.code.org:3000/v3/sources/2sZQjPkr7-vE6zRhbEGTPg";
+    final UserProjectFileManager fileManager = new UserProjectFileManager(url);
     // Create and invoke the code execution environment
     try (CodeBuilder codeBuilder = new CodeBuilder(inputAdapter, outputAdapter, fileManager)) {
       codeBuilder.compileUserCode();
--- a/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileManager.java
+++ b/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileManager.java
@@ -1,5 +1,6 @@
 package org.code.javabuilder;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import java.io.IOException;
 import java.net.URI;
 import java.net.http.HttpClient;
@@ -7,17 +8,20 @@ import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
 import java.time.Duration;
 import java.util.ArrayList;
+import java.util.Map;
 
 /** Manages the in-memory list of user project files. */
 public class UserProjectFileManager implements ProjectFileManager {
   private final String baseUrl;
-  private final String[] fileNames;
   private final ArrayList<ProjectFile> fileList;
+  private final ObjectMapper objectMapper;
 
-  public UserProjectFileManager(String baseUrl, String[] fileNames) {
+  private final String MAIN_SOURCE_FILE_NAME = "main.json";
+
+  public UserProjectFileManager(String baseUrl) {
     this.baseUrl = baseUrl;
-    this.fileNames = fileNames;
     this.fileList = new ArrayList<>();
+    this.objectMapper = new ObjectMapper();
   }
 
   /**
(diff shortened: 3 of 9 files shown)


org-code-javabuilder/lib/build.gradle
```diff
@@ -38,8 +38,8 @@ task packageSmall(type: Zip) {
 }
 
 java {
-    sourceCompatibility = JavaVersion.VERSION_1_8
-    targetCompatibility = JavaVersion.VERSION_1_8
+    sourceCompatibility = JavaVersion.VERSION_11
```
molly-moen, line 41: this change is not related to this PR but fixes some IntelliJ warnings--we should be using Java version 11 everywhere

jmkulwik: 🎉 Thanks for updating this!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileManager.java
```diff
@@ -27,11 +29,10 @@ public UserProjectFileManager(String baseUrl, String[] fileNames) {
    */
   public void loadFiles() throws UserFacingException, UserInitiatedException {
     HttpClient client = HttpClient.newBuilder().build();
-    // TODO: Enable multi-file. For now, we will always have exactly one file.
-    String fileName = fileNames[0];
+    // TODO: Support loading validation code
     HttpRequest request =
         HttpRequest.newBuilder()
-            .uri(URI.create(String.join("/", baseUrl, fileName)))
+            .uri(URI.create(String.join("/", baseUrl, MAIN_SOURCE_FILE_NAME)))
```
jmkulwik, line 35: This join logic originally existed because we were using the files api and didn't know what the user files would be named or how many files there would be. Now that we're using the sources API, the client should go ahead and pass the URL for the main.json in its full form.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileParser.java
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  /**
+   * Parses json string containing file data and stores it in this.fileList.
+   *
+   * @param json JSON String in UserSourceData format
+   * @throws UserFacingException
+   * @throws UserInitiatedException
+   */
+  public List<ProjectFile> parseFileJson(String json)
```
jmkulwik, line 23: NIT: (up to you) This seems like it should be a static function since it's just doing a direct transform of data. Not sure how that would impact testing though.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileManager.java
molly-moen: we will eventually also need to parse validation.json--do you image we will pass both full urls as parameters?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileParser.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class UserProjectFileParser {
+  private final ObjectMapper objectMapper;
+
+  public UserProjectFileParser() {
+    this.objectMapper = new ObjectMapper();
+  }
+
+  /**
+   * Parses json string containing file data and stores it in this.fileList.
```
jmkulwik, line 17: NIT: I think the comment needs to be updated as `this.fileList` isn't being used here.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileManager.java
```diff
@@ -60,7 +59,10 @@ public void loadFiles() throws UserFacingException, UserInitiatedException {
    */
   @Override
   public ProjectFile getFile() {
-    // TODO: Enable multi-file. For now, we will always have exactly one file.
+    // TODO: Enable multi-file. For now, load the first file.
+    if (fileList == null) {
+      return null;
```
jmkulwik, line 64: Trying to think through the scenario where this would happen. Should this throw instead?

org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalMain.java
```diff
@@ -14,7 +11,9 @@
   public static void main(String[] args) {
     final LocalInputAdapter inputAdapter = new LocalInputAdapter();
     final LocalOutputAdapter outputAdapter = new LocalOutputAdapter(System.out);
-    final LocalProjectFileManager fileManager = new LocalProjectFileManager();
+    // replace with your testing url.
+    String url = "http://localhost-studio.code.org:3000/v3/sources/2sZQjPkr7-vE6zRhbEGTPg";
+    final UserProjectFileManager fileManager = new UserProjectFileManager(url);
```
jmkulwik, line 16: I'm working on a minimal websocket implementation of the local lambda function that is designed to work with dashboard. I'm trying to decide whether it makes sense to keep this local main function around. i.e. if we kept it, we'd use it to test javabuilder in isolation. Opinions?
Not blocking for this PR, but I bring it up for this reason: if we keep LocalMain, I think we'd want to keep it using the LocalProjectFileManager and have the websocket implementation be the one that reaches out to dashboard for its project.

org-code-javabuilder/lib/src/test/java/org/code/javabuilder/UserProjectFileParserTest.java
```diff
@@ -0,0 +1,33 @@
...
+public class UserProjectFileParserTest {
+  private UserProjectFileParser userProjectFileParser;
+
+  @BeforeEach
+  public void setup() {
+    this.userProjectFileParser = new UserProjectFileParser();
+  }
+
+  @Test
+  public void canParseValidFileJson() throws UserFacingException, UserInitiatedException {
```
jmkulwik, line 18: NIT: Consider adding tests for the failure conditions as well

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileParser.java
molly-moen: I think in general static functions are harder to test. For this case, the main reason it's not static is so we can share an ObjectMapper instance across multiple calls to parseFileJson

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileManager.java
molly-moen: Good question, I also wasn't sure about this. The only reasons it would occur is if the Javabuilder code was calling the functions incorrectly (calling get before load), or the load failed. If the load failed we would throw an exception earlier to the end user. It felt like a lot of exception declaration to add a throw here, but I can switch it if that is our preferred pattern. Thinking about it more this doesn't cover the case where the 'code' loaded successfully but it was an empty object--then we will get an index out of bound exception.

org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalMain.java
molly-moen: I don't have a strong preference beyond keeping the friction relatively low to testing locally. I can update LocalProjectFileManager to use the new parser instead. I found it much easier to call out to a url because it's a pain to format java code into a string manually, so I would rather have dashboard do it for me.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileManager.java
molly-moen: this function will be obsolete once we enable multi-file, since we will switch to returning the entire list, so I would lean towards not adding complexity here.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileManager.java
jmkulwik: Changing this comment to non-blocking. There's an [offline conversation](https://codedotorg.slack.com/archives/C01EF4GJ9GE/p1618437424028900) regarding validation.json going on in slack which shouldn't block this PR.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileParser.java
jmkulwik: 👍

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileManager.java
jmkulwik: We'd want to keep what's happening clear to the user and the developer. Exceptions have the benefit of being pretty verbose and customizable as to what's happening. That doesn't mean we need to use them in every case.

org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalMain.java
jmkulwik: 👍 I'd say if you're going to keep using this pattern in your local testing, leave it as-is. If you're done, change it back. Once the websocket change is ready, the local websocket server will be the default way to do this.

molly-moen: for multi-file, since we don't have it implemented on the javalab side yet I'll probably have to switch back to using local and will update LocalProjectFileManager to use the new file format. I can switch it back for now, although as of now LocalProjectFileManager does not exercise the new json parsing logic.

molly-moen: switched back since I am done using this. In the follow up to this (support running multi-file) I'll update the LocalProjectFileManager to use a main.json instead of MyClass.java

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileManager.java
jmkulwik: Sorry, my browser must have had this thread cached, and I didn't see your most recent comment. 
✔️  Sounds good to me!

jmkulwik (approved):
Looks good!

PR #20, opened 2021-04-14
Set up local testing server for integration with dashboard

jmkulwik:
This sets up a minimal local websocket server so developers can work on code studio/dashboard at the same time as javabuilder.

Follow-up work on dashboard in order to use this change is here: https://github.com/code-dot-org/code-dot-org/pull/40116


org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketServer.java
```diff
@@ -0,0 +1,85 @@
...
+   */
+  @OnOpen
+  public void onOpen(Session session) {
+    // Temporary project url until the dashboard side is integrated
+    String projectUrl = "http://localhost-studio.code.org:3000/v3/files/MoGwvaZmQSQZaImS3LliCQ";
+    String[] fileNames = new String[] {"MyClass.java"};
+
+    outputAdapter = new WebSocketOutputAdapter(session);
+    inputAdapter = new WebSocketInputAdapter();
+    final UserProjectFileManager fileManager = new UserProjectFileManager(projectUrl, fileNames);
```
molly-moen, line 40: my recently merged PR changed the constructor for UserProjectFileManager, so you'll need to update this to no longer include filenames

molly-moen (commented):
Clarifying question: we have some follow-up work to do on dashboard to make this work for local development, correct?

jmkulwik:
> Clarifying question: we have some follow-up work to do on dashboard to make this work for local development, correct?

Yes. I'll go ahead and link the draft PR where I'm working on the dashboard change.

org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketServer.java
jmkulwik: Fixed. Thanks!

molly-moen (approved):
Looks good to me!

PR #21, opened 2021-04-15
Javabuilder: support multi-file projects

molly-moen:
This PR enables Javabuilder to parse, compile and run projects that contain multiple files. Of the files, there must be exactly one main method, otherwise we will throw an error when trying to run the project. See [this PR](https://github.com/code-dot-org/java-ide/pull/19) for details on the expected input format for parsing the files. Currently we only support java files, we will need to support other text files as a follow-up.

This PR also updates LocalProjectFileManager to only support the new json file format for input, and includes a couple sample main.jsons for multi-file Java projects.


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaRunner.java
```diff
@@ -61,4 +54,43 @@ public void runCode()
...
+
+  /**
+   * Finds the main method in the set of files in fileManager if it exists.
+   *
+   * @param urlClassLoader class loader pointing to location of compiled classes
+   * @return the main method if it is found
+   * @throws UserFacingException if there is an issue loading a class
+   * @throws UserInitiatedException if there is more than one main method or no main method
+   */
+  public Method findMainMethod(URLClassLoader urlClassLoader)
```
molly-moen, line 66: I think there would be value in having unit tests on JavaRunner, but any unit test I attempted ended up being mostly mocking objects and checking I returned the right mock, which did not seem helpful. Any tips on best practices here would be helpful--are unit tests useful for something like this that requires loading already compiled code, or do we need some sort of integration test?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaRunner.java
```diff
@@ -61,4 +54,43 @@ public void runCode()
...
+          throw new UserInitiatedException(
+              "Your code can only have one main method. We found at least two classes with main methods.");
+        } else {
+          mainMethod = tempMain;
+        }
+      } catch (ClassNotFoundException e) {
+        // this should be caught earlier in compilation
+        throw new UserFacingException(
+            "We hit an error on our side while running your program. Try Again", e);
+      } catch (NoSuchMethodException e) {
```
jmkulwik, line 86: Using an exception for normal code flow is a bit confusing semantically - i.e. I'd normally expect to see some sort of error here. It also can have performance impacts, depending on the JVM and compiler. It's also Looks like `Class` also has `getDeclaredMethods` which returns an array of methods. That could prevent this exception from being hit after each non-main class is searched.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/ProjectFileManager.java
```diff
@@ -12,8 +14,6 @@
    */
   void loadFiles() throws UserFacingException, UserInitiatedException;
 
-  /**
-   * @return The user's Java file. Currently there is only one file that can be used or returned.
-   */
-  ProjectFile getFile();
+  /** @return The user's Java files. */
+  List<ProjectFile> getFiles();
```
jmkulwik, line 18: 🎉

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserCodeCompiler.java
```diff
@@ -59,14 +61,16 @@ private CompilationTask getCompilationTask(DiagnosticCollector<JavaFileObject> d
     }
 
     // create file for user-provided code
```
jmkulwik, line 63: Tiny NIT: file -> files

org-code-javabuilder/lib/src/test/java/org/code/javabuilder/UserProjectFileParserTest.java
```diff
@@ -39,4 +39,26 @@ public void throwsExceptionOnInvalidJson() throws UserFacingException, UserIniti
...
+            + "      \"text\": \"import java.util.Scanner;\\n\\npublic class ScannerSample {\\n\\tpublic static void main(String[] args) {\\n\\t\\tHelperClass.printWelcomeMessage();\\n\\t\\tScanner console = new Scanner(System.in);\\n\\t\\tSystem.out.println(\\\"Enter your name: \\\");\\n\\t\\tString name = console.next();\\n\\t\\tSystem.out.println(\\\"Hello \\\" + name);\\n\\t}\\n}\\n\",\n"
+            + "      \"visible\": true\n"
+            + "    },\n"
+            + "    \"HelperClass.java\": {\n"
+            + "      \"text\": \"\\npublic class HelperClass {\\n\\tpublic static void printWelcomeMessage() {\\n\\t\\tSystem.out.println(\\\"Hello!\\\");\\n\\t}\\n}\\n\",\n"
+            + "      \"visible\": true\n"
+            + "    }\n"
+            + "  },\n"
+            + "  \"animations\": {}\n"
+            + "}";
```
jmkulwik, line 58: Suggestion: Since this test is testing the file names and size, omit unneeded sections. i.e. `text` could be something short like `"my code"`.

org-code-javabuilder/lib/src/main/resources/main_painter.json
```diff
@@ -0,0 +1,25 @@
+{
+  "source": {
```
jmkulwik, line 2: 🎉

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaRunner.java
jmkulwik: Not sure if this is what you're referring to: I could see it being useful to test that this always throws a `UserInitiatedException` if no main is found or if 2 mains are found.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaRunner.java
molly-moen: good find!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaRunner.java
molly-moen: I agree, although to write a unit test that does that I have to mock basically everything, which doesn't seem like the most useful test to me.

jmkulwik (approved):
LGTM!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaRunner.java
molly-moen: maybe? Part of the problem I'm having is we have multiple related prerequisites for running JavaRunner (classes must be both defined in the fileManager and compiled to the executable location). I can add a note to the existing task to add tests for JavaRunner that we may need to do some refactoring.

PR #22, opened 2021-04-16
Updated lambdas to use query string parameters instead of headers

jmkulwik:
Browser websocket APIs cannot use HTTP headers the way that commandline utils like wscat can. Because of this, tokens are generally passed in the [query string parameters](https://www.freecodecamp.org/news/how-to-secure-your-websocket-connections-d0be0996c556/). 

[Slack discussion](https://codedotorg.slack.com/archives/C03CK49G9/p1618521915378900)

### Security Considerations
The header and query string are both encrypted in HTTPS and WSS traffic. The query string is slightly more vulnerable because the encrypted URL is more likely to be recorded in network logs. Given the short-lived nature of the tokens, this is unlikely to affect our security posture. 

Token maximum lifespan is 15 minutes. We'd likely issue one token per connection to javabuilder. Due to the AWS limits on asynchronous lambdas, javabuilder would only ever be able to run for 15 minute sessions. Given the requirements from curriculum, we'd likely limit this to an even shorter time of 5 minutes. The token would have the same lifespan of javabuilder.


api-gateway-routes/api_gateway_proxy_function.rb
```diff
@@ -36,7 +33,7 @@ def on_connect(event, context)
     :queueUrl => sqs_queue.queue_url,
     :apiEndpoint => api_endpoint,
     :connectionId => request_context["connectionId"],
-    :projectUrl => project_url,
+    :projectUrl => request_context["authorizer"]["project_url"],
```
molly-moen, line 36: currently the token does not contain the project_url--do we need to add it? It does contain the channel_id

jmkulwik: For now, at least, I'd like to pass in the project_url here. I've got a follow-up PR on dashboard to do that. There's pretty much one place where we build the URLs for accessing the S3 API, so if we ever update it, I'd like to get that automatically in javabuilder.

molly-moen (approved):
Glad this was a quick update!

PR #23, opened 2021-04-19
Provision Javabuilder CodePipeline Continuous Integration Stack

sureshc:
Provision a pipeline that builds/updates a Javabuilder staging system anytime new commits are added to the `main` branch in the GitHub repository. Enable an Engineer to manually Approve deploying that commit to the production Javabuilder instance.


.ruby-version
```diff
@@ -1 +0,0 @@
-2.7.4
```
Hamms, line 1: why was this removed?

deploy-continuous-integration.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash -xe
```
Hamms, line 1: is the print flag intentional here?

continuous-integration.yml
```diff
@@ -0,0 +1,208 @@
...
+#        SecretToken: !Ref AWS::AccountId
+#      Filters:
+#        - JsonPath: "$.ref"
+#          MatchEquals: refs/heads/{Branch}
+#      Authentication: GITHUB_HMAC
+#      TargetPipeline: !Ref Pipeline
+#      TargetAction: Source
+#      Name: javabuilder-webhook
+#      TargetPipelineVersion: !GetAtt Pipeline.Version
+#      RegisterWithThirdParty: true
```
Hamms, line 169: stray comment?

Hamms (approved):
I flagged a couple things here that jump out at me; everything else looks basically like I'd expect it to, but I don't have enough context here to be confident in all the details. But from what little I do know, this looks good to me!

continuous-integration.yml
sureshc: Oops. This is still a bit of a work in progress. I commented out the GitHub Webhook temporarily to get the rest of the pipeline working. I need to figure out how to get this working.

deploy-continuous-integration.sh
sureshc: Not sure! I copied/pasted [from the p5-replay project](https://github.com/code-dot-org/p5-replay/blob/d26c5b120cf1799a089237918045c3ade794945b/deploy-ci.sh#L1).

.ruby-version
sureshc: This was somehow incompatible with AWS CodeBuild and the Ruby 2.7 runtime it uses. CodeBuild didn't like us specifying a patch version (.4). Now that the rest of the Pipeline is working, I'm going to try and get this rbenv configuration integrated back into the project.

deploy-continuous-integration.sh
Hamms: Hah, gotcha. I think it makes sense to keep around in a script file that's only doing a few things, but note this flag can quickly become the source of a whole lot of noise as scripts grow

deploy-continuous-integration.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# Deploys the Continuous Integration service CloudFormation stack.
```
cat5inthecradle, line 3: If you're feeling extra autonomous, you can make _another_ cloudformation stack that deploys a CodePipeline pipeline which is responsible for updating the `javabuilder-continuous-integration` stack. If you do this, then you can edit the "continuous-integration.yml" file, merge to main, and have it automatically update the CI pipeline. 

The new "CI-Deployer" stack & pipeline is incredibly unlikely to ever need updating, whereas the javabuilder-continuous-integration stack will see changes from time to time.

The advantage is that we eliminate a scenario where a developer needs powerful 'deployer' level permissions.

One gotcha is that you want to avoid merging CI and App updates in the same merge, because you'll have a race condition where the pipeline is being updated at the same time the app is being deployed. I have not looked into ways to avoid this or make it safer. It is possible to have your App pipeline run where early steps are defined by the previous commit, and later steps are defined by the new commit. I've always done more frequent, smaller deploys to production so it was rarely an issue.

cat5inthecradle:
Thanks for paving the way! Closing this PR as the pipeline has now been implemented 🎉

PR #26, opened 2021-04-21
Javabuilder: Support text files

molly-moen:
This PR enables us to support Java programs that have text files in their startup code and/or that read from or write to text files. Previously we would throw an exception for non-Java files; now any non-Java file is treated as a text file and written to the working directory.

In order to enable this feature, we have added a [layer](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-modify.html#runtime-wrapper) to our lambda that changes the working directory to be within the lambda `/tmp` folder. This allows file reads and writes to occur without students or curriculum writers needing to specify reading from/writing to the tmp folder. 

## Future Work
If user code creates or edits a text file in their code, we don't do anything with it and it will be deleted. We will need to add support for returning this file to the user somehow.


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileParser.java
```diff
@@ -21,20 +20,25 @@ public UserProjectFileParser() {
...
       throws UserFacingException, UserInitiatedException {
-    List<ProjectFile> fileList = new ArrayList<ProjectFile>();
     try {
       UserSourceData sourceData = this.objectMapper.readValue(json, UserSourceData.class);
       Map<String, UserFileData> sources = sourceData.getSource();
       for (String fileName : sources.keySet()) {
         UserFileData fileData = sources.get(fileName);
-        fileList.add(new ProjectFile(fileName, fileData.getText()));
+
+        if (fileName.indexOf(".java") > 0) {
```
jmkulwik, line 32: May want to update this to check if the file ends in java. i.e. I've seen people do myFile.java.txt before.

jmkulwik (approved):
Looks good to me!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileParser.java
```diff
@@ -21,20 +20,25 @@ public UserProjectFileParser() {
    * @throws UserFacingException
    * @throws UserInitiatedException
    */
-  public List<ProjectFile> parseFileJson(String json)
+  public void parseFileJson(
```
jamescodeorg, line 23: I recommend creating a class for the return value here (instead of passing in lists that are filled in by this method).  This is more idiomatic and extensible and I think this class will be valuable -- it is a structured representation of all the assets in a student's workspace.

(Also, the @returns in the header comment should be updated.)

jamescodeorg (commented):
Sorry for the late comments.  A few suggestions for you to think about as you continue to work on this code.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFileParser.java
molly-moen: made a follow-up PR to address these comments here! https://github.com/code-dot-org/java-ide/pull/28

PR #28, opened 2021-04-23
Javabuilder: Refactor load and build steps

molly-moen:
Follow-up to comments on [26](https://github.com/code-dot-org/java-ide/pull/26). Now that the load and build steps include non-Java files which are handled differently in our build process, we needed to do some refactoring to clean up the code paths. No functionality has changed, just moving code around. This PR does the following:
- Remove the `FileManager` class and replace with a `FileLoader` that is environment (local, cloud) specific. The `FileLoader` load method returns a new class, `UserProjectFiles` which contains all text and Java files, and can be extended in the future to handle other file types (audio, video, etc.).
- `CodeBuilder`'s `compileUserCode` method has been retooled to be `buildUserCode`, and will now save text files and compile Java files
- `UserCodeCompiler` and `JavaRunner` now only need to access the list of `JavaProjectFile`
- some minor cleanup--use `this.` for all class field access and removed an unused class field.


org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalProjectFileLoader.java
```diff
@@ -0,0 +1,24 @@
...
+import org.code.javabuilder.*;
+
+/** Intended for local testing only. Loads the main.json file from the resources folder. */
+public class LocalProjectFileLoader implements ProjectFileLoader {
+
+  @Override
+  public UserProjectFiles loadFiles() throws UserFacingException, UserInitiatedException {
+    try {
+      String mainJson =
+          new String(
```
jamescodeorg, line 16: Maybe use `Files.readString()`?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeBuilder.java
```diff
@@ -1,31 +1,29 @@
...
   private final InputAdapter inputAdapter;
-  private final ProjectFileManager fileManager;
   private final File tempFolder;
   private final PrintStream sysout;
   private final InputStream sysin;
+  private final UserProjectFiles userProjectFiles;
 
   public CodeBuilder(
-      InputAdapter inputAdapter, OutputAdapter outputAdapter, ProjectFileManager fileManager)
+      InputAdapter inputAdapter, OutputAdapter outputAdapter, UserProjectFiles userProjectFiles)
```
jamescodeorg, line 20: This is a nice improvement!  Having CodeBuilder take a "data" object makes it more decoupled and easier to test and reuse.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeBuilder.java
```diff
@@ -85,4 +86,19 @@ public void close() throws InternalFacingException {
...
     }
   }
+
+  /** Save any non-source code files to storage */
+  private void saveProjectAssets() throws UserFacingException {
+    // Save all text files to current folder.
+    List<TextProjectFile> textProjectFiles = this.userProjectFiles.getTextFiles();
+    for (TextProjectFile projectFile : textProjectFiles) {
+      String filePath = projectFile.getFileName();
+      try (PrintWriter out = new PrintWriter(filePath)) {
```
jamescodeorg, line 96: Not sure if it's easier or not, but `Files.writeString` might be an option here.

jamescodeorg (approved):
Nice improvements!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeBuilder.java
molly-moen: updated!

org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalProjectFileLoader.java
molly-moen: updated!

jmkulwik (approved):
Awesome cleanup.

PR #29, opened 2021-04-24
Add support for a public websocket connectivity test

jmkulwik:
This sets up a public Javabuilder endpoint for the purposes of testing a school's websocket support. The endpoint does not give access to the long-running lambda that executes code.

This supports a similar PR on the code-dot-org repo: https://github.com/code-dot-org/code-dot-org/pull/40250

Some details from a [slack conversation about this change](https://codedotorg.slack.com/archives/C03CK49G9/p1619134069451500)

### DIFF
--- a/api-gateway-routes/api_gateway_proxy_function.rb
+++ b/api-gateway-routes/api_gateway_proxy_function.rb
@@ -1,6 +1,7 @@
 require 'json'
 require 'aws-sdk-sqs'
 require 'aws-sdk-lambda'
+require 'aws-sdk-apigatewaymanagementapi'
 require 'uri'
 
 def lambda_handler(event:, context:)
@@ -21,6 +22,11 @@ def on_connect(event, context)
   # Add in some logging to make debugging easier
   puts request_context
 
+  # Return early if this is the user connectivity test
+  authorizer = request_context['authorizer']
+  if authorizer && authorizer['connectivityTest']
+    return { statusCode: 200, body: "connection successful" }
+  end
   # -- Create SQS for this session --
   sqs_client = Aws::SQS::Client.new(region: region)
   sqs_queue = sqs_client.create_queue(
@@ -30,12 +36,12 @@ def on_connect(event, context)
 
   # -- Create Lambda for this session --
   lambda_client = Aws::Lambda::Client.new(region: region)
-  api_endpoint = "https://#{request_context['apiId']}.execute-api.#{region}.amazonaws.com/#{request_context['stage']}"
+  api_endpoint = get_api_endpoint(event, context)
   payload = {
     :queueUrl => sqs_queue.queue_url,
     :apiEndpoint => api_endpoint,
     :connectionId => request_context["connectionId"],
-    :projectUrl => request_context["authorizer"]["project_url"]
+    :projectUrl => authorizer["project_url"]
   }
   response = lambda_client.invoke({
     function_name: 'javaBuilderExecuteCode:4',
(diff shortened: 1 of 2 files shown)


api-gateway-routes/api_gateway_proxy_function.rb
```diff
@@ -54,8 +60,21 @@ def on_disconnect(event, context)
 end
 
 def on_default(event, context)
-  sqs = Aws::SQS::Client.new(region: get_region(context))
   message = event["body"]
+  # Return early if this is the user connectivity test
+  if message == 'connectivityTest'
```
molly-moen, line 65: How risky is it that a user will just hold open a websocket via the connectivity test? Will that cause any issues for us?

api-gateway-routes/api_gateway_proxy_function.rb
```diff
@@ -21,6 +22,11 @@ def on_connect(event, context)
   # Add in some logging to make debugging easier
   puts request_context
 
+  # Return early if this is the user connectivity test
+  authorizer = request_context['authorizer']
+  if authorizer && authorizer['connectivityTest']
+    return { statusCode: 200, body: "connection successful" }
```
sureshc, line 28: My knowledge of websockets is limited. Does this return an HTTP 200 response to the initial HTTP Upgrade request? Can we also send a message on the websocket to the browser and require the client to wait for that message to validate that bi-directional messaging is working? I’m guessing that the `$connect` route can’t do that.

jmkulwik: Take a look at the `default` route (below) and the client-side part of the test: https://github.com/code-dot-org/code-dot-org/pull/40250

With those two, we make sure the bi-directional message also happens.

api-gateway-routes/api_gateway_proxy_function.rb
sureshc: Maybe we can have the client explicitly disconnect the WebSocket after it has confirmed bidirectional connectivity and also we could explicitly disconnect the connectivity test WebSocket from API Gateway and/or a Lambda if the client doesn't close it within a configured time out?

api-gateway-routes/api_gateway_proxy_function.rb
sureshc: Never mind. I see that the default route supports a bidirectional exchange of messages!

api-gateway-routes/api_gateway_proxy_function.rb
jmkulwik: If a user just holds the WebSocket open, there wouldn't really be any effect in usability:
"API Gateway doesn't enforce a quota on concurrent connections. The maximum number of concurrent connections is determined by the rate of new connections per second and maximum connection duration of two hours. For example, with the default quota of 500 new connections per second, if clients connect at the maximum rate over two hours, API Gateway can serve up to 3,600,000 concurrent connections."

Looking at cost:
* WebSocket APIs, the API Gateway free tier includes one million messages (sent or received) and 750,000 connection minutes for up to 12 months.
* $0.25 per million connection minutes
* $1/million messages

One mitigation: The client does explicitly disconnect after a successful connectivity test or an error message.

API Gateway has some default timeouts:
```
Idle Connection Timeout | 10 minutes 
Connection duration for WebSocket API | 2 hours 
```
Would we want to make these tighter?

molly-moen: might as well decrease the connection duration since the lambda cannot run for more than 15 minutes anyway. But sounds like it's pretty low risk overall

molly-moen (approved):

api-gateway-routes/api_gateway_proxy_function.rb
sureshc: Looks like the API Gateway idle WebSocket timeout is 10 minutes, so we have some basic protection against overconsumption of server resources: https://docs.aws.amazon.com/apigateway/latest/developerguide/limits.html#apigateway-execution-service-websocket-limits-table

sureshc (approved):

PR #31, opened 2021-05-06
Student Facing Painter API

JillianK:
Create a new org.code.neighborhood package that contains the student facing painter API and tests.
The following requirements are met:
- Users must be able to interact with the API by creating a new Painter object without taking action to create an instance of the grid object
- Multiple painter objects must be able to coexist and use the same grid object
- Users must be able to use the methods that are specified on the Painter Interface in the Student Facing API

[Neighborhood plan](https://docs.google.com/document/d/1SqSLogXsOKJD8QexbZAbxCpUGRdNP9Yj7zW8tN6oADI/edit)
[Student Facing API mockup](https://docs.google.com/document/d/1DGF3Y3pBVbB1H5G9aXownsok6otQYcxs_Ow1FIy4Za0/edit)
[jira ticket](https://codedotorg.atlassian.net/browse/CSA-290?atlOrigin=eyJpIjoiOTE2ZTU5YzhjYmMyNDdmY2JlYzlmMzNiZDljNDZkZTgiLCJwIjoiaiJ9)


org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Grid.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.ArrayList;
+import java.io.File;  // Import the File class
+import java.io.FileInputStream; // Import the Scanner class to read text files
+import java.lang.System;
+import org.json.simple.JSONArray;
+import org.json.simple.JSONObject;
+import org.json.simple.parser.JSONParser;
+import org.json.simple.parser.ParseException;
+
+public class Grid {
```
jamescodeorg, line 12: I'm not sure but I think it might help to have a `Level` or `World` object that owns both a grid and a list of painters.  This might be easier to work with than the static singleton instance.

You might still need a singleton (global?) instance of Level, but that object should be much simpler and just be a container for the most part.

To make it easier to unit test, consider making Grid an interface, especially if we think that we might expose this interface to students at some point.

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Grid.java
```diff
@@ -0,0 +1,114 @@
...
+                System.out.println("\nTry adding a grid.txt file.\n");
+                e.printStackTrace();
+                System.exit(1);
+            }
+        }
+        return instance;
+    }
+
+    private Grid(String gridDescription) {
+        String[] lines = gridDescription.split("\n");
```
jamescodeorg, line 40: It might be nice to separate out how a grid is created from what a grid does.  I'm imagining a constructor that takes just GridSquare[][].  There would then be a separate static constructor (static method that returns a Grid instance) that creates a Grid given a filename/string.  If you wanted further separation, that code could live in a separate class called `GridFactory`.  (Among other things, this means the Grid class would not need to import any file related dependencies.)

If we do all that, then unit testing the functionality in Grid becomes straightforward -- you create a Grid manually using the constructor and hand-crafting a GridSquare[][].  Then you can separately unit test the static constructor.

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/GridSquare.java
```diff
@@ -0,0 +1,119 @@
+package org.code.neighborhood;
+import org.json.simple.JSONObject;
+
+public class GridSquare {
+    private String color;
+    private boolean passable;
+    private int paintCount;
+    private enum SquareType{ WALL, OPEN, START, FINISH, OBSTACLE, STARTANDFINISH, UNKNOWN};
```
jamescodeorg, line 8: Did we consider having isStart and isFinish be flags instead of SquareTypes?  It would remove some of the combinatorial things going on in this enum and also simplify the logic for methods like isPassable.

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/GridSquare.java
```diff
@@ -0,0 +1,119 @@
...
+    private SquareType squareType;
+
+    public GridSquare(JSONObject descriptor) {
```
jamescodeorg, line 11: This class would be more decoupled and more unit testable if the constructor took primitive types instead of the JSONObject.

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/GridSquare.java
```diff
@@ -0,0 +1,119 @@
...
+                case 5:
+                    this.squareType = SquareType.STARTANDFINISH;
+                    this.passable = true;
+                    break;
+                default:
+                    this.squareType = SquareType.UNKNOWN;
+                    this.passable = false;
+            }
+        } catch (NumberFormatException e) {
+            System.out.println("\nPlease check the format of your grid.txt file, especially the tileType\n");
```
jamescodeorg, line 44: This is a good example -- If the constructor took (string tileType, int value), this code would move to the caller which will likely contain all of the code related to parsing the file format and catching/handling formatting errors.

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/GridSquare.java
```diff
@@ -0,0 +1,119 @@
...
+        }
+    }
+
+    // Determines whether the given coordinate can be moved into
+    public boolean isPassable() {
+        return this.passable;
+    }
+
+    // Decreases the paintCount by 1 if there is available paint
+    public void collectPaint() {
```
jamescodeorg, line 78: I kind of expected this to return Color or something that represents the paint collected.

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/GridSquare.java
```diff
@@ -0,0 +1,119 @@
...
+        if (this.containsPaint()) {
+            this.paintCount--;
+        } else {
+            System.out.println("There's no paint to collect here");
+        }
+    }
+
+    // Returns the square to a non-painted state
+    public void removePaint() {
```
jamescodeorg, line 87: Is this function the inverse of setColor?  If so, maybe we can name it removeColor or unsetColor?

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/GridSquare.java
```diff
@@ -0,0 +1,119 @@
...
+                e.printStackTrace();
+                System.exit(1);
+            }
+        } else {
+            this.paintCount = 0;
+        }
+    }
+
+    // Sets the color of the square to the given color
+    public void setColor(String color) {
```
jamescodeorg, line 62: I'm wondering if there should be a Color enum, at least internally?  (Java enums aren't the easiest to work with and might not be something we want to introduce to students for awhile?)

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Painter.java
```diff
@@ -0,0 +1,144 @@
+package org.code.neighborhood;
+
+public class Painter {
+    private int xLocation;
+    private int yLocation;
+    private Direction direction;
+    private int remainingPaint;
```
jamescodeorg, line 7: Just curious, do we track the color of the paint that the painter has picked up?  Or is there only one color of paint?

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Painter.java
```diff
@@ -0,0 +1,144 @@
...
+        this.remainingPaint = paint;
+        this.grid = Grid.getInstance();
+        this.id = "Painter-"+this.grid.registerPainter();
+
+        System.out.println("created painter");
+    }
+
+    // Turn the painter one compass direction left (i.e. North -> West)
+    public void turnLeft() {
+        this.direction.turnLeft();
```
jamescodeorg, line 24: If we make direction itself an enum, this line would be something like `this.direction = this.direction.turnLeft()`.

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Painter.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    // Move the painter one square forward in the direction the painter
+    // is facing
+    public void move() {
```
jamescodeorg, line 29: I think moveForward() is a little bit clearer but this is probably part of the interface that can't be changed anymore?

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Painter.java
```diff
@@ -0,0 +1,144 @@
...
+    private int xLocation;
+    private int yLocation;
+    private Direction direction;
+    private int remainingPaint;
+    private Grid grid;
+    private String id;
+
+    public Painter(int x, int y, String direction, int paint) {
+        this.xLocation = x;
+        this.yLocation = y;
```
jamescodeorg, line 13: Suggestion: When working with grids, I often see a `Coord` or `Position` type defined that might make some of the code in this class a bit more readable.  (For example, instead of referencing `xLocation` and `yLocation`, you would reference `position.X`.)

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Painter.java
```diff
@@ -0,0 +1,144 @@
...
+            this.grid.getSquare(this.xLocation, this.yLocation).collectPaint();
+            this.remainingPaint++;
+        } else {
+            System.out.println("There is no paint to collect here");
+        }
+    }
+
+    // Returns True if there is paint in the square where the painter
+    // is standing.
+    public Boolean isOnPaint() {
```
jamescodeorg, line 87: I'm finding the method names very confusing.  A square can contain a bucket of paint or be painted a certain color.  The painter can also paint a square.  All three of these are referred to using the word 'paint' in different contexts.  I think it would be helpful to use different words consistently for each of these three things.  (This could make a big difference for students with less fluency in English.)

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Painter.java
```diff
@@ -0,0 +1,144 @@
...
+        return this.grid.getSquare(this.xLocation, this.yLocation).hasColor();
+    }
+
+    // Returns True if there is a paint bucket in the square where the
+    // painter is standing.
+    public Boolean isOnBucket() {
```
jamescodeorg, line 93: Should these methods all return `boolean` instead of `Boolean`?

jamescodeorg (commented):
Exciting to see this come together!

As always, feel free to accept/ignore as many of the comments as you'd like.  (I have no expectations as a "guest" reviewer that comments have to be incorporated.)

Also, I know that this was just a draft and I tried to give more structural suggestions (although some minor comments about naming, etc. did sneak in).

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Painter.java
jmkulwik: @hanwalden ^

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Painter.java
jmkulwik: Conceptually, the paint coloring is "magic" (I believe that's the official term 🙂) - it doesn't have a color associated with it until it is used to paint a square.

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Painter.java
jmkulwik: @hanwalden ^

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Painter.java
jamescodeorg: Love it!  My kids want to paint their rooms, where can I get some of this paint?  :-)

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/GridFactory.java
```diff
@@ -0,0 +1,66 @@
...
+import org.json.simple.JSONObject;
+import org.json.simple.parser.JSONParser;
+import org.json.simple.parser.ParseException;
+
+import java.io.File;
+import java.io.FileInputStream;
+
+public class GridFactory {
+    protected Grid createGridFromJSON(String filename) {
+        File file = new File("grid.txt");
```
molly-moen, line 13: nit: add a constant for "grid.txt"

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/GridFactory.java
```diff
@@ -0,0 +1,66 @@
...
+            JSONArray gridSquares = (JSONArray) obj;
+            int height = gridSquares.size();
+            if(height == 0) {
+                throw new UnsupportedOperationException("Please check the format of your grid.txt file");
+            }
+            int width = ((JSONArray) gridSquares.get(0)).size();
+            GridSquare[][] grid = new GridSquare[width][height];
+            int currentHeight = height;
+            for (int currentY = 0; currentY < height; currentY++) {
+                currentHeight--;
```
molly-moen, line 41: can you add some comments in here about the expected format of the `description` string and why we are starting from the maxHeight instead of 0?

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/GridFactory.java
```diff
@@ -0,0 +1,66 @@
...
+                JSONArray line = (JSONArray) gridSquares.get(currentHeight);
+                if (line.size() != width) {
+                    throw new UnsupportedOperationException("width of line " + currentHeight + " does not match others. Cannot create grid.");
+                }
+                for (int currentX = 0; currentX < line.size(); currentX++) {
+                    JSONObject descriptor = (JSONObject) line.get(currentX);
+                    try {
+                        int tileType = Integer.parseInt(descriptor.get("tileType").toString());
+                        if(descriptor.containsKey("value")) {
```
molly-moen, line 50: can we add constants for "value" and "tileType"

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Painter.java
```diff
@@ -0,0 +1,144 @@
...
+    private String id;
+
+    public Painter(int x, int y, String direction, int paint) {
+        this.xLocation = x;
+        this.yLocation = y;
+        this.direction = Direction.fromString(direction);
+        this.remainingPaint = paint;
+        this.grid = World.getInstance().getGrid();
+        this.id = "Painter-" + lastId++;
+        System.out.println("created painter");
```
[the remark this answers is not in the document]
JillianK: We don't need it apparently so I'll take it out

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/GridFactory.java
```diff
@@ -9,8 +9,11 @@
 import java.io.FileInputStream;
 
 public class GridFactory {
+    private static final String gridFileName = "grid.txt";
```
molly-moen, line 12: nit: style for constants is ALL_CAPS

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Helpers.java
```diff
@@ -0,0 +1,16 @@
...
+
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class Helpers {
+    private static final String HEX_WEB_COLOR_PATTERN
+            = "^#([a-fA-F0-9]{6}|[a-fA-F0-9]{3})$";
+    private static final Pattern PATTERN = Pattern.compile(HEX_WEB_COLOR_PATTERN);
+    private static final Set<String> WEB_COLORS = Set.of("white", "silver", "gray", "black", "red", "maroon", "yellow", "olive", "lime", "green", "aqua", "teal", "blue", "navy", "fuchsia", "purple");
```
molly-moen, line 11: should we add a comment/todo here to add the 'extended' colors or are we sticking with the basic colors?

molly-moen (approved):
🎉

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Painter.java
```diff
@@ -0,0 +1,146 @@
...
+    // Returns True if there is paint in the square where the painter
+    // is standing.
+    public boolean isOnPaint() {
+        return this.grid.getSquare(this.xLocation, this.yLocation).hasColor();
+    }
+
+    // Returns True if there is a paint bucket in the square where the
+    // painter is standing.
+    public boolean isOnBucket() {
+        return this.grid.getSquare(this.xLocation, this.yLocation).containsPaint();
```
molly-moen, line 96: I think we might need a `hasBucket` method for a square--if there is a bucket with 0 paint in it, would we still want to return true here? My gut says yes, but I'm not sure. @jmkulwik

JillianK: How would we know if a square is a bucket?

molly-moen: there's a bucket asset, although I'm not sure how we define that along with a paint count. This could be a follow-up item, might be good to discuss with curriculum what defines a bucket.

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Helpers.java
JillianK: I'll do this as a follow up!

jmkulwik:
Add a link to the task and the student-facing API in the description. Also, add the Neighborhood engineering plan: https://docs.google.com/document/d/1SqSLogXsOKJD8QexbZAbxCpUGRdNP9Yj7zW8tN6oADI/edit#

org-code-javabuilder/lib/build.gradle
```diff
@@ -25,6 +25,8 @@ dependencies {
     implementation group: 'com.amazonaws', name: 'aws-java-sdk-apigatewaymanagementapi', version: '1.11.978'
     // https://mvnrepository.com/artifact/com.amazonaws/aws-java-sdk-sqs
     implementation group: 'com.amazonaws', name: 'aws-java-sdk-sqs', version: '1.11.978'
+    // https://mvnrepository.com/artifact/org.json/json
+    implementation group: 'com.googlecode.json-simple', name: 'json-simple', version: '1.1.1'
```
jmkulwik, line 29: This doesn't look like the package matches the comment associated with it. Did you intend to use json-simple or org.json?

JillianK: I meant to use json-simple, I'll replace that comment

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Direction.java
```diff
@@ -0,0 +1,63 @@
+package org.code.neighborhood;
+
+public enum Direction {
```
jmkulwik, line 3: TIL enums in Java do _way_ more than simple enums in other languages.

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Direction.java
```diff
@@ -0,0 +1,63 @@
...
+    if (text.equalsIgnoreCase("north")) {
+      return NORTH;
+    } else if (text.equalsIgnoreCase("east")) {
+      return EAST;
+    } else if (text.equalsIgnoreCase("south")) {
+      return SOUTH;
+    } else if (text.equalsIgnoreCase("west")) {
+      return WEST;
+    } else {
+      System.out.println("bad direction");
```
jmkulwik, line 59: Should this be an exception? (If yes, it doesn't necessarily need to be updated now - it can wait for the exception PR to be merged)

JillianK: It should be, I used exceptions everywhere else so I'll update this to match

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Grid.java
```diff
@@ -0,0 +1,49 @@
...
+    for (int y = height - 1; y >= 0; y--) {
+      ArrayList<String> squares = new ArrayList<String>();
+      for (int x = 0; x < width; x++) {
+        squares.add(grid[x][y].getPrintableDescription());
+      }
+      System.out.println(String.join(",", squares));
+    }
+  }
+
+  // Determines whether the given coordinate can be moved into
```
jmkulwik, line 26: NIT: Update the comment here to describe what would cause this to return true (or false). Reasoning - with the long chain of && conditions necessary here, it's easier to reason about what's happening if there's a "pretty" description.

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/GridFactory.java
```diff
@@ -0,0 +1,77 @@
...
+  protected Grid createGridFromJSON(String filename) {
+    File file = new File(GRID_FILE_NAME);
+    FileInputStream fis;
+    try {
+      fis = new FileInputStream(file);
+      byte[] data = new byte[(int) file.length()];
+      fis.read(data);
+      fis.close();
+      return createGridFromString(new String(data, "UTF-8"));
+    } catch (Exception e) {
```
jmkulwik, line 24: Where possible, enumerate the types of exceptions that could be caught.

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/GridFactory.java
```diff
@@ -0,0 +1,77 @@
...
+      throw new UnsupportedOperationException("Try adding a grid.txt file.");
```
jmkulwik, line 25: Is this the only reason we'd hit an exception?

JillianK: oh you're totally right, we could hit exceptions from createGridFromString and we shouldn't catch those here. The only one we should be catching is from reading the file.

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/GridFactory.java
```diff
@@ -0,0 +1,77 @@
...
+    String parseLine = String.join("", lines);
+    JSONParser jsonParser = new JSONParser();
+    try {
+      Object obj = jsonParser.parse(parseLine);
+      JSONArray gridSquares = (JSONArray) obj;
+      int height = gridSquares.size();
+      if (height == 0) {
+        throw new UnsupportedOperationException("Please check the format of your grid.txt file");
+      }
+      int width = ((JSONArray) gridSquares.get(0)).size();
```
jmkulwik, line 43: We'll want to add a check here that the grid is square (when this was first written, that wasn't a requirement, but has since been clarified)

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Painter.java
hanwalden: Yes, we shortened it because of the amount of typing, and it's tightly tied to the curriculum. Not something we can easily change.

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/GridSquare.java
```diff
@@ -0,0 +1,122 @@
...
+      default:
+        this.squareType = SquareType.UNKNOWN;
+        this.passable = false;
+    }
+  }
+
+  // Sets the color of the square to the given color
+  public void setColor(String color) {
+    if (!ColorHelpers.isColor(color)) {
+      System.out.println("Invalid color, please check your color format");
```
jmkulwik, line 67: Should this be an exception too?

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/GridSquare.java
```diff
@@ -0,0 +1,122 @@
...
+  public boolean isPassable() {
+    return this.passable;
+  }
+
+  // Decreases the paintCount by 1 if there is available paint
+  public void collectPaint() {
+    if (this.containsPaint()) {
+      this.paintCount--;
+    } else {
+      System.out.println("There's no paint to collect here");
```
jmkulwik, line 85: Check in with curriculum for these cases. We may want to set these as exceptions too.

jmkulwik: Unfortunately, this was a point of ambiguity on the original API. Figuring these out now will save us a lot of time in the future.

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/GridSquare.java
```diff
@@ -0,0 +1,122 @@
...
+    } else if (!this.color.equals("")) {
+      return this.color;
+    } else {
+      return String.valueOf(this.paintCount);
+    }
+  }
+
+  // Returns true if the color variable is populated
+  public boolean hasColor() {
+    return !this.color.equals("");
```
jmkulwik, line 115: Huh this feels a little odd. I would have expected `null` to be the "not set" value. Any reason we're preferring empty string here?

jmkulwik: (if we do prefer empty string, use `String.isEmpty()` instead)

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Painter.java
```diff
@@ -0,0 +1,146 @@
...
+  private int remainingPaint;
+  private Grid grid;
+  private String id;
+
+  public Painter(int x, int y, String direction, int paint) {
+    this.xLocation = x;
+    this.yLocation = y;
+    this.direction = Direction.fromString(direction);
+    if (this.direction == null) {
+      throw new UnsupportedOperationException("Invalid direction given to painter");
```
jmkulwik, line 17: Separate out exceptions that are directly caused by an error in the student's code versus exceptions caused by levelbuilders or by internal server errors. 

Example: If there's an issue with the grid.txt, we'd want to tell users to contact support and provide them with the level ID. If it's an issue with their code, we'd want to provide them with a stack trace.

For now, we don't have the full pipeline set up, but we need different exception types for "caused by the user" versus "caused by the levelbuilder" and possibly for "caused by the system"

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Painter.java
hanwalden: I don't have any suggestions around here. We tried to be clear with the naming, but I can see how it's challenging. 
isOnPaint - standing on Paint
hasPaint - painter has paint in their bucket
paint - action verb

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Painter.java
```diff
@@ -0,0 +1,146 @@
...
+public class Painter {
+  private static int lastId = 0;
+  private int xLocation;
+  private int yLocation;
+  private Direction direction;
+  private int remainingPaint;
+  private Grid grid;
+  private String id;
+
+  public Painter(int x, int y, String direction, int paint) {
```
jmkulwik, line 12: Also check for null values here and x/y out of bounds. This is the section that will be directly used by the students, so we'll want to be pretty strict on how we handle odd inputs.

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Painter.java
hanwalden: I thought we decided that if there is no paint, the bucket disappears.

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Painter.java
```diff
@@ -0,0 +1,146 @@
...
+        this.yLocation++;
+      } else if (this.direction.isSouth()) {
+        this.yLocation--;
+      } else if (this.direction.isEast()) {
+        this.xLocation++;
+      } else {
+        this.xLocation--;
+      }
+    } else {
+      System.out.println("You can't go that way");
```
jmkulwik, line 43: We should throw an exception here.

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Painter.java
jmkulwik: That definitely simplifies things ^ 👍

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Painter.java
```diff
@@ -0,0 +1,146 @@
...
+  public boolean isOnBucket() {
+    return this.grid.getSquare(this.xLocation, this.yLocation).containsPaint();
+  }
+
+  // Returns True if remainingPaint > 0
+  public boolean hasPaint() {
+    return this.remainingPaint > 0;
+  }
+
+  private boolean isValidMovement(Direction direction) {
```
jmkulwik, line 104: NIT: Put all private methods after the public methods (it's just the convention we're using)

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Painter.java
```diff
@@ -0,0 +1,146 @@
...
+    if (direction.isNorth()) {
```
jmkulwik, line 105: Tiny NIT: The other if statements use `this.direction` but this one uses `direction`

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Painter.java
```diff
@@ -0,0 +1,146 @@
...
+      return this.grid.validLocation(this.xLocation + 1, this.yLocation);
+    } else {
+      return this.grid.validLocation(this.xLocation - 1, this.yLocation);
+    }
+  }
+
+  // Returns True if there is no barrier one square ahead in the
+  // requested direction.
+  public boolean canMove(String direction) {
+    return isValidMovement(Direction.fromString(direction));
```
jmkulwik, line 119: NIT (convention) member functions should be prefaced with `this.`

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Painter.java
JillianK: that's because direction is a parameter

org-code-javabuilder/lib/src/test/java/org/code/neighborhood/GridFactoryTest.java
```diff
@@ -0,0 +1,98 @@
...
+  @Test
+  void createGridFromStringWithInvalidJSONThrowsException() {
+    GridFactory gridFactory = new GridFactory();
+    Exception exception =
+        assertThrows(
+            UnsupportedOperationException.class,
+            () -> {
+              gridFactory.createGridFromString("not valid json here:");
+            });
+    String expectedMessage = "Please check the format of your grid.txt file";
```
jmkulwik, line 28: All of our error messages are going to be specified on the client side so they can be translated. Here and other places, its sufficient to check that the right exception type is thrown without checking the error string.

jmkulwik: Ok. Here it is. 

All of our error messages are going to be specified on the client side so they can be translated. Rather than adding exception messages here, go ahead an add an exception enum like this:
https://github.com/code-dot-org/javabuilder/pull/32/files#diff-588560c58930937a4accd4ce0b91566398194bff4efdff36953ed6acf997f14c
For now, use the enum.toString directly as the exception message. We'll interpret those keys on the client side as we're building out neighborhood there.

org-code-javabuilder/lib/src/test/java/org/code/neighborhood/GridSquareTest.java
```diff
@@ -0,0 +1,173 @@
...
+    assertTrue(outputStreamCaptor.toString().trim().contains("There's no paint to collect here"));
+    // paintCount should be 0
+    assertFalse(s.containsPaint());
+  }
+
+  @Test
+  void getPrintableDescriptionReturnsXForNotPassable() {
+    GridSquare s = new GridSquare(0);
+    assertFalse(s.isPassable());
+    assertEquals(s.getPrintableDescription(), "x");
```
jmkulwik, line 148: Up to you if you want to include these cases. getPrintableDescription was only there for our debugging purposes originally.

org-code-javabuilder/lib/src/test/java/org/code/neighborhood/GridTest.java
```diff
@@ -0,0 +1,60 @@
...
+    assertFalse(grid.validLocation(1, 0));
+  }
+
+  @Test
+  void validLocationReturnsFalseForInvalidLocations() {
+    GridSquare s = new GridSquare(1);
+    GridSquare[][] squares = new GridSquare[1][1];
+    squares[0][0] = s;
+    Grid grid = new Grid(squares);
+    assertTrue(grid.validLocation(0, 0));
```
jmkulwik, line 26: NIT: This assertTrue should probably be in its own test as it doesn't match the goal of this test.

org-code-javabuilder/lib/src/test/java/org/code/neighborhood/GridTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+  @Test
+  void constructorBuildsGrid() {
+    GridSquare s = new GridSquare(1);
+    GridSquare[][] squares = new GridSquare[1][1];
+    squares[0][0] = s;
+    Grid grid = new Grid(squares);
+    assertTrue(grid.validLocation(0, 0));
+    assertFalse(grid.validLocation(0, 1));
+    assertFalse(grid.validLocation(1, 0));
```
jmkulwik, line 17: NIT: Looks like this test gets duplicated by the following test. If that's so, leave it out so there are fewer places to update when changing code in the future..

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Painter.java
jmkulwik: Should it be used as a parameter in 107 and 109 as well then?

jmkulwik: As a side note: to prevent future confusion, name the parameter differently than the class variable. i.e. the class variable could be `direction` while the parameter would be `intendedDirection` or similar.

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/GridFactory.java
```diff
@@ -0,0 +1,81 @@
...
+        currentHeight--;
+        JSONArray line = (JSONArray) gridSquares.get(currentHeight);
+        if (line.size() != width) {
+          throw new UnsupportedOperationException(
+              "width of line " + currentHeight + " does not match others. Cannot create grid.");
+        }
+        for (int currentX = 0; currentX < line.size(); currentX++) {
+          JSONObject descriptor = (JSONObject) line.get(currentX);
+          try {
+            int tileType = Integer.parseInt(descriptor.get(GRID_SQUARE_TYPE_FIELD).toString());
```
molly-moen, line 63: is the intent here that tileType is the id of the sprite? Or is it corresponding to wall, obstacle, etc.?

JillianK: I think it is corresponding to a wall, obstacle, etc.

molly-moen: Based on GridSquare it looks like it's wall, obstacle, etc. I think we'll also need an asset id here. It would be good to keep tileType as well since that's expected by Maze.

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/GridSquare.java
JillianK: not really, I just followed the sample code. I'll switch it

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Painter.java
JillianK: This one is a student code error if they give an invalid direction but I can update all of the grid reading ones to throw IOExceptions instead (although I think you mentioned that at some point students will be able to define their own grid.txt)

jmkulwik: True. We can definitely keep the different error messages (see the enum comment below), but creating different exception types will give us better control over how we handle the exceptions and communicate them to the user.

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/Painter.java
JillianK: oh shoot I see haha, I made them all use direction

org-code-javabuilder/lib/build.gradle
jmkulwik: Follow-up question: I also added a different JSON library on concurrent change. I was planning to swap over to com.googlecode.json-simple, but when I looked up the package, it looked somewhat unmaintained. i.e. the last version was from 2012 and it's hosted in google's code archive. It looks like you might have tried out org.json at first. Was there anything in particular that made you swap?

JillianK: I think I just found more examples for json-simple, I can switch over to org.json. Can you link the change where you added org.json so I can see how it works?

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/GridSquare.java
JillianK: We decided to just show the error message instead of stopping the program over being out of paint (the painter just wouldn't actually pick up any paint the way that Painter.takePaint is written) and we're following the same pattern for when the painter tries to paint without anything in their bucket.

org-code-javabuilder/lib/build.gradle
jmkulwik: There's a couple unrelated changes here, but ClientMessage.java is where I use org.json: https://github.com/code-dot-org/javabuilder/pull/32/commits/5521818e6846fb7e7a5ecb23d16ef176feae4d96#diff-bf69cb62b03f4ac01a380376939981a2d0db735a5b0a9c42d390fef9380cd017
^ This particular commit moves from using org.json to json-simple

org-code-javabuilder/lib/src/test/java/org/code/neighborhood/PainterTest.java
```diff
@@ -90,4 +90,33 @@ void takePaintPrintsErrorIfSquareHasNoPaint() {
...
+    painter.paint("red");
+    assertTrue(
+        outputStreamCaptor
+            .toString()
+            .trim()
+            .contains("There is no more paint in the painter's bucket"));
+  }
+
+  @Test
+  void paintDecrementsPaint() {
```
jmkulwik, line 116: Haha - love this test name :)

jmkulwik (approved):
Great work on this Jillian!

PR #34, opened 2021-05-14
Painter API Followup: support all of the web colors and a way to make an empty grid of a given size

JillianK:
This is a follow up to [this PR](https://github.com/code-dot-org/javabuilder/pull/31) that adds all of the web colors to the color helper and a World constructor that makes an empty grid of a given size as discussed in [this slack thread](https://codedotorg.slack.com/archives/C01EF4GJ9GE/p1620944753309200?thread_ts=1620941487.305100&cid=C01EF4GJ9GE).

If you can give my color spelling a second look when you review that would be great! I was pretty careful but with 140 colors I wouldn't be surprised if one of them is wrong. You can see the full list of extended colors I pulled from [here](https://en.wikipedia.org/wiki/Web_colors)


org-code-javabuilder/lib/src/main/java/org/code/neighborhood/ColorHelpers.java
```diff
@@ -7,14 +7,151 @@
...
+          "mintcream",
+          "snow",
+          "ivory",
+          "darkslategray",
+          "dimgray",
+          "slategray",
+          "lightslategray",
+          "darkgray",
+          "lightgray",
+          "gainsboro");
```
jmkulwik, line 151: The variety of white and offwhite colors is giving me serious flashbacks to picking a neutral color for our house.

org-code-javabuilder/lib/src/main/java/org/code/neighborhood/GridFactory.java
```diff
@@ -75,4 +75,16 @@ protected Grid createGridFromString(String description) throws IOException {
       throw new IOException(ExceptionKeys.INVALID_GRID.toString());
     }
   }
+
+  // Creates an empty size x size grid with every square being open
+  // and having assetId 0.
+  protected Grid createEmptyGrid(int size) {
```
jmkulwik, line 81: This should also be public since it will be accessed from an outside package. Also, mind adding a quick note to the comment describing how someone should set up their code to use this? i.e. I think it's something like `new World(); world.getFactory().createEmptyGrid(8)` right?

jmkulwik: Oops. I take that back.

jmkulwik (approved):
LGTM!

PR #35, opened 2021-05-17
Allow student code to access and execute Neighborhood commands

jmkulwik:
This is a refactor that puts Neighborhood in its own sub-project. This allows us to use built-in gradle functionality to build Neighborhood into a jar that can then be consumed by student code.

This is achieved by building Neighborhood into a [fat jar](https://stackoverflow.com/a/36539885) and placing it in the resources directory of the javabuilder sub-project. This can then be added to the classpath for the student code compiler and to the classes that are loaded when the student code runs.

As an addon to this refactor, we also move the client messaging code to its own sub-project so it can be imported by both Neighborhood and Javabuilder in the future.

PainterCode

Task: https://codedotorg.atlassian.net/browse/CSA-340
Neighborhood Doc: https://docs.google.com/document/d/1SqSLogXsOKJD8QexbZAbxCpUGRdNP9Yj7zW8tN6oADI/edit#

Next steps: Adjust the neighborhood code to use the ClientMessage format to signal actions to the client.


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserCodeCompiler.java
```diff
@@ -64,8 +67,16 @@ private CompilationTask getCompilationTask(DiagnosticCollector<JavaFileObject> d
       files.add(
           new JavaSourceFromString(projectFile.getClassName(), projectFile.getFileContents()));
     }
+    List<String> optionList = new ArrayList<String>();
+    optionList.add("-classpath");
+    // Include the neighborhood jar in the student code classpath so the student code can access neighborhood classes.
+    try {
+      optionList.add(Paths.get(Objects.requireNonNull(getClass().getClassLoader().getResource("neighborhood-full.jar")).toURI()).toString());
```
molly-moen, line 74: why do we need to add it to the classpath here as well as to the url class loader?

org-code-javabuilder/neighborhood/build.gradle
```diff
@@ -0,0 +1,39 @@
+/*
```
molly-moen, line 1: is the reason this is within `org-code-javabuilder` is this script builds neighborhood specifically for javabuilder?

jmkulwik: Yes. The different parts are loosely interdependent. Using gradle sub-projects, we can still publish specific packages from here, but it's much easier to import shared utilities which we'll need to do in order to manage passing signals from the mini-apps to Javalab.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserCodeCompiler.java
jmkulwik: My basic understanding: The classpath allows the compiler to make associations between classes whereas the url class loader pulls the classes into memory. I could definitely use clarification, however, as these are hypotheses based on what I was observing.

I was having a hard time find documentation explaining what I was observing. The basic observations, however, are these:
* Without using the classpath in the compiler, we get "package not found" errors. 
* Without loading the jar file at runtime, the JVM is able to see that the package exists, but is not able to find the classes. i.e. we get "class not found" errors.

molly-moen: huh that's interesting. Everything I can find online says "class not found" occurs when the jar does not exist on the classpath.

jmkulwik: Hmm is that true as well for running it via Method.invoke as opposed to from the command line? 

Maybe there's a special setting we need to add to the compiler step so we don't need to use it at runtime?

molly-moen: I wonder if this is the cause ([reference](https://www.geeksforgeeks.org/classnotfoundexception-vs-noclassdeffounderror-java/)):

> This exception also occurs when you have two class loaders and if a ClassLoader tries to access a class which is loaded by another classloader in Java.

molly-moen: it may be that we have to load it both places--if we do, can we add a constant or helper method so we aren't referencing `"neighborhood-full.jar"` twice?

molly-moen (approved):
LGTM!

org-code-javabuilder/util/build.gradle
```diff
@@ -0,0 +1,22 @@
+/*
```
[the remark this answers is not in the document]
jmkulwik: Good call on `protocol`. I'll update that.

For your third note here. Any advice/feedback you can give would be welcome. I spent several hours trying to come up with a design pattern that would let us separate Neighborhood from Javabuilder and yet still allow both of them to communicate with the client. I couldn't come up with anything and set up a _very_ hacky workaround that will let us unblock curriculum while we do some in-depth design. https://github.com/code-dot-org/javabuilder/pull/36

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaRunner.java
```diff
@@ -27,7 +27,11 @@ public JavaRunner(URL executableLocation, List<JavaProjectFile> javaFiles) {
    */
   public void runCode()
       throws UserFacingException, InternalFacingException, UserInitiatedException {
-    URL[] classLoaderUrls = new URL[] {this.executableLocation};
+    // Include the neighborhood classes in the code we are loading so student code can access them.
```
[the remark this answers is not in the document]
jmkulwik: i.e. Levels that have the neighborhood UI enabled?

Good call. I'll make a task for us to address this in the future (detecting that would me a more involved change than would make sense to include in this PR)

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserCodeCompiler.java
```diff
@@ -64,8 +67,22 @@ private CompilationTask getCompilationTask(DiagnosticCollector<JavaFileObject> d
...
+    optionList.add("-classpath");
+    // Include the neighborhood jar in the student code classpath so the student code can access
+    // neighborhood classes.
+    try {
+      optionList.add(
+          Paths.get(
+                  Objects.requireNonNull(
+                          getClass().getClassLoader().getResource("neighborhood-full.jar"))
+                      .toURI())
+              .toString());
```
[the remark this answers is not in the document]
jmkulwik: 👍

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaRunner.java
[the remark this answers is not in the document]
jmkulwik: Task: https://codedotorg.atlassian.net/browse/CSA-344

org-code-javabuilder/util/build.gradle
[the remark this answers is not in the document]
jamescodeorg: I'm a little stuck on this too.  I think we either need a "global" of some sort (similar to System.out) or we need to initialize World externally so that a handler can be injected.  Maybe ping me offline if you want to brainstorm some more on this?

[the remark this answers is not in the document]
jmkulwik: 👍 Thanks for putting some thought into it. I'll get a design brainstorm meeting set up.

PR #36, opened 2021-05-17
Updated server-side code to send Neighborhood messages to the client

jmkulwik:
With this change, we set up server-side messaging in Neighborhood in order to send signals to Javalab. 

This is paired with this pr from the code-dot-org repo: https://github.com/code-dot-org/code-dot-org/pull/40616

Task: https://codedotorg.atlassian.net/browse/CSA-340
Neighborhood Doc: https://docs.google.com/document/d/1SqSLogXsOKJD8QexbZAbxCpUGRdNP9Yj7zW8tN6oADI/edit#

### Concern
This PR sets up a hacky, immediately deprecated method of passing messages from Neighborhood to Javabuilder. What we do here is call system.out from Neighborhood. This passes a normal "system out" message to Javabuilder. Javabuilder checks whether the message is formatted as a command from Neighborhood. If so, it passes it to the client as such. 

This has a number of problems: 
1. If a student happens to call System.out on a json string formatted like a Neighborhood command, they will not see their message and Javalab will likely throw an exception.
1. A Java PrintStream does not guarantee when it will call flush(). We could end up with half of a Neighborhood command passed to students. This would print to the user-facing console and would not invoke the intended Neighborhood command.

This should be seen as a hacky workaround to temporarily unblock curriculum writing. We must do a more formal design of this system before users can access it.

Design Task: https://codedotorg.atlassian.net/browse/CSA-342

### DIFF
--- a/org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/NeighborhoodOutputHandler.java
+++ b/org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/NeighborhoodOutputHandler.java
@@ -1,0 +1,9 @@
+package org.code.neighborhood;
+
+public class NeighborhoodOutputHandler {
+  public static void sendMessage(NeighborhoodSignalMessage message) {
+    // This is a hack that we are temporarily using while we design a better system to handle passing signals from Javabuilder mini apps to Java Lab
+    // TODO: replace this with a more formal messaging system.
+    System.out.print(message.getFormattedMessage());
+  }
+}
--- a/org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/NeighborhoodOutputHandler.java
+++ b/org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/NeighborhoodOutputHandler.java
@@ -2,7 +2,8 @@ package org.code.neighborhood;
 
 public class NeighborhoodOutputHandler {
   public static void sendMessage(NeighborhoodSignalMessage message) {
-    // This is a hack that we are temporarily using while we design a better system to handle passing signals from Javabuilder mini apps to Java Lab
+    // This is a hack that we are temporarily using while we design a better
+    // system to handle passing signals from Javabuilder mini apps to Java Lab
     // TODO: replace this with a more formal messaging system.
     System.out.print(message.getFormattedMessage());
   }
--- a/org-code-javabuilder/lib/src/main/java/dev/javabuilder/DebuggingMessage.java
+++ b/org-code-javabuilder/lib/src/main/java/dev/javabuilder/DebuggingMessage.java
@@ -6,6 +6,6 @@ import org.code.protocol.ClientMessageType;
 /** A message directed to the client's terminal. Equivalent to System.out.print. */
 public class DebuggingMessage extends ClientMessage {
   DebuggingMessage(String value) {
-    super(ClientMessageType.DEBUG, value, null);
+    super(ClientMessageType.DEBUG, value);
   }
 }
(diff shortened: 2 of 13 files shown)


org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/NeighborhoodOutputHandler.java
```diff
@@ -0,0 +1,10 @@
+package org.code.neighborhood;
+
+public class NeighborhoodOutputHandler {
+  public static void sendMessage(NeighborhoodSignalMessage message) {
+    // This is a hack that we are temporarily using while we design a better
+    // system to handle passing signals from Javabuilder mini apps to Java Lab
+    // TODO: replace this with a more formal messaging system.
+    System.out.print(message.getFormattedMessage());
```
molly-moen, line 8: is there a way we can use dependency injection and call sendMessage on the output adapter directly here?

jmkulwik: There might be, but I definitely reached the edge of my understanding of Java theory. The general problem:

In order to compile student projects with Neighborhood, the Neighborhood classes need to be on the student classpath. We can do this with the raw source (i.e. without a jar), but we would also need to include the dependencies on the classpath. This is possible, but would require some thought around how we handle dependency management since gradle would not do this for us.

The option I ended up going with here was separating Neighborhood into its own sub-project, using a fat-jar, and letting gradle do all the work. Separating Neighborhood into a sub-project was something that needed to happen anyway to follow best practices and allow us to publish Neighborhood if we ever wanted to. It was also something we needed to do if we want to ensure students don't access the core Javabuilder code.

Either way, we need to load Neighborhood into our classloader when we run the student code. Unknown: Is it possible to inject state into Neighborhood when it is loaded alongside student code? i.e. is it possible to use dependency injection? If Neighborhood is imported as a jar, I guessed this would be impossible. i.e. Neighborhood would use its own copy of the output adapter, thereby ignoring our entrypoint setup. This is perhaps the first critical question. And I ran out of time to answer it.

A follow-up question I considered: If we find out we can't use dependency injection, is there something else we can use? This ended up being a big open ended question. i.e. we do different I/O setup based on our different entry points (AWS/local websocket/local headless). I brainstormed a couple ways for Neighborhood to decide where to direct its output, but none of the answers were good. The best I could come up with here was to use environment variables set at the entry points and adjusting the output adapter to be stateless.

molly-moen: makes sense. I'm also not sure if it's possible. Would you mind copying this info into the follow-up task so we keep it in mind as we fine tune this?

molly-moen (approved):
This seems like a reasonable workaround for now. I am a bit concerned about the second problem you outlined (no guarantee of the flush time)--we should be aware of this as we are developing and see if it happens at all.

org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/NeighborhoodOutputHandler.java
jmkulwik: ^ Long explanation, but the short answer is I don't know. I'm hopeful someone knows of a better design pattern. If needed, we can use this to unblock curriculum while we work to discover something better. In the meantime, this code has a hard EOL of August 16th.

jmkulwik: Will do!

jamescodeorg: I don't have any quick answers but would be happy to join when you brainstorm this.  I _think_ that invoking a method via reflection is almost the same as invoking it directly which means that we can also call other methods to setup state before calling the student's `main` function.  (In fact, I think we're already doing something similar to this with System.out, right?)

PR #39, opened 2021-05-18
Neighborhood: add all client signals

molly-moen:
Add all the signals neighborhood needs to send to the client for various client-facing actions. Move had been implemented previously. This PR adds:
- INITIALIZE
- PAINT
- REMOVE_PAINT
- TAKE_PAINT
- HIDE_PAINTER
- SHOW_PAINTER

These signals will be used on javalab to invoke various painter actions.
As a part of this PR, also updated the variable `visible` on `UserFileData` to be `isVisible` since we have updated the variable on the javalab side.

### DIFF
--- a/org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/Grid.java
+++ b/org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/Grid.java
@@ -1,6 +1,7 @@
 package org.code.neighborhood;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 
 public class Grid {
   private GridSquare[][] grid;
@@ -41,12 +42,16 @@ public class Grid {
 
   // Hides all buckets from the screen
   public void hideBuckets() {
-    System.out.println("You hid the buckets");
+    NeighborhoodOutputHandler.sendMessage(
+            new NeighborhoodSignalMessage(NeighborhoodSignalKey.HIDE_BUCKETS, new HashMap<>())
+    );
   }
 
   // Displays all buckets on the screen
   public void showBuckets() {
-    System.out.println("You displayed the buckets");
+    NeighborhoodOutputHandler.sendMessage(
+            new NeighborhoodSignalMessage(NeighborhoodSignalKey.SHOW_BUCKETS, new HashMap<>())
+    );
   }
 
   protected int getSize() {
--- a/org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/NeighborhoodSignalKey.java
+++ b/org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/NeighborhoodSignalKey.java
@@ -2,7 +2,7 @@ package org.code.neighborhood;
 
 public enum NeighborhoodSignalKey {
   // Initialize a new painter
-  INITIALIZE,
+  INITIALIZE_PAINTER,
   // Move the painter one space forward
   MOVE,
   // Paint the current location
@@ -14,5 +14,11 @@ public enum NeighborhoodSignalKey {
   // Hide the painter on the screen
   HIDE_PAINTER,
   // Show the painter on the screen
-  SHOW_PAINTER
+  SHOW_PAINTER,
+  // Turn the painter left
+  TURN_LEFT,
+  // Hide all paint buckets
+  HIDE_BUCKETS,
+  // Show all paint buckets
+  SHOW_BUCKETS
 }
--- a/org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/NeighborhoodSignalKey.java
+++ b/org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/NeighborhoodSignalKey.java
@@ -1,6 +1,18 @@
 package org.code.neighborhood;
 
 public enum NeighborhoodSignalKey {
+  // Initialize a new painter
+  INITIALIZE,
   // Move the painter one space forward
-  MOVE
+  MOVE,
+  // Paint the current location
+  PAINT,
+  // Remove all paint from current location
+  REMOVE_PAINT,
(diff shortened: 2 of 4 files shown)


org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/NeighborhoodSignalKey.java
```diff
@@ -1,6 +1,18 @@
...
+  // Paint the current location
+  PAINT,
+  // Remove all paint from current location
+  REMOVE_PAINT,
+  // Take paint from the bucket
+  TAKE_PAINT,
+  // Hide the painter on the screen
+  HIDE_PAINTER,
+  // Show the painter on the screen
+  SHOW_PAINTER
```
jmkulwik, line 17: I think we also need TURN_LEFT

org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/NeighborhoodSignalKey.java
```diff
@@ -1,6 +1,18 @@
 package org.code.neighborhood;
 
 public enum NeighborhoodSignalKey {
+  // Initialize a new painter
+  INITIALIZE,
```
jmkulwik, line 5: Is this initializing the whole maze or an individual painter? If it's the latter, a name such as INITIALIZE_PAINTER may make more sense.

org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/NeighborhoodSignalKey.java
jmkulwik, line 17: One item that wasn't on the list but we'll also need to add is HIDE_BUCKETS to make all the buckets on a level disappear.

molly-moen: good catch

org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/NeighborhoodSignalKey.java
molly-moen: it's a single painter, I will update

org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/NeighborhoodSignalKey.java
jmkulwik: Definitely can be a follow-up. Looks like this was missed even on the API description.

molly-moen: it's a quick update, I'll add it

org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/Grid.java
```diff
@@ -41,12 +42,16 @@ public GridSquare getSquare(int x, int y) {
...
-    System.out.println("You hid the buckets");
+    NeighborhoodOutputHandler.sendMessage(
+            new NeighborhoodSignalMessage(NeighborhoodSignalKey.HIDE_BUCKETS, new HashMap<>())
+    );
   }
 
   // Displays all buckets on the screen
   public void showBuckets() {
-    System.out.println("You displayed the buckets");
+    NeighborhoodOutputHandler.sendMessage(
```
jmkulwik, line 52: [Non-blocking comment] I think Hannah wanted students to be able to call this directly from the Painter object. We'll want to double check the use case. Definitely not blocking for this PR as that could take a couple cycles to figure out.

jmkulwik (approved):
Nice work!

jamescodeorg (commented):
Where can I find Maze's "interface" on the client side?

I'm wondering if it would be better to send the resolved state rather than the action (e.g. SET_DIRECTION instead of TURN_LEFT or TURN_RIGHT) but I guess it depends a bit on what the interface is on the client side?

Also, Hannah mentioned that students will be asked to extend Painter with methods such as turnRight().  Would that work by having painter turn left 3 times?  Or will there be a way for students to extend the set of commands/messages?

molly-moen:
@jamescodeorg the controller we will use to access maze functions is [here](https://github.com/code-dot-org/maze/blob/main/src/mazeController.js), let me know if that's enough info!
For the turn question, we could go either way. I kept it aligned with what the command was on the server-side, but from the client side we will be calling turn with a direction (see function [here](https://github.com/code-dot-org/maze/blob/main/src/mazeController.js#L247)), so SET_DIRECTION would also make sense. However, students will not be extending the set of commands/messages--this logic is all hidden from them. So their turnRight() would be turnLeft() 3 times.

org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/Painter.java
```diff
@@ -181,4 +190,22 @@ private boolean isValidMovement(Direction movementDirection) {
...
+    return details;
+  }
+
+  private void sendOutputMessage(NeighborhoodSignalKey signalKey, HashMap<String, String> details) {
+    NeighborhoodOutputHandler.sendMessage(new NeighborhoodSignalMessage(signalKey, details));
+  }
+
+  private void sendInitializationMessage() {
+    HashMap<String, String> initDetails = this.getSignalDetails();
+    initDetails.put("direction", this.direction.getDirectionString());
```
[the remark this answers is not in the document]
molly-moen: painter id is added in the `getSignalDetails` method

jamescodeorg:
Got it, thanks!  Lots of interesting questions around how to keep the back-end and front-end in sync, it will be interesting to see how this evolves!

PR #42, opened 2021-05-25
Unify user-facing exceptions under one common exception type

jmkulwik:
This is a prefactor for a change that unifies our output protocol to enable mini-apps to use our output adapter. The full change, with all prefactors, can be viewed here: https://github.com/code-dot-org/javabuilder/pull/41 

This unifies the output protocol for exceptions by setting one parent exception type (`JavabuilderException`) for all internally-created exceptions that will be passed to a user. All exceptions should inherit from this exception type.


molly-moen (approved):
LGTM!

PR #43, opened 2021-05-25
Add Error type and migrate Neighborhood to use the JavabuilderThrowableProtocol

jmkulwik:
This is a prefactor for a larger change that unifies our output protocol to enable mini-apps to use our output adapter. The full change can be viewed here: https://github.com/code-dot-org/javabuilder/pull/41 

This adds a `JavabuilderRuntimeException` type that can be used for [unchecked exceptions](https://www.geeksforgeeks.org/checked-vs-unchecked-exceptions-in-java/). This is especially important for mini-apps such as Neighborhood where our setup code can throw errors that should not be caught by the student. As part of this change, Neighborhood is migrated to using this Error type.

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalMain.java
+++ b/org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalMain.java
@@ -1,8 +1,8 @@
 package dev.javabuilder;
 
 import org.code.javabuilder.*;
-import org.code.protocol.JavabuilderError;
 import org.code.protocol.JavabuilderException;
+import org.code.protocol.JavabuilderRuntimeException;
 
 /**
  * Intended for local testing only. This is a local version of the Javabuilder lambda function. The
@@ -22,7 +22,7 @@ public class LocalMain {
         codeBuilder.buildUserCode();
         codeBuilder.runUserCode();
       }
-    } catch (JavabuilderException | JavabuilderError e) {
+    } catch (JavabuilderException | JavabuilderRuntimeException e) {
       outputAdapter.sendMessage(e.getExceptionMessage());
       System.out.println("\n" + e.getLoggingString());
     } catch (InternalFacingException e) {
--- a/org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalMain.java
+++ b/org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalMain.java
@@ -1,6 +1,7 @@
 package dev.javabuilder;
 
 import org.code.javabuilder.*;
+import org.code.protocol.JavabuilderError;
 import org.code.protocol.JavabuilderException;
 
 /**
@@ -21,7 +22,7 @@ public class LocalMain {
         codeBuilder.buildUserCode();
         codeBuilder.runUserCode();
       }
-    } catch (JavabuilderException e) {
+    } catch (JavabuilderException | JavabuilderError e) {
       outputAdapter.sendMessage(e.getExceptionMessage());
       System.out.println("\n" + e.getLoggingString());
     } catch (InternalFacingException e) {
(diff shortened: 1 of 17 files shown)


molly-moen (approved):
LGTM!

jamescodeorg (commented):
I've spent some time with this and don't think I fully understand what's going on here yet.  I need to step away, but wanted to leave a few thoughts:
1) I'm a little confused by throwable vs. error vs. exception here.  IIRC, application-level exception hierarchies generally inherit from RuntimeException (for unchecked exceptions) or Exception (for checked exceptions).  It's usually an anti-pattern to catch Error or Throwable but I could imagine us wanting to do something like that here so we can report such low-level errors to the student but we'd have to be pretty careful.  Often when an Error is thrown, the system is in a pretty unstable state.
2) I _think_ we may want a "mini-app library" that sits between neighborhood/theater and protocol.  Think a bit about whether there is a common interface that all mini-apps should implement and whether there is a common interface that all mini-apps should depend on.
3) I still think that it would help to have an initialize method for mini-apps to reduce the amount of static classes and methods needed.  I will think about this a bit more.

jmkulwik:
> IIRC, application-level exception hierarchies generally inherit from RuntimeException (for unchecked exceptions) or Exception (for checked exceptions). It's usually an anti-pattern to catch Error or Throwable but I could imagine us wanting to do something like that here so we can report such low-level errors to the student but we'd have to be pretty careful. Often when an Error is thrown, the system is in a pretty unstable state.

^ Good callout there. In this case, I think we'll want to move all or at least most of our Exceptions to RuntimeException - I didn't know about that one when I added in the Error case. The JavabuilderException/Errors are intended to communicate to the user that the lambda "crashed" and there isn't anything they could have done about it - their next step should be to try again or, failing that, to reach out to our support team.

I think we also want to catch other Errors and Throwables at the top level LambdaHandler to ensure we log them in a way that we can read later. Right now, that isn't set up, but we'll want to make sure they end up in the CloudWatch logs.

jmkulwik:
> I think we may want a "mini-app library" that sits between neighborhood/theater and protocol. Think a bit about whether there is a common interface that all mini-apps should implement and whether there is a common interface that all mini-apps should depend on.

That generally sounds reasonable. There's not any common interface that mini-apps should implement that I can think of, but they do all depend on the InputAdapter, OutputAdapter, ClientMessage, and JavabuilderException classes.

jmkulwik:
> I still think that it would help to have an initialize method for mini-apps to reduce the amount of static classes and methods needed. I will think about this a bit more.

I'm coming around to your way of thinking on this. I'm just not quite sure what it would look like. i.e. Javabuilder, Neighborhood, and soon Theater should all be pretty separate. It seems odd to me that Javabuilder would take a dependency on Neighborhood. But I'm not sure if I'm thinking about it the right way. I'd definitely welcome any thoughts you have.

jamescodeorg:
I spent some more time with this and I think I understand the design here better now.  One thing that was confusing to me was having GlobalProtocol plumbed throughout the whole stack.  It might be simpler just to have GlobalProtocol take no parameters on construction and have methods to set the input/output adapters.  We could also try having GlobalProtocol just be a static class; this would be simpler still but might affect unit testability.  This would limit the places in the code that need to interact with this "global variable".

After sketching out the system, I take back part of my second comment above.  I think the protocol jar you've already broken out probably _is_ the mini-app library.  I would recommend keeping packages as coherent as possible and consider moving the error/exception classes into a separate package (in the same jar).  (Renaming some of packages and jars later might help with clarity.)

If you're interested in trying out 3, I think we might be able to avoid the direct dependency by having the mini-apps implement an interface and then calling that interface via reflection at run-time.  This is similar to how Spring does runtime injection of dependencies.

Let me know if you want to chat/pair sometime.

PR #44, opened 2021-05-25
Adds a GlobalProtocol singleton that contains IO protocols

jmkulwik:
This unifies our output protocol to enable mini-apps to use our output adapter. There were a few prefactors to enable this. The full change, with prefactors can be viewed here: https://github.com/code-dot-org/javabuilder/pull/41 

This removes our "hacky workaround" of the `FormattedClientMessage` and allows mini apps to directly use the outputAdapter without relying on System Out.


org-code-javabuilder/protocol/src/main/java/org/code/protocol/GlobalProtocol.java
```diff
@@ -0,0 +1,32 @@
+package org.code.protocol;
+
+public class GlobalProtocol {
```
molly-moen, line 3: nit: add a comment on this class explaining its purpose

org-code-javabuilder/protocol/src/main/java/org/code/protocol/GlobalProtocol.java
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  public OutputAdapter getOutputAdapter() {
+    return this.outputAdapter;
+  }
+
+  public InputAdapter getInputAdapter() {
+    return this.inputAdapter;
+  }
+}
```
molly-moen, line 32: nit: add a blank line to end of this file

org-code-javabuilder/protocol/src/main/java/org/code/protocol/GlobalProtocol.java
```diff
@@ -0,0 +1,32 @@
...
+  private final OutputAdapter outputAdapter;
+  private final InputAdapter inputAdapter;
+
+  private GlobalProtocol(OutputAdapter outputAdapter, InputAdapter inputAdapter) {
+    this.outputAdapter = outputAdapter;
+    this.inputAdapter = inputAdapter;
+  }
+
+  public static void create(OutputAdapter outputAdapter, InputAdapter inputAdapter) {
+    GlobalProtocol.protocolInstance = new GlobalProtocol(outputAdapter, inputAdapter);
```
molly-moen, line 14: should this throw an error if there is already a protocolInstance?

org-code-javabuilder/lib/src/main/resources/main_painter.json
```diff
@@ -1,29 +1,14 @@
...
+    "ScannerSample.java": {
```
molly-moen, line 3: nit: rename this class?

org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/Grid.java
```diff
@@ -40,20 +40,6 @@ public GridSquare getSquare(int x, int y) {
     }
   }
 
-  // Hides all buckets from the screen
```
molly-moen, line 43: did we confirm these functions should be moved? Per the student-facing api they are on Grid, which is the logical place for them to go.

molly-moen (commented):
This is a nice solution!

org-code-javabuilder/protocol/src/main/java/org/code/protocol/GlobalProtocol.java
jmkulwik: That's a good question - for the local scenario or the lambda scenario there's no valid instance where GlobalProtocol.protocolInstance would already exist. But for the local websocket server scenario, it would i.e. the instance gets set onOpen, but never gets removed until the server is turned off.

jmkulwik: Now that I'm writing this out, there is an opportunity to unset the instance onClose. Although thinking through that, we might run into problems running javabuilder in multiple tabs. But it's also only for a local development scenario.

org-code-javabuilder/lib/src/main/resources/main_painter.json
jmkulwik: ✔️

org-code-javabuilder/protocol/src/main/java/org/code/protocol/GlobalProtocol.java
jmkulwik: ✔️

org-code-javabuilder/protocol/src/main/java/org/code/protocol/GlobalProtocol.java
jmkulwik: ✔️

org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/Grid.java
jmkulwik: Turns out they were supposed to stay on Grid. Moved them back.

org-code-javabuilder/protocol/src/main/java/org/code/protocol/GlobalProtocol.java
jmkulwik: I'm not sure there's an immediate good answer here. Happy to take suggestions. For now, I'm leaving as-is

molly-moen: My general thought here (non-blocking) is a singleton should only have 1 instance of the object. Allowing create to overwrite it seems like an anti-pattern. If we want it to be able to be overwritten for certain cases that's ok, but if we don't then maybe throw an exception or just have this be a no-op?

molly-moen (approved):
LGTM! One remaining question on the singleton, but non-blocking

org-code-javabuilder/protocol/src/main/java/org/code/protocol/GlobalProtocol.java
jmkulwik: Added a task to track this. I'll try out a couple things. For now, merging to unblock Theater development.

jmkulwik: https://codedotorg.atlassian.net/browse/CSA-445

PR #46, opened 2021-05-26
Javabuilder: Load serialized maze for neighborhood projects

molly-moen:
This PR adds the loading the serialized maze for neighborhood projects. If a project is a neighborhood project, it will send `"options: "{\"useNeighborhood\": \"true\"}` as a new parameter to the authorizer. If we see that flag, we will fetch the serialized maze from dashboard (utilizing the pre-existing `iss` (issuer) parameter and a new `levelId` parameter from the authorizer) and save it as a "grid.txt" file for Neighborhood to pick up. This enables our neighborhood levels to properly recognize things like obstacles on the backend.


jmkulwik (approved):
LGTM!

PR #48, opened 2021-06-02
Switched coordinates so 0,0 is in the top left

jmkulwik:
Originally, we wanted to set up Neighborhood with the origin in the bottom left. However, since we are planning to support pixel operations on the curriculum units following Neighborhood, we'll want to use origin in the top left for consistency. This updates the coordinates.

This pairs with a similar change in the code-dot-org repo: https://github.com/code-dot-org/code-dot-org/pull/40889

originTopLeft

## Links

- spec: [Slack discussion](https://codedotorg.slack.com/archives/C01EF4GJ9GE/p1622146181166200)
- jira ticket: [CSA-407](https://codedotorg.atlassian.net/browse/CSA-407)

### DIFF
--- a/org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/GridFactory.java
+++ b/org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/GridFactory.java
@@ -47,10 +47,8 @@ public class GridFactory {
 
       // We start at the maximum height because we're reading the grid from top to bottom in the
       // file.
-      int currentHeight = height;
       for (int currentY = 0; currentY < height; currentY++) {
-        currentHeight--;
-        JSONArray line = (JSONArray) gridSquares.get(currentHeight);
+        JSONArray line = (JSONArray) gridSquares.get(currentY);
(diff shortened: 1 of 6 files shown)


madelynkasula (approved):

org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/GridFactory.java
```diff
@@ -47,10 +47,8 @@ protected Grid createGridFromString(String description) throws IOException {
 
       // We start at the maximum height because we're reading the grid from top to bottom in the
```
hannahbergam, line 48: Does this comment still make sense as-is?

hannahbergam (approved):

org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/GridFactory.java
jmkulwik: Good catch!

PR #50, opened 2021-06-10
Set up the Theater framework on Javabuilder

jmkulwik:
This sets up a couple things for Theater.
* The Theater library module
* The jar file
* Integrates the compiler and class loader with the new Theater Jar file
* Sets up testing and linting in Theater
* The initial signals for client-side theater
* The encoding of gif -> base64 for transfer across the websocket

What this looks like on the client side
TheaterSample2


org-code-javabuilder/theater/src/test/java/org/code/theater/ImageEncoderTest.java
```diff
@@ -0,0 +1,26 @@
...
+
+  @Test
+  public void returnsBase64Image() {
+    TheaterMessage message = ImageEncoder.encodeImageToMessage(new BufferedImage(1, 1, 1));
+    Assertions.assertTrue(message.getDetail().toString().contains("\"image\""));
+    Assertions.assertTrue(
+        message
+            .getDetail()
+            .toString()
+            .contains("R0lGODlhAQABAPAAAAAAAAAAACwAAAAAAQABAEAIBAABBAQAOw=="));
```
molly-moen, line 24: nit: add comment explaining this

org-code-javabuilder/theater/src/main/java/org/code/theater/CatImage.java
```diff
@@ -0,0 +1,80 @@
...
+    // your local computer
+    // IJ.save(image, "<my local directory>\beach.gif");
+
+    ImageProcessor ip = image.getProcessor();
+    BufferedImage bufferedImage = (BufferedImage) ip.createImage();
+    outputAdapter.sendMessage(ImageEncoder.encodeImageToMessage(bufferedImage));
+  }
+
+  public void buildCanvas() {
+    ImagePlus image = IJ.createImage("test", 300, 150, 1, 24);
```
molly-moen, line 50: general q: are we thinking we will use ImageJ for image creation in the 'real' version or is this a placeholder?

molly-moen (approved):
LGTM, a couple non-blocking comments

org-code-javabuilder/lib/build.gradle
```diff
@@ -59,6 +59,7 @@ build.dependsOn packageSmall
 
 task buildZip(type: Zip) {
```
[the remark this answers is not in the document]
jmkulwik: That's a good idea. I'll do that as a follow-up.

org-code-javabuilder/theater/src/main/java/org/code/theater/CatImage.java
jmkulwik: This is a placeholder. We certainly can use ImageJ, but it's not required.

jmkulwik: (I had used it while investigating how to set up Theater and figured I might as well include it for demo purposes in case it would be useful down the line)

PR #51, opened 2021-06-10
Proposed Theater API

moneppo:
**Note to the reviewer**: We don't plan to merge this PR; however, it outlines the classes and methods that we will have available to students for the Theater mini app. Please leave comments, questions, concerns, etc directly on the PR in the same manner as for code review or spec review.

## Spec
The spec for Theater can be found here: https://docs.google.com/document/d/1zVCVa2MRvMQqt1m13_KEP8vjWcf61Rd8N_u3wuuic04/edit#bookmark=id.ag66gbya5zt1 

(Section copied from spec) A major innovation in creative expression over the past decade is the short form video. Short form videos have gained recent popularity on platforms like TikTok and instagram, but their usage has been a major part of internet culture in the form of animated GIFs for much longer, enabling dozens of different forms of collaborative self-expression. We have an opportunity to tap into this medium to engage students, provided we don’t try to chase after the most recent fads in the space. There are 3 elements that make short form video a compelling medium of expression on the internet: ease of consumption, wide range of creation & remix options, and instant sharing with viral potential. Theater has the potential to provide an environment that enables new creative opportunities in short form video.

## Dev Plan
The dev plan for Theater can be found here: https://docs.google.com/document/d/1F8g6D1P3tIMzEpteSf4a4G_4XO2QIKBAHLN_S6KeRlI/edit# 

(Section copied from the dev plan) Theater is a simplified sound, image, and canvas library that produces a gif file and a sound file that will be played back concurrently on the client’s browser. When a student writes Theater code, the following happens:
* The student performs pixel, drawing, and sound manipulation commands. They specify time between visual frames and/or audio playback.
* We encode these visual frames as a compressed non-looping gif. The audio is encoded as a web-safe audio format such as FLAC, MP3, or WAV
* When the play command is executed, the gif and audio file are sent to the client. The gif is displayed on a canvas or similar container on Java Lab. The audio is played concurrently.


org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
```diff
@@ -0,0 +1,181 @@
...
+    * 
+    * @param sound an array of samples to play.
+    */
+   public static void playSound(double[] sound) {
+   }
+
+   /**
+    * Plays the sound referenced by the file name.
+    * 
+    * @param filename an array of samples to play.
```
molly-moen, line 19: nit: update description of filename

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
```diff
@@ -0,0 +1,181 @@
...
+    * @throws SoundException if the file can't be found in the project.
```
molly-moen, line 20: would it make more sense to throw a `FileNotFoundException` instead of a `SoundException`?

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
```diff
@@ -0,0 +1,181 @@
...
+    * 
+    * @param width
+    */
+   public static void setStrokeWidth(double width) {
+   }
+
+   /**
+    * Sets the color of lines drawn.
+    * 
+    * @param color
```
molly-moen, line 162: will the color be similar to neighborhood--any html color or hex code?

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
```diff
@@ -0,0 +1,181 @@
...
+    * 
+    * @param file     the name of the file in the asset manager
+    * @param x        the left side of the image in the canvas
+    * @param y        the top of the image in the canvas
+    * @param width    the width to draw the image on the canvas
+    * @param height   the height to draw the image on the canvas
+    * @param rotation the amount to rotate the image
+    */
+   public static void drawImage(String filename, double x, double y, double width, double height, double rotation)
+         throws ImageException {
```
molly-moen, line 65: when does it throw an `ImageException`?

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
```diff
@@ -0,0 +1,181 @@
+package org.code.theater;
+
+public enum Instrument {
+   PIANO, BASS
+}
+
+public static class Theater {
```
molly-moen, line 7: What is our reasoning for everything being static? The `Theater` has state (the series of images/sounds) and my intuition would be for it to have instance methods instead of static methods.

molly-moen: another piece of state the theater could have is its height and width. Are we expecting students to get the height and width of the overall theater programmatically? Or are they just going to be told it's 400x400?

moneppo: The motivation for static methods is that we don't want students to have the ability to instantiate and use multiple theaters. We could go with a singleton pattern (in fact no matter what, behind the scenes I'd imagine this would be implemented with a singleton), but my reasoning was that would be another concept we'd need to explain.

Like Game Lab, the height is planned to be a fixed 400x400 pixels, so we can guarantee that.

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
moneppo: Yes - I will clarify that in the doc comment.

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
moneppo: Oy, yea I think this needs to be a `FileNotFoundException` too.

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
moneppo: Oh, yes, that's much better.

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
molly-moen: Makes sense, I agree it would be too much to try to explain a singleton to students. I do think it would be nice to have `getHeight` and `getWidth` methods on this class then

moneppo: Cool, I can add those!

jmkulwik: Would something like `Theater.getCanvas()` be more intuitive than the static methods?

jmkulwik: Then all operations are done against Canvas (and perhaps against a similar class for audio)

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
```diff
@@ -16,10 +30,10 @@ public static void playSound(double[] sound) {
    /**
     * Plays the sound referenced by the file name.
     * 
-    * @param filename an array of samples to play.
-    * @throws SoundException if the file can't be found in the project.
+    * @param filename the file to play in the asset manager.
+    * @throws FileNotFoundException if the file can't be found in the project.
     */
-   public static void playSound(String filename) throws SoundException {
+   public static void playSound(String filename) throws FileNotFoundException {
```
jmkulwik, line 36: Just to double check. Students will have learned about exception handling at this point? i.e. that was a concern when we were building the Neighborhood API - students wouldn't know how to handle exceptions yet.

org-code-javabuilder/theater/src/main/java/org/code/Pixel.java
```diff
@@ -0,0 +1,57 @@
+package org.code;
+
+public class Pixel {
```
madelynkasula, line 3: this could always be added later, but should there be a getter/setter for red, green, and blue? (e.g., `getColor()` and `setColor(int r, int g, int b)`)

kind of tangential, but this question then makes me wonder if there should be a `Color` type that can express colors as an int[] or string. then, (excuse my weird ruby/java pseudocode...) `Image#clear(String color)` would become `Image#clear(Color color)`. this does add some complexity to the student interaction, though, as they'd have to understand the multiple ways a color can be expressed, but maybe that's a good lesson for students?

org-code-javabuilder/theater/src/main/java/org/code/Image.java
```diff
@@ -0,0 +1,67 @@
...
+
+public class Image {
+    /**
+     * Creates a new image object, using the pixel information from the file
+     * uploaded to the asset manager.
+     * 
+     * @param filename the name of the image loaded into the asset manager for the
+     *                 project
+     */
+    public Image(String filename) throws ImageException;
```
jmkulwik, line 11: If we're creating custom exceptions that will be visible to the student, we should also add them to the api so they are documented.

jmkulwik: Also, in what cases would the ImageException be thrown?

org-code-javabuilder/theater/src/main/java/org/code/Image.java
```diff
@@ -0,0 +1,67 @@
+package org.code;
```
jmkulwik, line 1: NIT: This will be `org.code.theater`

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
```diff
@@ -0,0 +1,196 @@
...
+    * Draws text on the image.
+    * 
+    * @param text   the text to draw
+    * @param x      the distance from the left side of the image to draw the text.
+    * @param y      the distance from the top of the image to draw the text.
+    * @param color  the color to draw the text, using any CSS color string (e.g. #234 or green)
+    * @param font   the name of the font to draw the text in
+    * @param height the height of the text in pixels.
+    */
+   public static void drawText(String text, double x, double y, String color, String font, double height) {
```
jmkulwik, line 107: Should this also have a rotation value?

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
```diff
@@ -0,0 +1,196 @@
...
+
+   /**
+    * Draw a regular polygon on the canvas.
+    * 
+    * @param x     the center X coordinate of the polygon
+    * @param y     the center Y coordinate of the polygon
+    * @param sides the number of sides of the polygon
+    * @param size  the distance from the center to each point on the polyon
+    */
+   public static void drawPolygon(double x, double y, int sides, double size) {
```
jmkulwik, line 129: NIT: Should this be `drawRegularPolygon`?

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
```diff
@@ -0,0 +1,196 @@
...
+    * @param CSS color string 
+    */
+   public static void setStrokeColor(String color) {
+   }
+
+   /**
+    * Sets the fill color for all shapes drawn
+    * 
+    * @param color CSS color string to fill any shape. Defaults to "none", and 
+    *              setting it to "none" ensures that any shapes are not filled.
```
jmkulwik, line 186: It might make more sense to allow students to set `null` here instead. Or perhaps we create a second function that doesn't take an input. Java doesn't have default values.

org-code-javabuilder/theater/src/main/java/org/code/Image.java
moneppo: What if I wanted to use it in the Park?

org-code-javabuilder/theater/src/main/java/org/code/Image.java
moneppo: Oh, I may have missed this - it should be `FileNotFoundException`

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
dancodedotorg: 👋 Hi. Former high school geometry teacher here. Sorry for eavesdropping, but I might also suggest...
- +1 to Jessica's comment that this only works for a regular polygon
- I think x and y are the coordinates of the _center_ of the polygon. Phrasing it as "of the polygon" is ambiguous - is it one of the vertices? The center? etc
- I'd suggest using the term "radius" instead of "size". The definition - "distance from the center to each point" - matches the definition of the radius of a regular polygon, which is how it's taught in high school geometry - https://www.mathopenref.com/polygonradius.html.

org-code-javabuilder/theater/src/main/java/org/code/Pixel.java
moneppo: Great questions. There's a tension with this API between enabling a high ceiling and the fact that the theater will be used relatively early in the course. I think what you're suggesting is pretty important for making the theater sufficiently capable in the realm of color manipulation, but I'm not clear on exactly how colors will be exposed to students early on. I'll check with the writers to see what they think.

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
moneppo: Jamila indicated that using exceptions is fine as they can work around them with helper code if they need to.

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
moneppo: Jamila believes that the static approach is the most sensible, let's go for that.

org-code-javabuilder/theater/src/main/java/org/code/Image.java
jmkulwik: We can call it something different than "theater" but the package cannot be org.code. It needs a more descriptive name.

org-code-javabuilder/theater/src/main/java/org/code/Pixel.java
moneppo: After a discussion with Jamila, we decided that we're going to keep string-based colors for now and, depending on use cases and time, add APIs/objects that allow for more powerful color handling. Two ideas that came up were:
1. `Color` class that could be initialized with ints or a string
2. A converter method from ints to string and from string to ints, so either can be used in pixel operations and drawing operations.

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
moneppo: Agreed, I think a second function is the clearest pattern.

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
moneppo: Yes, I'll update it.

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
moneppo: Yes, good catch

org-code-javabuilder/theater/src/main/java/org/code/Image.java
moneppo: Recommendations? This will also be true for `SoundLoader`

jmkulwik: Perhaps `org.code.image` and `org.code.sound`? The curriculum team may have opinions here.

org-code-javabuilder/theater/src/main/java/org/code/Pixel.java
madelynkasula: that sounds great -- i agree with that approach completely 👍🏻

jamescodeorg: For the future, I think it would be useful to have constants defined somewhere for common colors (similar to [java.awt.Color](https://docs.oracle.com/en/java/javase/11/docs/api/java.desktop/java/awt/Color.html).  Of the two ideas above, I think I like 1 but directing students to use constants (e.g. Color.BLACK) rather than construct a new one using a string.

org-code-javabuilder/theater/src/main/java/org/code/SoundLoader.java
```diff
@@ -0,0 +1,17 @@
...
+public enum Instrument {
+  PIANO, BASS
+}
+
+public static class SoundLoader {
+  /**
+   * Extracts audio samples from WAV or AU file into array
+   * 
+   * @param filename the name of the audio file
+   * @return the array of samples
```
jamescodeorg, line 12: I'm not very familiar with how sound is encoded.  Could you elaborate a bit on what the values in this array represent?

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
```diff
@@ -0,0 +1,202 @@
+package org.code.theater;
+
+public enum Instrument {
```
jamescodeorg, line 3: This is also defined in SoundLoader.java?

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
jamescodeorg: Sorry to reopen this, but I feel very strongly that a class with getters and setters like this one shouldn't be static.  This will end up being very confusing for the student where these types of objects in Java are almost always instances.

As a compromise, what if we had a regular constructor, but it could only be called once?  Calling it a second time would throw an exception.  The student could then write `theater = new Theater()` to get a Theater, similar to Painter.

(I'm happy to chat with Jamila directly if that's easier.)

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
```diff
@@ -0,0 +1,202 @@
...
+    * 
+    * @param file     the name of the file in the asset manager
+    * @param x        the left side of the image in the canvas
+    * @param y        the top of the image in the canvas
+    * @param width    the width to draw the image on the canvas
+    * @param height   the height to draw the image on the canvas
+    * @param rotation the amount to rotate the image
+    * @throws FileNotFoundException if the file can't be found in the project.
+    */
+   public static void drawImage(String filename, double x, double y, double width, double height, double rotation)
```
jamescodeorg, line 79: As a general comment, all of these draw commands seem to be pretty complicated and take a lot of parameters that are the same type.  (This makes is easy to accidentally pass the wrong value for a parameter.)  Not sure how best to mitigate this, maybe we can have "simple" versions of these methods or make some of the parameters more stateful (similar to how fillColor is stateful)?

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
```diff
@@ -0,0 +1,202 @@
...
+    */
+   public static double getHeight() {
+   }
+   
+   /**
+    * Plays the array of samples provided.
+    * 
+    * @param sound an array of samples to play.
+    */
+   public static void playSound(double[] sound) {
```
jamescodeorg, line 27: Do we want to consider having a Sound class similar to how we have an Image class?  I think this would be more consistent and give us a lot more flexibility going forward.

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
```diff
@@ -0,0 +1,202 @@
...
+    * 
+    * @param color CSS color string to fill any shape.
+    */
+   public static void setFillColor(String color) {
+   }
+ 
+   /**
+    * Removes the fill color so any shapes have no fill.
+    */
+   public static void removeFill() {
```
jamescodeorg, line 194: Should this be `removeFillColor` for consistency with `setFillColor`?

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
```diff
@@ -0,0 +1,202 @@
...
+    * the width and height provided
+    * 
+    * @param img      the Image object to draw on the canvas
+    * @param x        the left side of the image in the canvas
+    * @param y        the top of the image in the canvas
+    * @param width    the width to draw the image on the canvas
+    * @param height   the height to draw the image on the canvas
+    * @param rotation the amount to rotate the image
+    */
+   public static void drawImage(Image img, double x, double y, double width, double height, double rotation) {
```
jamescodeorg, line 94: (nit) We do a good job of not abbreviating variable names elsewhere so maybe we can use `image` instead of `img` here?

jamescodeorg (commented):
Is there a sample program somewhere that represents something at the upper end of complexity of what we expect students to write?  My general feeling is that writing something more complex using this API will be a bit unwieldy but perhaps we just aren't expecting students to build things that are that complex?

org-code-javabuilder/theater/src/main/java/org/code/Pixel.java
moneppo: I discussed some of this with Jamila and we think there might be a way to use a `Color` class that makes the transition from string usage in Neighborhood easier using constants as you describe. I'll push an update with the proposal, stay tuned.

org-code-javabuilder/theater/src/main/java/org/code/Image.java
```diff
@@ -0,0 +1,68 @@
...
+     * Create a new image object from the pixels provided. If the number of pixels
+     * is not equal to the width multiplied by the height, the remaining pixels will
+     * be filled with black. If there are too many pixels, the remaining pixels will
+     * be cut off in the image.
+     * 
+     * @param pixels the pixels with which to create the image.
+     * @param width  the width of the image to create.
+     * @param height the height of the image to create.
+     */
+    public Image(Pixel[] pixels, int width, int height) {
```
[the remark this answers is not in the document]
moneppo: Looks like I had been carrying some old assumptions about pixel access - this comment sparked a discussion and we realized my mistake! Yes, I think direct pixel access using `setPixel` and `getPixel` look like a more appropriate system for this.

org-code-javabuilder/theater/src/main/java/org/code/Image.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /**
+     * Creates an empty image filled with white pixels.
+     * 
+     * @param width  the width of the image to create.
+     * @param height the height of the image to create.
+     */
+    public Image(int width, int height) {
```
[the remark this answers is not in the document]
moneppo: We could have both, e.g. one that defaults to white and one that takes a color. I'll write it up and see how it looks

org-code-javabuilder/theater/src/main/java/org/code/Pixel.java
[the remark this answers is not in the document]
moneppo: Yes!

org-code-javabuilder/theater/src/main/java/org/code/Pixel.java
```diff
@@ -0,0 +1,57 @@
...
+    public int getBlue() {
+    }
+
+    /**
+     * Sets the amount of red (ranging from 0 to 255) in the color of the pixel.
+     * 
+     * @param value the amount of red (ranging from 0 to 255) in the color of the
+     *              pixel.
+     */
+    public void setRed(int value) {
```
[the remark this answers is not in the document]
moneppo: Good catch, I'll document the behavior (for the record the plan is clamping, e.g. `setRed(100000)` will result in a red value of 255)

org-code-javabuilder/theater/src/main/java/org/code/SoundLoader.java
moneppo: Each represents a sample value, which you can think of like the position of the speaker at a point in time. Samples played back at a fixed rate (44.1kHz) makes CD-quality sound. This will be taught in the course, I'll add this description to the javadoc

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
moneppo: Oops

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
moneppo: Chatted with Jamila, she thinks this is good idea. While I did think this made the easiest to use API, it's using static methods to "hack" in global state, so yea, I think this is a worthwhile change.

org-code-javabuilder/theater/src/main/java/org/code/Image.java
moneppo: `org.code.media` seems to be acceptable to everyone

moneppo:
> Is there a sample program somewhere that represents something at the upper end of complexity of what we expect students to write? My general feeling is that writing something more complex using this API will be a bit unwieldy but perhaps we just aren't expecting students to build things that are that complex?

Right now we're trying to build the plane while it's flying a bit, so we don't have much good example code. The best thing we have are existing Lab code which use very similar patterns, specifically Game Lab as the design of the drawing API is based on its drawing blocks. The key differences are that the Theater isn't interactive, and will have a low update rate (up to 10 frames per second) compared to Game Lab (60 frames per second).

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
moneppo: We opted for direct use of `double[]` as the lessons that use sounds are very array-oriented and this type makes the lessons clearer. We may add a sound object in the future, depending on how our use cases outside of the course take shape.

jamescodeorg:
One more random thought, do we want to consider supporting an alternate version of the API that uses the spelling `theatre` (since that might be more popular outside of the U.S.)?

jmkulwik:
> Is there a sample program somewhere that represents something at the upper end of complexity of what we expect students to write? My general feeling is that writing something more complex using this API will be a bit unwieldy but perhaps we just aren't expecting students to build things that are that complex?

@jamescodeorg - this doesn't use this exact API, but it was one of the early concepts of what a student might build: https://codedotorg.slack.com/archives/C01JAMPCQT1/p1620868048030700

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
moneppo: Lots of discussion went back and forth on the design of these methods :) 

These are based on the Game Lab & App Lab drawing commands, which are in turn based on processing.org/p5js.org drawing commands. Those tools are widely used in introductory courses in Java and Javascript respectively. In my experience and reports I've heard from teachers that use those tools, the long list of identical parameters can be confusing at first but students get used to them pretty quickly. Stateful systems like the HTML5 canvas which use `moveTo` and `lineTo` have some other comprehension issues associated with them, and we decided that a (somewhat) stateless model for drawing would be best for this audience.

jamescodeorg: Thanks for the context, sounds good to me!

jmkulwik:
FYI @moneppo these are the commands we'll be likely to be able to support given our recent discoveries about "headless" mode for Java graphics. https://www.oracle.com/technical-resources/articles/javase/headless.html

org-code-javabuilder/theater/src/main/java/org/code/media/Color.java
```diff
@@ -0,0 +1,110 @@
...
+    /**
+     * Sets the amount of blue (ranging from 0 to 255).
+     * 
+     * @param value the amount of blue (ranging from 0 to 255) in the color of the
+     *              pixel.
+     */
+    public void setBlue(int value) {
+    }
+
+    public static final Color white = new Color(255, 255, 255);
```
jamescodeorg, line 81: I think the convention for Java constants is UPPER_SNAKE_CASE?

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
```diff
@@ -0,0 +1,210 @@
...
+public enum Instrument {
+   PIANO, BASS
+}
+
+public class Stage {
+   /**
+    * Returns the width of the theater canvas. Right now this will always be 400
+    * pixels.
+    */
+   public double getWidth() {
```
jamescodeorg, line 12: Will we be working in fractional pixels?  Wondering if width and height should be ints or doubles?  How about x and y?  (This applies to all APIs in this class.)

jamescodeorg (commented):
A few more follow-up comments.

org-code-javabuilder/theater/src/main/java/org/code/media/Color.java
moneppo: `awt.Color` offers both so I chose lowercase: https://docs.oracle.com/javase/7/docs/api/java/awt/Color.html

That isn't to say it's the right choice :) I'll confer with curriculum, I think all caps can be a little off-putting but then again I used it for the instruments enum...

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
moneppo: AP CSA only teaches `double` as the primitive numeric type 🤷

org-code-javabuilder/theater/src/main/java/org/code/media/Color.java
jamescodeorg: I think the uppercase ones are the "correct" ones: https://stackoverflow.com/questions/15258824/how-come-java-has-two-color-values-for-each-color.

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
jamescodeorg: I see `int` and `Integer` referenced in the AP CSA Java subset?  https://secure-media.collegeboard.org/digitalServices/pdf/ap/ap-computer-science-a-java-subset.pdf

moneppo: Oh you're right, I had been thinking floats. Yes we should move these to `int`.

org-code-javabuilder/theater/src/main/java/org/code/media/Color.java
moneppo: I quickly checked and curriculum is okay with uppercase, so Molly has a change already for that :)

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
moneppo: @molly-moen can you move the getWidth/getHeight methods to return ints? 

All other methods that refer to x & y should remain doubles so you can draw shapes at fractional positions.

PR #53, opened 2021-06-11
Theater: Add GifWriter

molly-moen:
Add a new GifWriter class which writes `BufferedImage`s to a `ByteArrayOutputStream` to create a gif. Each frame of the gif specifies the delay between it and the next frame. That gif can then be sent as an encoded Base64 String to the client. 

`GifWriter` can only generate one gif per instantiation, and the consuming class must call `close` in order to complete the gif. The `CatImage` sample program now includes a `play` method that closes the gif and sends it to the client. This PR also updates `ImageEncoder` to accept a `ByteArrayOutputStream` instead of a single `BufferedImage`.

Note: we will need to update the code-dot-org client to accept a gif properly, as of now it uses a canvas which will only display the first frame of the gif. We can update to use an `img` to show the gif fully.

### DIFF
--- a/org-code-javabuilder/theater/src/main/java/org/code/theater/GifWriter.java
+++ b/org-code-javabuilder/theater/src/main/java/org/code/theater/GifWriter.java
@@ -8,34 +8,48 @@ import javax.imageio.metadata.IIOInvalidTreeException;
 import javax.imageio.metadata.IIOMetadata;
 import javax.imageio.metadata.IIOMetadataNode;
 import javax.imageio.stream.ImageOutputStream;
+import org.code.protocol.InternalErrorKey;
+import org.code.protocol.InternalJavabuilderError;
 
-// Writer to generate a gif from a set of images.
+// Writer to generate a gif from a set of images. The gif will be stored in the given
+// ByteArrayOutputStream.
 // Adapted from https://memorynotfound.com/generate-gif-image-java-delay-infinite-loop-example/
 public class GifWriter {
   private ImageWriter writer;
   private ImageWriteParam params;
   private ImageOutputStream imageOutputStream;
 
-  public GifWriter(ByteArrayOutputStream out) throws IOException {
-    this.imageOutputStream = ImageIO.createImageOutputStream(out);
-    this.writer = ImageIO.getImageWritersBySuffix("gif").next();
-    this.params = writer.getDefaultWriteParam();
-
-    this.writer.setOutput(this.imageOutputStream);
-    this.writer.prepareWriteSequence(null);
+  public GifWriter(ByteArrayOutputStream out) {
+    try {
+      this.imageOutputStream = ImageIO.createImageOutputStream(out);
+      this.writer = ImageIO.getImageWritersBySuffix("gif").next();
+      this.params = writer.getDefaultWriteParam();
+      this.writer.setOutput(this.imageOutputStream);
+      this.writer.prepareWriteSequence(null);
+    } catch (IOException e) {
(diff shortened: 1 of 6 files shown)


org-code-javabuilder/theater/src/main/java/org/code/theater/GifWriter.java
```diff
@@ -0,0 +1,113 @@
...
+    } catch (IOException e) {
+      throw new InternalJavabuilderError(InternalErrorKey.INTERNAL_RUNTIME_EXCEPTION, e.getCause());
+    }
+  }
+
+  /**
+   * Write the given image as the next frame of the gif.
+   *
+   * @param img BufferedImage
+   * @param delay delay in milliseconds until the next frame of the gif.
```
jmkulwik, line 39: Does the delay come before or after the current frame that is being added?

org-code-javabuilder/theater/src/main/java/org/code/theater/GifWriter.java
```diff
@@ -0,0 +1,113 @@
...
+  /**
+   * Update the metadata for the next frame with the given delay and image type. See Gif metadata
+   * specification here:
+   * https://javadoc.scijava.org/Java7/javax/imageio/metadata/doc-files/gif_metadata.html
+   *
+   * @param delay int delay in milliseconds after this frame
+   * @param imageType int
+   * @return Updated IIOMetadata for the next frame
+   */
+  private IIOMetadata getMetadataForFrame(int delay, int imageType) {
```
jmkulwik, line 72: Should this be `updateMetadataForFrame`?

org-code-javabuilder/theater/src/main/java/org/code/theater/GifWriter.java
```diff
@@ -0,0 +1,113 @@
...
+   * Find metadata node with the given name from the rootNode
+   *
+   * @param rootNode
+   * @param nodeName
+   * @return Requested IIOMetadataNode
+   */
+  private static IIOMetadataNode getNode(IIOMetadataNode rootNode, String nodeName) {
+    int nNodes = rootNode.getLength();
+    for (int i = 0; i < nNodes; i++) {
+      if (rootNode.item(i).getNodeName().equalsIgnoreCase(nodeName)) {
```
jmkulwik, line 105: Is there any possibility of name collisions here?

org-code-javabuilder/theater/src/main/java/org/code/theater/GifWriter.java
molly-moen: it's after the frame being added

org-code-javabuilder/theater/src/main/java/org/code/theater/GifWriter.java
molly-moen: I think get is actually more accurate, I think I will update the comment. It takes the default metadata, makes some changes, and returns that.

org-code-javabuilder/theater/src/main/java/org/code/theater/GifWriter.java
molly-moen: I don't think so. The metadata is in a very specific format, so there should only be one node per name https://javadoc.scijava.org/Java7/javax/imageio/metadata/doc-files/gif_metadata.html

org-code-javabuilder/theater/src/main/java/org/code/theater/GifWriter.java
molly-moen: I'll make the comment more clear

jmkulwik (approved):
LGTM! Nice work!

PR #54, opened 2021-06-14
Add git hook back in and ensure mini apps are built when the websocket server is launched

jmkulwik:
This cleans up our build.gradle to do a couple things:

1. Ensure the pre-commit git hook gets added back in. We now run tests and the linter every commit. This can be skipped by using `git commit --no-verify`
2. Ensure the Theater and Neighborhood mini apps get built when the appRun command is triggered.


molly-moen (approved):
thanks!

PR #55, opened 2021-06-15
Neighborhood: Add a no-parameters canMove function

sanchitmalhotra126:
Adds a no-parameter version of canMove() to the Neighborhood API.

Tested with unit tests and local environment.

[CSA-462](https://codedotorg.atlassian.net/browse/CSA-462)

### DIFF
--- a/org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/Painter.java
+++ b/org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/Painter.java
@@ -154,6 +154,11 @@ public class Painter {
     return this.isValidMovement(Direction.fromString(direction));
   }
 
+  /** @return True if there is no barrier one square ahead in the current direction. */
+  public boolean canMove() {
+    return this.isValidMovement(this.direction);
+  }
+
   /** @return the color of the square where the painter is standing. */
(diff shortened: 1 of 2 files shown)


jmkulwik (approved):
Nice! I just reached out for clarity from curriculum on whether they'd like this to replace the old canMove function or to exist in addition to it. That is non-blocking for this PR as we can always remove the other one in a follow-up. Up to you on whether you'd like to wait for their response or not for merging this change.
Ask: https://codedotorg.slack.com/archives/C01EF4GJ9GE/p1623789933389500?thread_ts=1623684410.357100&cid=C01EF4GJ9GE

molly-moen (approved):
🎉

jmkulwik:
> Nice! I just reached out for clarity from curriculum on whether they'd like this to replace the old canMove function or to exist in addition to it. That is non-blocking for this PR as we can always remove the other one in a follow-up. Up to you on whether you'd like to wait for their response or not for merging this change.
> Ask: https://codedotorg.slack.com/archives/C01EF4GJ9GE/p1623789933389500?thread_ts=1623684410.357100&cid=C01EF4GJ9GE

Curriculum confirmed - we should keep both around.

PR #56, opened 2021-06-17
Theater: Update gif writer to use correct dimensions

molly-moen:
In testing gif generation I found that the generated gif would be the dimensions of whatever the first image in the gif was. This is not ideal if the first image is smaller than the second image. This changes sets the gif to be 400x400 pixels, which is the size we will communicate with students.

~One note here is if an image is added that is greater than 400x400, it will be cropped, not resized. We should determine if this is the desired behavior. We could either update the gif writer or part of the future Theater API if we determine resizing is the desired behavior. We also could parameterize the desired size of the gif in the future.~ Determined we are good with cropping as the theater api forces students to pick a size.


jmkulwik (approved):
Looks great! Thanks for making this fix!

PR #58, opened 2021-06-17
Painter: Throw an exception when attempting to paint on a bucket

sanchitmalhotra126:
Throws an exception when attempting to paint on a bucket (i.e. when the current square has paint).

Testing: unit test and localhost dashboard


molly-moen (approved):
Thanks!

PR #59, opened 2021-06-21
Add Theater API

molly-moen:
Add the Theater API to javabuilder. Currently the api has no functionality beyond all classes are available to be called, and calling `Theater.stage.play()` will print out `play`.

See [this PR](https://github.com/code-dot-org/javabuilder/pull/51) for full discussion on the API.


org-code-javabuilder/media/build.gradle
```diff
@@ -0,0 +1,23 @@
+plugins {
+    // Apply the java-library plugin for API and implementation separation.
+    id 'java-library'
```
jmkulwik, line 3: Also add `id 'com.github.sherter.google-java-format' version '0.8'` to enable the linter.

jmkulwik (approved):
LGTM!

PR #60, opened 2021-06-22
Investigation for headless mode visual commands

jmkulwik:
This demos some of the features that are possible in headless mode and sets `System.setProperty("java.awt.headless", "true");` which should make testing in headless mode easier. This can exist just as a draft PR for reference or can be merged if helpful.

I removed ImageJ in favor of just using awt. There's more documentation on running awt headless. 
Some docs:
https://www.baeldung.com/java-headless-mode 
https://www.oracle.com/technical-resources/articles/javase/headless.html


molly-moen (approved):
Thanks! I think we should merge this since it's an improvement over what we had before and does some cleanup.

PR #61, opened 2021-06-22
Theater API: Change constants to uppercase

molly-moen:
We decided to change the color constants to be uppercase instead of lowercase.


sanchitmalhotra126 (approved):
LGTM

PR #62, opened 2021-06-24
Add audio read/write capabilities to Theater (WAV only)

sanchitmalhotra126:
Adds capabilities to read and write audio data in Theater. Currently only supports wav files at 44.1kHz and 16-bit. I'm investigating MP3 and FLAC support separately, but this sets up the main reading and writing pieces we need for the API.

This also uses the same encoding method as GifWriter in that it writes the final audio output stream to a base64 string and send that to the client. I imagine this implementation will change when we write files to S3 instead, but it should only be a matter of writing out to a file instead of an output stream.

Also includes `CatMusic.java`, similar to `CatImage.java` for testing.

Unit tests included for the utility classes, but held off on tests for the writer since that implementation may change. But I can definitely add more tests based on feedback.


org-code-javabuilder/media/src/main/java/org/code/media/AudioWriter.java
```diff
@@ -0,0 +1,65 @@
...
+    final byte[] bytes = AudioUtils.convertDoubleArrayToByteArray(samples);
+    try {
+      this.rawOutputStream.write(bytes);
+    } catch (IOException e) {
+      throw new SoundException(e);
+    }
+  }
+
+  public void writeAudioFile(String filename) throws SoundException {
+    writeAudioSamples(SoundLoader.read(filename));
```
molly-moen, line 34: nit: `this.writeAudioSamples`

org-code-javabuilder/media/src/main/java/org/code/media/AudioWriter.java
```diff
@@ -0,0 +1,65 @@
...
+   *
+   * @throws SoundException
+   */
+  public void writeToAudioStreamAndClose() throws SoundException {
+    final byte[] bytes = this.rawOutputStream.toByteArray();
+    AudioUtils.writeBytesToOutputStream(bytes, this.audioOutputStream);
+
+    try {
+      this.rawOutputStream.close();
+      this.audioOutputStream.close();
```
molly-moen, line 55: for my information: is the reason we need two streams that we need to write the `audioOutputStream` all at once?

org-code-javabuilder/media/src/main/java/org/code/media/SoundLoader.java
```diff
@@ -10,6 +17,43 @@
    * @throws SoundException when the file cannot be found
    */
   public static double[] read(String filename) throws SoundException {
```
molly-moen, line 19: I wonder if we should throw a `FileNotFoundException` if the file isn't found and a `SoundException` if there's an issue with the file? This would have consistency with how we plan to handle exceptions for images.

org-code-javabuilder/media/src/main/java/org/code/media/SoundLoader.java
```diff
@@ -10,6 +17,43 @@
...
+    if (filename == null) {
+      throw new SoundException("Missing file name");
+    }
+
+    try {
+      final File file = new File(filename);
+      if (file.exists()) {
+        return AudioSystem.getAudioInputStream(file);
+      }
+      throw new SoundException("File does not exist: " + filename);
```
molly-moen, line 54: I think we want to send keys (see Neighborhood's `ExceptionKey` class) so we can translate messages in the future, is that right @jmkulwik ?

molly-moen (commented):
This is great! Just a couple comments

org-code-javabuilder/media/src/main/java/org/code/media/AudioWriter.java
sanchitmalhotra126: Yep, so `rawOutputStream` just collects the raw audio samples themselves, but when writing to a file we also need the appropriate file headers, etc which `AudioSystem.write()` provides for us. So `audioOutputStream` is used to write out the full contents of the audio file.

I think we could use one output stream if we collect the raw bytes, reset the stream, and write the full file back to that stream? Although I think eventually this will write out to a file rather than an output stream so this may go away.

org-code-javabuilder/media/src/main/java/org/code/media/SoundLoader.java
sanchitmalhotra126: Yeah, makes sense, I'll add that

org-code-javabuilder/media/src/main/java/org/code/media/SoundLoader.java
sanchitmalhotra126: Yep good call - would we want to catch the `SoundException` in the theater package and add the key there, or do it here? It may get a little tricky with dependencies since theater depends on media here, so we wouldn't be able to use a Theater-specific key I think?

molly-moen: I wonder if we should add a `MediaRuntimeException` like we do for neighborhood and theater? Curious for Jessie's thoughts since she did some work on our exception handling overall. Although I like the idea of having special exceptions in the method header for educational reasons.

molly-moen (approved):
LGTM! Still that minor question on exception keys, but that may warrant further discussion and isn't blocking

org-code-javabuilder/media/src/main/java/org/code/media/SoundLoader.java
jmkulwik: Agreed, we'll want to send keys - that's generally how we plan to translate and also how we plan to make more user-friendly error messages. I think it would make sense to add a Media Exception and keys of some sort. It should extend JavabuilderException or JavabuilderRuntimeException - depending on whether we want to require the student to handle it or not. It should be set up similar to NeighborhoodRuntimeException & ExceptionKeys from the neighborhood package (although the naming convention used for the ExceptionKeys class could probably use another rev).

sanchitmalhotra126: Gotcha - where does that leave `SoundException` then? Would we wrap this in a `MediaRuntimeException` with a key corresponding to audio and the message?

sanchitmalhotra126: Looking at this a bit more - I think I'll remove `SoundException` in favor of `MediaRuntimeException` and throw MediaRuntimeExceptions/InternalJavaBuilderExceptions accordingly.

jmkulwik: `SoundException` itself can inherit from `JavabuilderException`. Then you'd need to make `SoundExceptionKey` entries for each unique message that the user will see. i.e. `"Error reading audio file"` could be `INVALID_AUDIO_FILE` or something similar. The string itself would live on the client in the i18n folder for Java Lab: https://github.com/code-dot-org/code-dot-org/blob/staging/apps/i18n/javalab/en_us.json

sanchitmalhotra126: Ah okay cool, yeah that makes sense. Will add.

jmkulwik: `SoundException` is specified in some parts of the student-facing API, so we'll need to have it for at least those areas. I believe Theater is one of the places where students will be learning about how exceptions work, so the SoundException will facilitate that learning.

sanchitmalhotra126: Updated with exception keys for `SoundException`

PR #63, opened 2021-06-24
Theater API: implement geometry drawing methods

molly-moen:
Implement all the non-image drawing methods in theater (such as `drawLine`, `drawShape`, etc.). To implement this I also implemented most of the `Color` class, excluding initializing a color via a string. I also added a new `TheaterRuntimeException` and started unit tests for `Stage`.

The drawing is done by creating a blank [BufferedImage](https://docs.oracle.com/en/java/javase/11/docs/api/java.desktop/java/awt/image/BufferedImage.html) and getting the [Graphics2D](https://docs.oracle.com/en/java/javase/11/docs/api/java.desktop/java/awt/Graphics2D.html) associated with that image. We will be able to use the Graphics2D to add images in Part 2 of this work. Every time `pause` is called on `Stage`, we write the existing image to the gif with the specified pause. When play is called we write the final frame and send it to the client. Our method of sending gifs may change to accommodate larger gifs.

### DIFF
--- a/org-code-javabuilder/media/src/main/java/org/code/media/Color.java
+++ b/org-code-javabuilder/media/src/main/java/org/code/media/Color.java
@@ -13,7 +13,9 @@ public class Color {
    * @throws IllegalArgumentException if the value specifies an unsupported color name or illegal
    *     hexadecimal value
    */
-  public Color(String color) throws IllegalArgumentException {}
+  public Color(String color) throws IllegalArgumentException {
+    // TODO: Implement this method
+  }
 
   /**
    * Create a new color based on the red, green, and blue values provided.
--- a/org-code-javabuilder/media/src/main/java/org/code/media/Color.java
+++ b/org-code-javabuilder/media/src/main/java/org/code/media/Color.java
@@ -2,6 +2,10 @@ package org.code.media;
 
 public class Color {
 
+  private int red;
+  private int green;
+  private int blue;
+
   /**
    * Creates a color from a string representation.
    *
@@ -18,7 +22,11 @@ public class Color {
    * @param green the green value from 0 - 255
    * @param blue the blue value from 0 - 255
    */
-  public Color(int red, int green, int blue) {}
+  public Color(int red, int green, int blue) {
+    this.red = red;
+    this.green = green;
+    this.blue = blue;
+  }
 
   /**
    * Returns the amount of red (ranging from 0 to 255).
@@ -26,7 +34,7 @@ public class Color {
    * @return a number representing the red value (between 0 and 255)
    */
   public int getRed() {
-    return -1;
+    return this.red;
   }
 
   /**
@@ -35,7 +43,7 @@ public class Color {
    * @return a number representing the green value (between 0 and 255) of the pixel.
    */
   public int getGreen() {
-    return -1;
+    return this.green;
   }
 
   /**
@@ -44,7 +52,7 @@ public class Color {
    * @return a number representing the blue value (between 0 and 255)
    */
   public int getBlue() {
-    return -1;
+    return this.blue;
   }
 
   /**
@@ -53,7 +61,9 @@ public class Color {
    *
    * @param value the amount of red (ranging from 0 to 255) in the color of the pixel.
    */
-  public void setRed(int value) {}
+  public void setRed(int value) {
+    this.red = value;
+  }
 
   /**
    * Sets the amount of green (ranging from 0 to 255). Values below 0 will be ignored and set to 0,
@@ -61,14 +71,18 @@ public class Color {
    *
    * @param value the amount of green (ranging from 0 to 255) in the color of the pixel.
    */
-  public void setGreen(int value) {}
+  public void setGreen(int value) {
+    this.green = value;
+  }
 
   /**
    * Sets the amount of blue (ranging from 0 to 255).
    *
    * @param value the amount of blue (ranging from 0 to 255) in the color of the pixel.
    */
-  public void setBlue(int value) {}
+  public void setBlue(int value) {
+    this.blue = value;
+  }
 
   public static final Color WHITE = new Color(255, 255, 255);
   public static final Color SILVER = new Color(192, 192, 192);
--- a/org-code-javabuilder/media/src/main/java/org/code/media/Color.java
+++ b/org-code-javabuilder/media/src/main/java/org/code/media/Color.java
@@ -25,9 +25,9 @@ public class Color {
    * @param blue the blue value from 0 - 255
    */
   public Color(int red, int green, int blue) {
-    this.red = red;
-    this.green = green;
-    this.blue = blue;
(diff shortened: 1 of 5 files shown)


org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
```diff
@@ -1,17 +1,63 @@
...
+import org.code.media.Color;
+import org.code.media.Image;
+import org.code.protocol.GlobalProtocol;
+import org.code.protocol.OutputAdapter;
 
 public class Stage {
-  /** Returns the width of the theater canvas. Right now this will always be 400 pixels. */
-  public double getWidth() {
-    return -1;
+  private BufferedImage image;
```
sanchitmalhotra126, line 13: can also be final right?

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
```diff
@@ -102,7 +154,14 @@ public void drawText(
    * @param endX the end X coordinate of the line.
    * @param endY the end Y coordinate of the line.
    */
-  public void drawLine(double startX, double startY, double endX, double endY) {}
+  public void drawLine(int startX, int startY, int endX, int endY) {
+    if (this.strokeColor == null) {
```
sanchitmalhotra126, line 158: Curious - any reason we allow drawing lines without a stroke color set but not other shapes?

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
```diff
@@ -112,7 +171,25 @@ public void drawLine(double startX, double startY, double endX, double endY) {}
    * @param sides the number of sides of the polygon
    * @param radius the distance from the center to each point on the polygon
    */
-  public void drawRegularPolygon(double x, double y, int sides, double radius) {}
+  public void drawRegularPolygon(int x, int y, int sides, int radius) {
+    if (this.strokeColor != null || this.fillColor != null) {
```
sanchitmalhotra126, line 175: nit: could invert and early return to reduce nesting a bit

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
```diff
@@ -45,13 +91,21 @@ public void playNote(Instrument instrument, int note, double seconds) {}
...
    *     smallest value can be .1 seconds.
    */
-  public void wait(double seconds) {}
+  public void pause(double seconds) {
+    this.gifWriter.writeToGif(this.image, (int) (seconds * 1000));
+  }
 
   /** Clear everything, starting from the beginning and emptying the canvas. */
-  public void reset() {}
+  public void reset() {
```
sanchitmalhotra126, line 99: Should we also reset `hasPlayed` here? Or do we only allow one `play` per program?

sanchitmalhotra126 (approved):
A couple small things I noticed but otherwise LGTM!

org-code-javabuilder/media/src/main/java/org/code/media/Color.java
```diff
@@ -53,30 +63,36 @@ public int getBlue() {
    *
    * @param value the amount of red (ranging from 0 to 255) in the color of the pixel.
    */
-  public void setRed(int value) {}
+  public void setRed(int value) {
+    this.red = value;
+  }
```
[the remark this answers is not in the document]
molly-moen: good call, I'll update this

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
```diff
@@ -1,17 +1,63 @@
...
+    this(new BufferedImage(WIDTH, HEIGHT, BufferedImage.TYPE_INT_RGB));
+  }
+
+  /**
+   * Initialize Stage with a specific BufferedImage. Used directly for testing. Stage should be
+   * initialized outside of org.code.theater using Theater.stage.
+   *
+   * @param image
+   */
+  protected Stage(BufferedImage image) {
```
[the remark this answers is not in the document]
molly-moen: currently we don't (we can already mock the output adapter because it's being creating by GlobalProtocol), and can mock the graphics by mocking `createGraphics` on `image`.

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
molly-moen: We only allow one play per program, reset is really more of a clear image than a full reset. I will update the comment and check with curriculum that we want to keep this method name.

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
molly-moen: Yes, this was a request from curriculum that it would be a special case here, since lines only have strokes and not fills.

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
molly-moen: yep, will update

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
```diff
@@ -143,37 +262,67 @@ public void drawEllipse(double x, double y, double width, double height) {}
...
+    } else {
+      this.gifWriter.writeToGif(this.image, 0);
+      this.gifWriter.close();
+      outputAdapter.sendMessage(ImageEncoder.encodeStreamToMessage(this.outputStream));
+      this.hasPlayed = true;
+    }
+  }
+
+  private java.awt.Color convertColor(Color color) {
+    return new java.awt.Color(color.getRed(), color.getGreen(), color.getBlue());
```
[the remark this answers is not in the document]
molly-moen: I also pondered this, I didn't think about a static method which seems like a better pattern then what I have here. If the user finds it they wouldn't be able to break anything so I wouldn't be too worried about this.

org-code-javabuilder/media/src/main/java/org/code/media/Color.java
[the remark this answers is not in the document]
molly-moen: ah I noticed in the comment we want to ignore out of range values and set to 0 if less than 0 and 255 if greater than 255. Will update

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
molly-moen: full conversation is [here](https://codedotorg.slack.com/archives/C01EF4GJ9GE/p1624553158138800) short version is we are renaming to `clear` and adding a Color parameter for the new background color

sanchitmalhotra126 (approved):
LGTM!

PR #64, opened 2021-06-28
Javabuilder: Retrieve assets and update how we get the sources url

molly-moen:
Previously we sent the sources url over to Javabuilder as a session parameter in order to retrieve project files. Now that we need both the assets url (for things like images and audio) and the sources url, we will instead use the parameters (already sent by the javabuilder session token) `channelId` and `iss` (dashboard hostname) in order to generate the urls we need. 

The url generation is done via the global protocol, which is available to all packages. We only generate the url via the protocol instead of load the file because the file loading logic is dependent on the type of file we want to load (for example, for images we can use an `ImageIO` helper method).

This PR also includes the initial change to use the assets url to load an image into memory. This functionality will be used and expanded upon in a follow-up PR for theater images.

## Links
- jira ticket: [https://codedotorg.atlassian.net/browse/CSA-424](https://codedotorg.atlassian.net/browse/CSA-424)

### DIFF
--- a/api-gateway-routes/api_gateway_proxy_function.rb
+++ b/api-gateway-routes/api_gateway_proxy_function.rb
@@ -41,10 +41,10 @@ def on_connect(event, context)
     :queueUrl => sqs_queue.queue_url,
     :apiEndpoint => api_endpoint,
     :connectionId => request_context["connectionId"],
-    :projectUrl => authorizer["project_url"],
     :levelId => authorizer["level_id"],
     :options => authorizer["options"],
-    :iss => authorizer["iss"]
+    :iss => authorizer["iss"],
+    :channelId => authorizer["channel_id"]
   }
   response = lambda_client.invoke({
     function_name: 'javaBuilderExecuteCode:11',
--- a/org-code-javabuilder/media/src/main/java/org/code/media/Image.java
+++ b/org-code-javabuilder/media/src/main/java/org/code/media/Image.java
@@ -1,8 +1,15 @@
 package org.code.media;
 
+import java.awt.image.BufferedImage;
 import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.net.URL;
+import javax.imageio.ImageIO;
+import org.code.protocol.GlobalProtocol;
 
 public class Image {
+  private BufferedImage bufferedImage;
+
   /**
    * Creates a new image object, using the pixel information from the file uploaded to the asset
    * manager.
@@ -10,7 +17,15 @@ public class Image {
    * @param filename the name of the image loaded into the asset manager for the project
    * @throws FileNotFoundException if the file doesn't exist in the asset manager.
    */
-  public Image(String filename) throws FileNotFoundException {}
+  public Image(String filename) throws FileNotFoundException {
+    try {
+      this.bufferedImage =
+          ImageIO.read(new URL(GlobalProtocol.getInstance().generateAssetUrl(filename)));
+    } catch (IOException e) {
+      // TODO: improve error handling
+      throw new FileNotFoundException();
+    }
+  }
 
   /**
    * Create a new image object, copying the source image provided.
@@ -63,7 +78,7 @@ public class Image {
    * @return the width of the image in pixels.
    */
   public int getWidth() {
-    return -1;
+    return this.bufferedImage.getWidth();
   }
 
   /**
@@ -72,7 +87,7 @@ public class Image {
    * @return the height of the image in pixels.
    */
   public int getHeight() {
-    return -1;
+    return this.bufferedImage.getHeight();
   }
 
   /**
@@ -81,4 +96,8 @@ public class Image {
    * @param color the color with which to fill the image.
    */
   public void clear(Color color) {}
+
+  public BufferedImage getBufferedImage() {
+    return this.bufferedImage;
+  }
 }
--- a/org-code-javabuilder/protocol/src/main/java/org/code/protocol/GlobalProtocol.java
+++ b/org-code-javabuilder/protocol/src/main/java/org/code/protocol/GlobalProtocol.java
@@ -12,14 +12,27 @@ public class GlobalProtocol {
   private static GlobalProtocol protocolInstance;
   private final OutputAdapter outputAdapter;
   private final InputAdapter inputAdapter;
+  private final String dashboardHostname;
+  private final String channelId;
 
-  private GlobalProtocol(OutputAdapter outputAdapter, InputAdapter inputAdapter) {
+  private GlobalProtocol(
+      OutputAdapter outputAdapter,
+      InputAdapter inputAdapter,
+      String dashboardHostname,
+      String channelId) {
     this.outputAdapter = outputAdapter;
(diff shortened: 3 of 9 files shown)


api-gateway-routes/api_gateway_proxy_function.rb
```diff
@@ -41,10 +41,10 @@ def on_connect(event, context)
     :queueUrl => sqs_queue.queue_url,
     :apiEndpoint => api_endpoint,
     :connectionId => request_context["connectionId"],
-    :projectUrl => authorizer["project_url"],
     :levelId => authorizer["level_id"],
     :options => authorizer["options"],
-    :iss => authorizer["iss"]
+    :iss => authorizer["iss"],
```
jamescodeorg, line 46: What does iss stand for?

jamescodeorg (commented):
Am I understanding correctly that there will be a client side change to build urls from iss and channel id?

As a general principle, I prefer to do as little non-display related logic on the client as possible.  If we think there will only be 2 urls (or even 3 or 5 urls), I would consider sending each of them from the server instead of having the client build them from parts.  This way, if the server ever wants to change the url or, say, use a different url format for different environments, this can be done transparently to the client.

api-gateway-routes/api_gateway_proxy_function.rb
molly-moen: issuer, it's a 'reserved claim' for a jwt token so we can't make it more specific

jamescodeorg: Got it. If we're generally using this variable as the dashboard hostname, it might be nice to rename it for readability.

molly-moen:
@jamescodeorg how are you defining client and server in this case? If javabuilder is the client here then yes, we are building the url from parts now instead of passing it over. That is a good point, we were coming at it from feeling it was redundant to send 2 very similar urls (assets and sources). We also have to do some url building on this side anyway by adding `/filename` for each file the user requests. I don't have a super strong opinion on it though. If we think the url is at all likely to change I can see the value in sending it over as a whole instead of in parts.

api-gateway-routes/api_gateway_proxy_function.rb
```diff
@@ -41,10 +41,10 @@ def on_connect(event, context)
     :queueUrl => sqs_queue.queue_url,
     :apiEndpoint => api_endpoint,
     :connectionId => request_context["connectionId"],
-    :projectUrl => authorizer["project_url"],
```
jamescodeorg, line 44: For my education, where does `authorizer["project_url"]` get set?

molly-moen: it gets sent from the client [here](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/javalab/JavabuilderConnection.js#L31) and passed through to javabuilder in [javabuilder_sessions_controller](https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/app/controllers/javabuilder_sessions_controller.rb#L43). If we were to update it to send the asset url as well I think we would create that url in javabuilder_sessions_controller.

jamescodeorg: Hmmm.  Not to open a can of worms, but is it ok for us to trust the value that the client sends here?  What happens if the client sends a url to a different project?  What happens if the client sends a random url to another website?  (Feel free to ignore this comment if you don't think it's relevant now.)

jamescodeorg:
Good question!  Your answer to the other questions made me realize that I don't fully understand how this information is passed through the system so I probably shouldn't be giving a recommendation on this.  Please feel free to take whichever approach you think is best!

api-gateway-routes/api_gateway_proxy_function.rb
molly-moen: This is actually solved by this PR because we no longer use it. We could change the pattern to generate the urls on the session token generation side (since we have the same information we are passing to javabuilder there). I'll discuss with the CSA team and decide which route we want to go down.

api-gateway-routes/api_gateway_proxy_function.rb
jmkulwik: If I'm understanding correctly, we can't rename the value from the authorizer, but can rename the value we pass in the payload? If it's straightforward to do that rename as part of this change, I agree that it would be nice. I keep having to remind myself what iss means too.

org-code-javabuilder/media/src/main/java/org/code/media/Image.java
```diff
@@ -1,16 +1,31 @@
...
    * @throws FileNotFoundException if the file doesn't exist in the asset manager.
    */
-  public Image(String filename) throws FileNotFoundException {}
+  public Image(String filename) throws FileNotFoundException {
+    try {
+      this.bufferedImage =
+          ImageIO.read(new URL(GlobalProtocol.getInstance().generateAssetUrl(filename)));
+    } catch (IOException e) {
+      // TODO: improve error handling
+      throw new FileNotFoundException();
```
jmkulwik, line 26: Mind adding a task for this?

org-code-javabuilder/protocol/src/main/java/org/code/protocol/GlobalProtocol.java
```diff
@@ -37,4 +50,12 @@ public OutputAdapter getOutputAdapter() {
   public InputAdapter getInputAdapter() {
     return this.inputAdapter;
   }
+
+  public String generateAssetUrl(String filename) {
+    return String.format("%s/v3/assets/%s/%s", this.dashboardHostname, this.channelId, filename);
+  }
+
+  public String generateSourcesUrl() {
+    return String.format("%s/v3/sources/%s", this.dashboardHostname, this.channelId);
```
jmkulwik, line 59: My one worry about generating the URLs on Javabuilder is this:
The way we access S3 is in general an area of tech debt that we hope to eventually clean up down the line. I would guess the URLs might change as part of that work. I don't think that should prevent us from adding this code here, but it might be worth adding a note to https://github.com/code-dot-org/code-dot-org/blob/staging/shared/middleware/files_api.rb explaining the dependency. Then future engineers would know to set up a redirect or to ensure all dependencies have been tracked down before changing the URLs.

org-code-javabuilder/media/src/main/java/org/code/media/Image.java
molly-moen: this is going to be a part of the task to implement the theater image api, this was a placeholder for testing this feature

org-code-javabuilder/protocol/src/main/java/org/code/protocol/GlobalProtocol.java
molly-moen: Sounds good. I'll add a comment in my related code-dot-org PR for now. We can revisit if we want to move the url generation into `javabuilder_sessions_controller` instead in the future. I would guess if we change how we access S3 in the future we might have the update javabuilder anyway since it is also accessing S3 (it might not be as simple as replacing a url if we add extra auth for example).

PR #65, opened 2021-06-30
Implement audio commands from the Theater API

sanchitmalhotra126:
This hooks up the theater audio API to the existing audio code, which enables most of the Theater audio functionality, except for:
- Actually loading audio files from asset URLs (pending implementation of asset URL construction)
- Playing instruments via `playNote()`. The underlying code is written assuming that we will have a list of file paths and/or URLs for the various audio samples, so all the logic for looking up audio samples exists, but the actual map of instrument and note to file/URL is empty (see `InstrumentSampleLoader`).

Tested locally & with unit tests.

### DIFF
--- a/org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
+++ b/org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
@@ -4,6 +4,7 @@ import java.awt.*;
 import java.awt.image.BufferedImage;
 import java.io.ByteArrayOutputStream;
 import java.io.FileNotFoundException;
+import org.code.media.AudioWriter;
 import org.code.media.Color;
 import org.code.media.Image;
 import org.code.protocol.GlobalProtocol;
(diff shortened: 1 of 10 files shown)


org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
```diff
@@ -325,7 +356,11 @@ public void play() {
     } else {
       this.gifWriter.writeToGif(this.image, 0);
       this.gifWriter.close();
-      outputAdapter.sendMessage(ImageEncoder.encodeStreamToMessage(this.outputStream));
+      this.outputAdapter.sendMessage(ImageEncoder.encodeStreamToMessage(this.imageOutputStream));
+
+      this.audioWriter.writeToAudioStreamAndClose();
```
molly-moen, line 361: nit: does `writeToAudioStreamAndClose()` take a noticeable amount of time? I'm wondering if we should wait to send the image until we have the audio ready (this may become irrelevant when we change how the data is stored).

molly-moen (approved):
Looks good!

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
sanchitmalhotra126: Good question - I'll move the `sendMessage` calls together for now at least.

PR #68, opened 2021-07-05
Fix CloudFormation template to actually invoke Lambda runtime wrapper script

sureshc:
The CloudFormation template was correctly deploying (via a Layer) the bash script that changes the Java runtime launch directory, but it wasn't actually configuring the Lambda to execute it. Add the environment variable to launch it. Also simplify the script to just change the `/tmp` directory. This script is only executed during the Init phase, so creating a randomly named subdirectory does not prevent files from one Invocation from persisting to a subsequent Invocation. The Request Handler will need to clear out the contents of the `/tmp` directory on each Invocation.

### DIFF
--- a/org-code-javabuilder/change_runtime_directory
+++ b/org-code-javabuilder/change_runtime_directory
@@ -1,6 +1,7 @@
 #!/bin/bash
-# This script is deployed as a lambda layer to ensure our lambda runs from a temporary folder.
-# This enables File I/O on our lambda.
-rm -rf /tmp/*
-cd $(mktemp -d)
+# This script is deployed as an AWS Lambda Layer and is invoked as a Lambda runtime wrapper script. It changes
+# the Java runtime to launch from the only write-able directory (/tmp) to make it easier for student projects to
+# write files. The Lambda Request Handler is responsible for clearing the contents of this directory at the start/end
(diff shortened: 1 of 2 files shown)


org-code-javabuilder/change_runtime_directory
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
-# This script is deployed as a lambda layer to ensure our lambda runs from a temporary folder.
-# This enables File I/O on our lambda.
-rm -rf /tmp/*
-cd $(mktemp -d)
+# This script is deployed as an AWS Lambda Layer and is invoked as a Lambda runtime wrapper script. It changes
+# the Java runtime to launch from the only write-able directory (/tmp) to make it easier for student projects to
+# write files. The Lambda Request Handler is responsible for clearing the contents of this directory at the start/end
```
molly-moen, line 4: I'm guessing we need to make a task to implement this on the Javabuilder side?

molly-moen (approved):
LGTM!

org-code-javabuilder/change_runtime_directory
sureshc: @jmkulwik added a Task for this https://codedotorg.atlassian.net/browse/CSA-537
Sorry for not noting that earlier.

PR #69, opened 2021-07-06
Theater: add color constructor for Strings

molly-moen:
Implement the constructor for `org.code.media.Color` that allows construction of a color from a String literal. The supported Strings are those Colors we have constants for. The constructor is case-insensitive.


org-code-javabuilder/media/src/main/java/org/code/media/Color.java
```diff
@@ -9,12 +11,18 @@
...
   public Color(String color) throws IllegalArgumentException {
-    // TODO: Implement this method
+    String colorAllCaps = color.toUpperCase();
+    if (!ColorConstantMap.containsKey(colorAllCaps)) {
+      throw new IllegalArgumentException("Invalid color " + color);
+    }
+    Color colorToCopy = ColorConstantMap.get(colorAllCaps);
+    this.red = colorToCopy.getRed();
+    this.green = colorToCopy.getGreen();
+    this.blue = colorToCopy.getBlue();
```
sanchitmalhotra126, line 25: nit: this can just be `this(colorToCopy)` right? or even `this(ColorConstantMap.get(colorAllCaps))`

sanchitmalhotra126 (approved):
One comment but otherwise LGTM!

org-code-javabuilder/media/src/main/java/org/code/media/Color.java
molly-moen: since we want to throw an exception if the color is invalid it doesn't work, because `this(...)` has to be the first line of the constructor.

sanchitmalhotra126: Ahh right yes, forgot about that :)

PR #72, opened 2021-07-12
Add Output Bucket to CloudFormation template

molly-moen:
Additions to the Javabuilder CloudFormation Template to add new components needed for writing theater files to S3. New components are:
- S3 Bucket with public read access. Objects in the bucket expire after 1 day since these are temporary files.
- Output Domain name configured via CloudFront 
- API Certificate for Output Domain Name

This is similar to our [p5-replay template](https://github.com/code-dot-org/p5-replay/blob/main/template.yml)

### DIFF
--- a/template.yml
+++ b/template.yml
@@ -20,6 +20,8 @@ Globals:
     Timeout: 30
     MemorySize: 3008
     Tracing: Active
+Conditions:
+  IsDevCondition: !Equals [!Ref BaseDomainName, "dev-code.org"] 
 Resources:
   Domain:
     Type: AWS::Route53::RecordSet
@@ -267,7 +269,7 @@ Resources:
   OutputBucket:
     Type: AWS::S3::Bucket
     Properties:
-      BucketName: !Sub  "cdo-dev-${SubDomainName}-output"
+      BucketName: !If [IsDevCondition, !Sub "cdo-dev-${SubDomainName}-output", !Sub "cdo-${SubDomainName}-output"] 
       CorsConfiguration:
         CorsRules:
           - AllowedMethods: [GET]
--- a/template.yml
+++ b/template.yml
@@ -262,6 +262,8 @@ Resources:
       Environment:
         Variables:
           AWS_LAMBDA_EXEC_WRAPPER: /opt/change_runtime_directory
+          OUTPUT_BUCKET_NAME: !Ref OutputBucket
+          GET_OUTPUT_URL: !Sub "https://${OutputDomain}"
   OutputBucket:
     Type: AWS::S3::Bucket
     Properties:
@@ -274,7 +276,7 @@ Resources:
       BucketEncryption:
         ServerSideEncryptionConfiguration:
           - ServerSideEncryptionByDefault:
-              SSEAlgorithm: 'aws:kms'
+              SSEAlgorithm: 'AES256'
       LifecycleConfiguration:
         Rules:
           - Id: ExpirationRule
--- a/template.yml
+++ b/template.yml
@@ -262,6 +262,82 @@ Resources:
       Environment:
         Variables:
           AWS_LAMBDA_EXEC_WRAPPER: /opt/change_runtime_directory
+  OutputBucket:
+    Type: AWS::S3::Bucket
+    Properties:
+      BucketName: !Sub  "cdo-dev-${SubDomainName}-output"
+      CorsConfiguration:
+        CorsRules:
+          - AllowedMethods: [GET]
+            AllowedOrigins: ['*']
+            AllowedHeaders: ['*']
+      BucketEncryption:
+        ServerSideEncryptionConfiguration:
+          - ServerSideEncryptionByDefault:
+              SSEAlgorithm: 'aws:kms'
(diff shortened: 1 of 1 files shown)


template.yml
```diff
@@ -262,6 +262,82 @@ Resources:
       Environment:
         Variables:
           AWS_LAMBDA_EXEC_WRAPPER: /opt/change_runtime_directory
+          OUTPUT_BUCKET_NAME: !Ref OutputBucket
```
molly-moen, line 265: not sure if this is the best way to get these parameters over to the lambda, if there's a better pattern let me know!

template.yml
```diff
@@ -262,6 +262,82 @@ Resources:
...
+        Aliases: [!Sub "${SubDomainName}-output.${BaseDomainName}"]
+        ViewerCertificate:
+          AcmCertificateArn: !Ref OutputAPICertificate
+          MinimumProtocolVersion: TLSv1
+          SslSupportMethod: sni-only
+        CustomErrorResponses:
+          - ErrorCode: 403
+            ErrorCachingMinTTL: 0
+        # TODO: enable logging when LogBucket is set up
+        # Logging:
```
molly-moen, line 326: Seems like we don't have a LogBucket on the dev environment so I left this commented out. LMK if this should be updated.

template.yml
```diff
@@ -262,6 +262,82 @@ Resources:
       Environment:
         Variables:
           AWS_LAMBDA_EXEC_WRAPPER: /opt/change_runtime_directory
+          OUTPUT_BUCKET_NAME: !Ref OutputBucket
+          GET_OUTPUT_URL: !Sub "https://${OutputDomain}"
+  OutputBucket:
+    Type: AWS::S3::Bucket
+    Properties:
+      BucketName: !Sub  "cdo-dev-${SubDomainName}-output"
```
sureshc, line 270: Is there a way to conditionally use the prefix `cdo-` when BaseDomainName is `code.org` and `cdo-dev-` when BaseDomainName is `dev-code.org`?

Not sure that's possible with [CloudFormation Condition Functions](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-conditions.html). If not, we may to make the S3 Bucket Name prefix a template Parameter or hard code the prefix to `cdo-` and hope that no-one tries to make an instance of Javabuilder in the Dev account that conflicts with one in the production account.

template.yml
sureshc: I'll try to provision a `cdo-dev-logs` bucket in the AWS Dev account to provide some parity.

template.yml
molly-moen: Good catch, meant to fix this before my PR. It looks like I can do it conditionally

molly-moen:
@sureshc one thing I had to do to get this to work was edit the lambda execution role in S3--does that only exist in S3 or is there somewhere in the codebase it should be documented?

template.yml
```diff
@@ -262,6 +264,82 @@ Resources:
...
+            AllowedHeaders: ['*']
+      BucketEncryption:
+        ServerSideEncryptionConfiguration:
+          - ServerSideEncryptionByDefault:
+              SSEAlgorithm: 'AES256'
+      LifecycleConfiguration:
+        Rules:
+          - Id: ExpirationRule
+            Status: Enabled
+            ExpirationInDays: 1
```
jmkulwik, line 286: Just a question (non-blocking). Is it easy to set this even lower? I could see 6 hours (or even 4) being a reasonable amount of time to retain these.

jmkulwik (approved):
Deferring to Suresh here, but LGTM!

template.yml
molly-moen: The options I saw were either setting a specific timestamp for expiration or setting expiration in days: [doc](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-lifecycleconfig-rule.html), and per [this](https://aws.amazon.com/blogs/aws/amazon-s3-object-expiration/) it seems expiration rules are only evaluated once per day. I agree it could be shorter, if we can find a way to do that I'm all for it!

template.yml
sureshc: I think this is the standard mechanism to pass configuration settings to a Lambda :)

sureshc:
I [updated the IAM stack to grant the Lambda Execution Role permissions](https://github.com/code-dot-org/javabuilder/blob/4a25bf476609da3633b64b0d9895d30970ff1db2/iam.yml#L42-L46) to PUT to any javabuilder output bucket, and I removed the policy you manually added to that role. I haven’t tested that yet. Please let me know if I broke this functionality.

template.yml
sureshc: Hoping to complete setup of a log bucket in the dev account this morning.

molly-moen:
@sureshc still works after the permission update!

sureshc (approved):
Thank you for doing this Infrastructure work for us! :)

I’ll submit a follow up to handle HTTP access logging for the Output endpoint.

PR #73, opened 2021-07-14
Theater: Write output to files

molly-moen:
Theater messages were often too large to send over the websocket, so we need to instead write the gif and audio generated on `play` to our S3 bucket. This PR adds that functionality to theater, and instead of sending back the bytes of the gif and audio it sends urls with the location of the gif and audio. It temporarily keeps in place the message with the raw bytes until we transition the client to use the urls to retrieve data.

As a security measure, we only allow two writes to S3 per client session, as this is all that is needed for theater.

Note: This PR is dependent on merging [#72](https://github.com/code-dot-org/javabuilder/pull/72)


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSFileWriter.java
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  @Override
+  public String writeToFile(String filename, InputStream input) throws JavabuilderException {
+    if (this.writes >= WRITES_PER_SESSION) {
+      throw new UserInitiatedException(UserInitiatedExceptionKey.TOO_MANY_WRITES);
+    }
+    String filePath = this.javabuilderSessionId + "/" + filename;
+    this.s3Client.putObject(this.outputBucketName, filePath, input, null);
+    this.writes++;
```
sanchitmalhotra126, line 37: Do we need to worry about resetting `writes` if writing fails? Or does `s3client` handle all the retry logic for us?

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
```diff
@@ -386,4 +394,28 @@ private void drawImageHelper(
...
+      imageMessage.put("url", imageUrl);
+      this.outputAdapter.sendMessage(new TheaterMessage(TheaterSignalKey.VISUAL_URL, imageMessage));
+
+      HashMap<String, String> audioMessage = new HashMap<>();
+      audioMessage.put("url", audioUrl);
+      this.outputAdapter.sendMessage(new TheaterMessage(TheaterSignalKey.AUDIO_URL, audioMessage));
+    } catch (JavabuilderException e) {
+      // we should not hit this (caused by too many file writes)
+      // in normal execution as it is only called via play,
+      // and play can only be called once.
```
sanchitmalhotra126, line 417: Given this, just a semantic question - should `TOO_MANY_WRITES` just be an internal exception then? Would a user be able to hit this case normally?

org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalMain.java
```diff
@@ -18,7 +18,7 @@ public static void main(String[] args) {
     // Create and invoke the code execution environment
     try {
       UserProjectFiles userProjectFiles = fileLoader.loadFiles();
-      GlobalProtocol.create(outputAdapter, inputAdapter, "", "");
+      GlobalProtocol.create(outputAdapter, inputAdapter, "", "", null);
```
sanchitmalhotra126, line 21: do you think it would be creating a `NoOpFileWriter` or `LocalFileWriter` of some sort just to prevent any edge case null pointer exceptions?

sanchitmalhotra126 (commented):
Couple small things, but looks good to me otherwise!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSFileWriter.java
molly-moen: I'll double check what happens if `putObject` fails, good call out

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
molly-moen: answered in slack

org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalMain.java
molly-moen: good idea, will add that

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSFileWriter.java
molly-moen: I added a put with retry method that tries to write 3 times if there are failures. If that fails we will throw an exception because this is a critical failure (student wouldn't be able to load audio and/or gif)

org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalMain.java
molly-moen: added!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaLogHandler.java
```diff
@@ -0,0 +1,26 @@
...
+
+  public LambdaLogHandler(LambdaLogger logger, String sessionId) {
+    this.logger = logger;
+    this.sessionId = sessionId;
+  }
+
+  @Override
+  public void logInfo(JSONObject eventData) {
+    this.logger.log("INFO: " + eventData.toString());
+  }
```
sanchitmalhotra126, line 19: Do we need the session ID in `logInfo` too?

sanchitmalhotra126 (approved):
LGTM!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaLogHandler.java
molly-moen: per Suresh we will only need it for debugging student issues, so for info we can go without session id

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSFileWriter.java
sureshc: Some retry / timeout patterns we’ve adopted in our main repository are here. Over time we may want to include this type of logic here as well https://github.com/code-dot-org/code-dot-org/blob/staging/lib/cdo/aws/s3.rb

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSFileWriter.java
```diff
@@ -0,0 +1,86 @@
...
+      this.writes++;
+    } else {
+      // We couldn't write to S3, send a message to the user
+      throw new UserFacingException(InternalErrorKey.INTERNAL_RUNTIME_EXCEPTION);
+    }
+
+    return this.getOutputURL + "/" + filePath;
+  }
+
+  private boolean putObjectWithRetry(String filePath, InputStream input, ObjectMetadata metadata) {
```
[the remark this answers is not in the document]
sureshc: I’m not familiar with the AWS Java SDK. Does the Java S3 client have any built-in retry features? I believe v3 of the AWS Ruby SDK client does, and even has a fancy adaptive retry feature.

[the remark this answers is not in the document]
molly-moen: oh you are right. Their api docs didn't say anything about auto-retrying (and said they threw exceptions on error, but this [doc](https://docs.aws.amazon.com/general/latest/gr/api-retries.html) says all SDKs have retry. I'll remove the retry logic and just have it fail if we get an exception

sureshc (approved):
Glad we found the “Write” solution for Theater output files :)

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
```diff
@@ -362,6 +366,9 @@ public void play() {
       this.gifWriter.close();
       this.audioWriter.writeToAudioStreamAndClose();
 
+      this.writeImageAndAudioToFile();
+
+      // TODO: Remove these messages once we've updated the client to load from file.
       this.outputAdapter.sendMessage(ImageEncoder.encodeStreamToMessage(this.imageOutputStream));
       this.outputAdapter.sendMessage(SoundEncoder.encodeStreamToMessage(this.audioOutputStream));
```
sanchitmalhotra126, line 373: Looking at this now: since we're now allowing the client to accept both URLs and base64 encoded files, if we send these back to back, won't we end up overwriting the S3 messages with the encoded file messages?

sanchitmalhotra126 (approved):
LGTM!

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
```diff
@@ -362,8 +366,14 @@ public void play() {
       this.gifWriter.close();
       this.audioWriter.writeToAudioStreamAndClose();
 
-      this.outputAdapter.sendMessage(ImageEncoder.encodeStreamToMessage(this.imageOutputStream));
-      this.outputAdapter.sendMessage(SoundEncoder.encodeStreamToMessage(this.audioOutputStream));
+      // If we are in local mode, write output data directly to output adapter
+      if (GlobalProtocol.getInstance().getDashboardHostname().contains("localhost")) {
```
jmkulwik, line 370: Rather than checking for the string "localhost" here, we should create an interface that is implemented differently for local vs aws instances of javabuilder. That way Stage.java doesn't need to know if the dashboard hostname contains localhost and can invoke the interface directly.

molly-moen: What kind of interface are you thinking of? I thought about doing something on the file writer but we need to encode the image/audio and it doesn't make much sense to move that over to `lib`

PR #74, opened 2021-07-14
Theater: Enable instrument file playback in Javabuilder

sanchitmalhotra126:
Adds the generated instrument files for piano and bass to Javabuilder and hooks them up to the student facing API. As discussed, these files for now live in the `resources` directory.

https://codedotorg.atlassian.net/browse/CSA-543

Example code for playing instruments in Javalab Theater:
```
import org.code.theater.*;

public class MyClass {
      Stage s = Theater.stage;
      
      // Play four notes, 3 seconds each
      s.playNote(Instrument.PIANO, 60, 3.0);
      s.playNote(Instrument.PIANO, 67, 3.0);
      s.playNote(Instrument.BASS, 36, 3.0);
      s.playNote(Instrument.BASS, 31, 3.0);

      // Invalid note values should log an error and play nothing
      s.playNote(Instrument.PIANO, 10, 3.0);
      s.playNote(Instrument.BASS, 70, 3.0);
      
      s.play();
   }
}
```

Tested locally.


jmkulwik:
Would you mind adding to the PR description a sample code snippet that could be used to test this?

org-code-javabuilder/media/src/main/java/org/code/media/AudioUtils.java
```diff
@@ -156,4 +136,67 @@ public static void writeBytesToOutputStream(byte[] bytes, ByteArrayOutputStream
...
+   */
+  public static double[] readSamplesFromAssetFile(String filename) throws FileNotFoundException {
+    try {
+      final URL audioFileUrl = new URL(GlobalProtocol.getInstance().generateAssetUrl(filename));
+      final AudioInputStream audioInputStream =
+          AudioUtils.convertStreamToDefaultAudioFormat(
+              AudioSystem.getAudioInputStream(audioFileUrl));
+      return AudioUtils.readSamplesFromInputStream(audioInputStream);
+    } catch (IOException e) {
+      throw new FileNotFoundException("Could not find file: " + filename);
```
jmkulwik, line 155: Is this a user-facing string? If so, we'll need to be able to translate it on the client side and this should be set up as a child of JavabuilderException. Take a look at the ILLEGAL_METHOD_ACCESS exception key and its use for an example of passing metadata (such as `filename`) to the client

molly-moen: I think in the API we want this to throw a FileNotFoundException that students will see and can try/catch on (we have the same problem on the image side that the message is in English instead of a key). Can we make it translatable and maintain the FileNotFoundException?

sanchitmalhotra126:
> Would you mind adding to the PR description a sample code snippet that could be used to test this from Java Lab?

Sure, added to PR description.

org-code-javabuilder/media/src/main/java/org/code/media/AudioUtils.java
sanchitmalhotra126: Yep, this is supposed to be same behavior as with images. I can turn this into a translatable exception for image and audio as a follow-up if that works?

org-code-javabuilder/media/src/main/java/org/code/media/AudioWriter.java
```diff
@@ -39,13 +39,22 @@ public void writeAudioSamples(double[] samples, double lengthSeconds) {
     this.writeAudioSamples(AudioUtils.truncateSamples(samples, lengthSeconds));
   }
 
-  public void writeAudioFile(String filename) throws SoundException, FileNotFoundException {
-    this.writeAudioSamples(SoundLoader.read(filename));
+  public void writeAudioAssetFile(String filename) throws SoundException, FileNotFoundException {
```
molly-moen, line 42: should this be `writeAudioAssetFromFile`? If I'm understanding correctly it's writing to a stream, not a file, is that right?

sanchitmalhotra126: Good point, yep it's writing to a stream. How's `writeAudioFromAssetFile` and `writeAudioFromLocalFile` (to distinguish the two)?

molly-moen: yep that's good!

org-code-javabuilder/media/src/main/java/org/code/media/AudioUtils.java
jmkulwik: Follow-up sounds good. Mind logging a work item to track it?

molly-moen (approved):
LGTM!

org-code-javabuilder/media/src/main/java/org/code/media/AudioUtils.java
sanchitmalhotra126: Sure thing, done: https://codedotorg.atlassian.net/browse/CSA-553

PR #78, opened 2021-07-22
Theater: Play audio in parallel instead of sequentially

sanchitmalhotra126:
As per curriculum/product this changes audio API behavior to have audio APIs play back audio in parallel rather than sequentially. This mirrors how draw commands work as well. Also adds the `playNoteWithPause` method to the API which allows users to play notes in sequence without having to explicitly call `pause` between each play command. 

Tested locally.

### DIFF
--- a/org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
+++ b/org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
@@ -105,22 +105,21 @@ public class Stage {
    *
    * @param instrument the instrument to play.
    * @param note the note to play. 60 represents middle C on a piano.
-   * @param seconds length of the note. Implementer's note: Behind the scenes, this should just be
-   *     implemented using an array of loopable sounds (Mike can generate these).
+   * @param seconds length of the note.
    */
   public void playNote(Instrument instrument, int note, double seconds) {
     this.playNote(instrument, note, seconds, false);
   }
 
   /**
-   * Plays a note with the selected instrument and adds a pause for the duration of the note.
+   * Plays a note with the selected instrument and adds a pause in drawing/audio for the duration of
+   * the note, so that subsequent play commands begin after the note has finished playing.
    * Convenience method for playing notes in sequence without needing to call pause() between each
    * one.
    *
    * @param instrument the instrument to play.
    * @param note the note to play. 60 represents middle C on a piano.
-   * @param seconds length of the note. Implementer's note: Behind the scenes, this should just be
-   *     implemented using an array of loopable sounds (Mike can generate these).
+   * @param seconds length of the note.
    */
   public void playNoteAndPause(Instrument instrument, int note, double seconds) {
     this.playNote(instrument, note, seconds, true);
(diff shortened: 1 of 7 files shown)


org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
```diff
@@ -107,16 +109,21 @@ public void playSound(String filename) throws FileNotFoundException {
...
-      System.out.printf("Could not play instrument: %s at note: %s%n", instrument, note);
-    }
+  /**
+   * Plays a note with the selected instrument and adds a pause for the duration of the note.
+   * Convenience method for playing notes in sequence without needing to call pause() between each
+   * one.
+   *
+   * @param instrument the instrument to play.
+   * @param note the note to play. 60 represents middle C on a piano.
+   * @param seconds length of the note. Implementer's note: Behind the scenes, this should just be
```
molly-moen, line 122: nit: time to remove "implementer's note" here?

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
```diff
@@ -420,4 +427,20 @@ private void writeImageAndAudioToFile() {
...
+      return;
+    }
+
+    try {
+      this.audioWriter.writeAudioFromLocalFile(sampleFilePath, noteLength);
+      if (shouldPause) {
+        this.pause(noteLength);
+      }
+    } catch (FileNotFoundException e) {
+      System.out.printf("Could not play instrument: %s at note: %s%n", instrument, note);
```
molly-moen, line 443: should this throw an exception instead?

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
```diff
@@ -107,16 +109,21 @@ public void playSound(String filename) throws FileNotFoundException {
...
 
-    try {
-      this.audioWriter.writeAudioFromLocalFile(sampleFilePath, seconds);
-    } catch (FileNotFoundException e) {
-      System.out.printf("Could not play instrument: %s at note: %s%n", instrument, note);
-    }
+  /**
+   * Plays a note with the selected instrument and adds a pause for the duration of the note.
+   * Convenience method for playing notes in sequence without needing to call pause() between each
+   * one.
```
molly-moen, line 118: Does the pause cause a pause of the note's length after the note plays, or 'pause' the length of the note starting at when the note starts playing, so there is no audible pause between notes if you always called `playNoteAndPause`?

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
sanchitmalhotra126: So per curriculum, if `playNote` is called with in invalid note/instrument, we just want return and log an error instead of throwing

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
sanchitmalhotra126: The second case, yeah - if you called `playNoteAndPause` with a time of 4 seconds, you'd hear the note play for 4 seconds and the next play command would play audio right after the note finished playing with no audible pause between. In contrast, multiple calls to `playNote` (and `playSound`) will play sounds at the same time unless pause is called in between.

molly-moen: Makes sense. Can you clarify that in the comment?

molly-moen (approved):

PR #80, opened 2021-07-23
Added support for decoding JWT tokens on the local websocket server.

jmkulwik:
This PR should be merged concurrently with a dependent PR in the code-dot-org repo: https://github.com/code-dot-org/code-dot-org/pull/41720

Prior to this change, the Local WebSocket deployment of Javabuilder used query string parameters to determine code execution specifications whereas the production deployment of Javabuilder would use the decoded JWT token generated by the dashboard server and passed from the client. This led to diverging code paths between the two deployment types. Now, the parameters for both local and production deployments come from the JWT token. 

This does not add support for verifying the token's authenticity. However, since this is a locally hosted environment, used for development purposes only, this is not needed.

This is done as a prefactor for [CSA-518](https://codedotorg.atlassian.net/browse/CSA-518) and [CSA-519](https://codedotorg.atlassian.net/browse/CSA-519) which will add a compile-only request type that will be used by the backpack and commit code features. This prefactor will allow us to add more fields to the options parameter (such as the list of files to compile) without diverging local and production Javabuilder further.

Instructions for decoding the JWT token came from here: https://www.baeldung.com/java-jwt-token-decode

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketServer.java
+++ b/org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketServer.java
@@ -2,8 +2,7 @@ package dev.javabuilder;
 
 import java.io.IOException;
 import java.nio.ByteBuffer;
-import java.util.List;
-import java.util.Map;
+import java.util.Base64;
 import javax.websocket.OnClose;
 import javax.websocket.OnMessage;
 import javax.websocket.OnOpen;
@@ -15,6 +14,7 @@ import org.code.protocol.GlobalProtocol;
 import org.code.protocol.JavabuilderException;
 import org.code.protocol.JavabuilderRuntimeException;
 import org.code.protocol.Properties;
+import org.json.JSONObject;
 
 /**
  * This sets up a simple WebSocket server for local development when interactions between dashboard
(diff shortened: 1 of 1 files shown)


org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketServer.java
```diff
@@ -38,21 +38,26 @@
...
-      useNeighborhood = Boolean.parseBoolean(params.get("useNeighborhood").get(0));
-    }
-    String levelId = "";
-    if (params.containsKey("levelId")) {
-      levelId = params.get("levelId").get(0);
+    if (options.has("useNeighborhood")) {
+      String useNeighborhoodStr = options.getString("useNeighborhood");
+      useNeighborhood = Boolean.parseBoolean(useNeighborhoodStr);
     }
+
```
jmkulwik, line 56: This section was updated to match the LambdaRequestHandler, the entry point for production Javabuilder
https://github.com/code-dot-org/javabuilder/blob/5cebd2f38f51437f177cd2e76a875165d8a4ddc5/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java#L43-L54

sanchitmalhotra126: Is it possible to share this code with `LambdaRequestHandler` as it has a similar function? Seems fine if not, since it's not too much, but just asking to help prevent any potential divergences here between local/prod.

molly-moen (approved):
Nice! To clarify--when running locally we won't need the public key, it'll just decode without the public key?

jmkulwik:
> Nice! To clarify--when running locally we won't need the public key, it'll just decode without the public key?

Yes. Details here: https://codedotorg.slack.com/archives/C01EF4GJ9GE/p1627322428173400

org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketServer.java
jmkulwik: Yes, I think the eventual goal would be to pass the options object into the various classes/methods that need it rather than querying its fields here. This would provide two benefits: it would further align the WebSocket and Lambda deployments and it would remove the need for testing on an AWS deployed javabuilder each time a new field is added to options.

We generally [prefer small PRs](https://docs.google.com/document/d/1-x-GcC2xWpUymU6YBWHzxmTAlASWQPMfAf8ASBqWlgA/edit). This seemed like a good atomic improvement that can be built upon in the future and a good pausing point in this work.

sanchitmalhotra126: Got it, makes sense!

sanchitmalhotra126 (approved):

PR #81, opened 2021-07-27
Code doesn't compile if there is an empty file

sanchitmalhotra126:
Handles the case where a student tries to run a project with an empty Java file (blank or only comments). I opted to keep this as a runtime error since technically the empty files do compile, but throw a ClassNotFoundException when trying to run.

Related code-dot-org PR for interpreting this error on the client: https://github.com/code-dot-org/code-dot-org/pull/41767

Repro steps:
- Navigate to a Javalab level
- Open and save a new file with the extension .java
- Run the program

Tested locally.


molly-moen (approved):
🎉

PR #84, opened 2021-07-29
Add compile/run status messages

molly-moen:
Add javabuilder status messages we can use to communicate with users about where in the compile/run process their code is. This PR adds the following messages:
- Compiling
- Compile complete
- Running
- Generating results (for Theater only, this is sent before we start the upload to S3 for gif and audio)
- Exited (either on an exception or a successful run)

This PR also includes a  small fix to the websocket server to get the options as a string, to fix neighborhood on local environments.


org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketServer.java
```diff
@@ -47,7 +47,7 @@ public void onOpen(Session session) {
     final String levelId = queryInput.getString("level_id");
     final String channelId = queryInput.getString("channel_id");
     final String dashboardHostname = "http://" + queryInput.get("iss") + ":3000";
-    final JSONObject options = new JSONObject(queryInput.get("options"));
+    final JSONObject options = new JSONObject(queryInput.getString("options"));
```
sanchitmalhotra126, line 50: 🥳

org-code-javabuilder/protocol/src/main/java/org/code/protocol/StatusMessageKey.java
```diff
@@ -0,0 +1,9 @@
+package org.code.protocol;
+
+public enum StatusMessageKey {
+  COMPILING,
+  COMPILATION_SUCCESSFUL,
+  RUNNING,
+  GENERATING_RESULTS,
+  EXITED
```
sanchitmalhotra126, line 8: nit: would Done/Finished be clearer? Or do we specifically want Exited here?

sanchitmalhotra126 (approved):

org-code-javabuilder/protocol/src/main/java/org/code/protocol/StatusMessageKey.java
molly-moen: I don't have a super strong opinion here, on the frontend we are displaying "completed" generally (what we do with this signal depends on the miniapp ). I can update this to "Completed" to make things match up.

sanchitmalhotra126: Ahh okay if it's different per miniapp then it doesn't really matter I suppose, I'm good with Exited too

PR #86, opened 2021-08-03
Theater: wait to create pixel array

molly-moen:
If you create an `Image` from a file and never edit it, we get a buffered image from the file, create a 2d array of pixels based on that image, then convert them back to a buffered image to write it to the stage. Since adding an image without editing it will be a fairly common workflow, optimize this flow to only create the 2d array of pixels the first time it's needed--either on `getPixel`, `setPixel` or `clear`. If we never create the 2d array, then we can just return back the original buffered image without the extra steps and memory.

This will fix an out of memory issue seen if a user has a large image file (seen at 790kb, not sure on the lower limit here) and never edits it. We would still get an out of memory issue if the user wanted to edit the image. It also speeds up our processing time if an image is never edited.

### DIFF
--- a/org-code-javabuilder/media/src/main/java/org/code/media/Image.java
+++ b/org-code-javabuilder/media/src/main/java/org/code/media/Image.java
@@ -12,6 +12,7 @@ public class Image {
   private int width;
   private int height;
   private static final Color DEFAULT_BACKGROUND_COLOR = Color.WHITE;
+  private BufferedImage bufferedImage;
 
   /**
    * Creates a new image object, using the pixel information from the file uploaded to the asset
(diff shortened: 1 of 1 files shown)


org-code-javabuilder/media/src/main/java/org/code/media/Image.java
```diff
@@ -76,6 +71,10 @@ public Image(int width, int height) {
    * @return Pixel at the given coordinate
    */
   public Pixel getPixel(int x, int y) {
+    if (this.pixels == null) {
+      this.createPixelArray();
+      this.bufferedImage = null;
```
molly-moen, line 76: we don't explicitly need to set bufferedImage to null, but since we are running on a fairly constrained set of resources I'm setting it to null to indicate that it can be garbage collected.

sanchitmalhotra126 (approved):

PR #88, opened 2021-08-11
Import assets from the level into Javabuilder

sanchitmalhotra126:
Previously, Javabuilder wasn't checking the level starter assets when trying to load assets. This change makes Javabuilder first check the list of starter asset file names and use the corresponding URL if it exists, or if not, use the standard assets URL for loading assets.

Testing: local & unit

Related TODO: https://codedotorg.atlassian.net/browse/CSA-690

### DIFF
--- a/org-code-javabuilder/protocol/src/main/java/org/code/protocol/AssetUrlGenerator.java
+++ b/org-code-javabuilder/protocol/src/main/java/org/code/protocol/AssetUrlGenerator.java
@@ -1,0 +1,100 @@
+package org.code.protocol;
+
+import java.io.IOException;
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.time.Duration;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+public class AssetUrlGenerator {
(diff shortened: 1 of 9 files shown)


org-code-javabuilder/protocol/src/main/java/org/code/protocol/AssetUrlGenerator.java
```diff
@@ -0,0 +1,100 @@
...
+    return String.format(
+        "%s/level_starter_assets/%s/%s?t=%d",
+        this.dashboardHostname, this.levelId, filename, System.currentTimeMillis());
+  }
+
+  private String generateStarterAssetsListUrl() {
+    return String.format("%s/level_starter_assets/%s", this.dashboardHostname, this.levelId);
+  }
+
+  // TODO: Create generic HTTP request handler and share with UserProjectFileLoader
```
jmkulwik, line 84: NIT: When you add a TODO, create a task for it and add the PR's link to the task. It's also often good practice to make a github thread on the TODO line with the task's link.

sanchitmalhotra126: Sure, I've created one [here](https://codedotorg.atlassian.net/browse/CSA-690). Will add to the description.

jmkulwik (approved):
Nice work! Nice job deduping the logic too

jmkulwik:
Sorry for the late comment - this would be a great candidate for an eyes/UI test

PR #92, opened 2021-08-23
Lambda forces disconnect on code completion

bencodeorg:
This change makes our Javabuilder lambdas disconnect from the client on code completion.

## Testing story

I was only able to test this change via Postman, with my deployed development version of Javabuilder. Here's what running the default code on /s/allthethings/lessons/44/levels/8 (prints hello world) produced without this change:

image

Here's with the change:

image


sanchitmalhotra126 (approved):

jmkulwik (approved):
LGTM!

A couple things I checked: 
- does the SQS get deleted? (i.e. does the delete action get called via API Gateway) - yes
- does the client gracefully handle cases where it tries to close the connection, but the server's already closed it? - yes
- does the server gracefully handle cases where it tries to close the connection, but the client's already closed it? - I think so - deleteConnection seems to return a result object rather than throwing an exception

Happy to chat through any of these if you'd like!

bencodeorg:
> LGTM!
> 
> A couple things I checked:
> 
> * does the SQS get deleted? (i.e. does the delete action get called via API Gateway) - yes
> * does the client gracefully handle cases where it tries to close the connection, but the server's already closed it? - yes
> * does the server gracefully handle cases where it tries to close the connection, but the client's already closed it? - I think so - deleteConnection seems to return a result object rather than throwing an exception
> 
> Happy to chat through any of these if you'd like!

Would love to talk through how you tested these things when you have a minute!

PR #95, opened 2021-08-27
Add Java Logger to Javabuilder error handling

jmkulwik:
_~Note: This PR is based on a different branch than normal due to a~ [prefactor PR](https://github.com/code-dot-org/javabuilder/pull/96)._

Here, we make use of the Java [Logger](https://docs.oracle.com/javase/7/docs/api/java/util/logging/Logger.html) to handle logging. We use a separate [Handler](https://docs.oracle.com/javase/7/docs/api/java/util/logging/Handler.html) for AWS and Local development. With the logger, we generate JSON logs that include the following:
* Connection ID (the unique ID generated by API Gateway that identifies an individual instance of Javabuilder from when a student clicks the "Run" button until the WebSocket disconnects.
* Level ID
* Project ID

On AWS CloudWatch this ends up looking like this:
image

Prior to this change, instances of errors would appear across multiple error logs. Like this:
image

This PR also sets up logging for every exception that was not caused by the user. These exceptions are the following:
* InternalServerError
* InternalServerRuntime
* InternalFacingException
* Throwables that were not set up by us

As part of this, we now communicate Throwables to the user. Right now, we don't know if they affect the user, so we show them some message rather than the previous behavior - display nothing after the user clicks the run button.
Prior to this change, the user would click the run button and then nothing would happen - depending on where the exception occurred, it could produce no output and then change back from the stop button to the run button eventually.
From WSCAT:
image
Now, users get the "Internal Error" message which is the same behavior as our other "server error" messages.
image

Finally, I added CodeBuilderWrapper to to make all the error handling consistent between local development and production. I'm not sure that this was the right approach and would especially appreciate feedback in this area.

~Validation note: This PR is not yet ready to merge. It still needs manual validation against the AWS deploy of Javabuilder.~


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeBuilderWrapper.java
```diff
@@ -0,0 +1,36 @@
+package org.code.javabuilder;
+
+import org.code.protocol.*;
+
+public class CodeBuilderWrapper {
```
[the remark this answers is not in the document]
jamescodeorg: A header comment here indicating what the wrapper does might be useful.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
```diff
@@ -73,48 +79,42 @@ public String handleRequest(Map<String, String> lambdaInput, Context context) {
...
+      cleanUpResources(connectionId, api);
+      return "error clearing tmpdir";
     }
 
+    // Load files to memory and create and invoke the code execution environment
+    CodeBuilderWrapper codeBuilderWrapper =
+        new CodeBuilderWrapper(userProjectFileLoader, outputAdapter);
+    codeBuilderWrapper.executeCodeBuilder();
+
+    cleanUpResources(connectionId, api);
```
jamescodeorg, line 109: Was it intentional to move this outside of the finally block (so this doesn't execute if executeCodeBuilder() throws an exception)?

org-code-javabuilder/protocol/src/main/java/org/code/protocol/JavabuilderRuntimeException.java
```diff
@@ -19,6 +23,20 @@ protected JavabuilderRuntimeException(Enum key, Throwable cause) {
     this.key = key;
   }
 
+  protected JavabuilderRuntimeException(Enum key, Boolean logOnCreate) {
+    this(key);
+    if (logOnCreate) {
```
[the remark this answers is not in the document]
jamescodeorg: Thanks for the context.  One advantage of logging higher up is that you can log exceptions that are created by the runtime or libraries but perhaps that as much of a concern here?

Having logging code in the exception itself feels weird to me.  Typically I think exceptions are logged where they are handled or up at the top-level.  But I don't feel that strongly about it if the logging isn't driving user behavior.

org-code-javabuilder/protocol/src/main/java/org/code/protocol/LoggerNames.java
```diff
@@ -0,0 +1,5 @@
+package org.code.protocol;
+
+public class LoggerNames {
+  public static final String MAIN_LOGGER = "mainLogger";
```
jamescodeorg, line 4: I don't know how sophisticated we want to make the logging, but IIRC, in other systems I've worked on, we used the fully qualified class name as the logger name to create a hierarchy of loggers.  This allowed for very granular control of the logging verbosity for different parts of the system.

jamescodeorg (commented):
Have we considered using Log4j2?  I believe many Java projects use Log4j2 for logging.  There are some features that you might find useful like the Thread Context Map and Thread Context Stack that would help with recording metadata like levelId, channelId, etc.  I think it's possible that you can avoid having separate LogHandler classes for local and lambda entirely by just configuring Log4j2 appropriately in each environment.

(I have experience with Log4j which predates Log4j2 so some of my information might be out-of-date.  If it helps, I can do more research next week.)

org-code-javabuilder/protocol/src/main/java/org/code/protocol/JavabuilderRuntimeException.java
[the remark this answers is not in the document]
jamescodeorg: I did a little bit of searching and maybe anti-pattern is a bit too strong but there does seem to be some agreement that it's not ideal.  Here's a thread I found on stackexchange: https://softwareengineering.stackexchange.com/questions/306032/should-i-log-errors-on-exception-throwing-constructors that has some viewpoints from others.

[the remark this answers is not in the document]
jmkulwik: Ok. After a bit more thought and looking through the thread you shared, I'm going to go ahead and move the logging to CodeBuilderWrapper. I couldn't find a strong reason for doing it one way or another, but one thing I've definitely learned through the years is that trusting people's instincts tends to pay off.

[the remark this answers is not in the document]
jmkulwik: A point in favor of using CodeBuilderWrapper for this: Thinking about it like Rails, we have one generic 500 page in our dashboard server that gets shown, regardless of the error type, and I'm guessing that's configured globally. Similarly, we'd want a globally configured way of handling these types of errors. This is a much more lightweight system (and this is the logger, rather than the 500 page), but that still seems like a design pattern that is helpful.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeBuilderWrapper.java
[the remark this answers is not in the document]
jmkulwik: 👍

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
```diff
@@ -54,8 +56,12 @@ public String handleRequest(Map<String, String> lambdaInput, Context context) {
...
     }
 
-    final JavabuilderLogger logger =
-        new LambdaLogHandler(context.getLogger(), javabuilderSessionId);
+    Logger logger = Logger.getLogger(MAIN_LOGGER);
+    logger.addHandler(
+        new LambdaLogHandler(
+            context.getLogger(), javabuilderSessionId, connectionId, levelId, channelId));
+    // turn off the default console logger
+    logger.setUseParentHandlers(false);
```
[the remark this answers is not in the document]
jmkulwik: I think they're _just_ different enough that it doesn't make sense.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
jmkulwik: It was, but I'll add it back in. I didn't realize the output adapters had a variety of ways they could throw exceptions.

org-code-javabuilder/protocol/src/main/java/org/code/protocol/LoggerNames.java
jmkulwik: In the future, we may want something like that, but for now, I think we can keep it simple

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeBuilderWrapper.java
```diff
@@ -0,0 +1,66 @@
...
+      JSONObject eventData = new JSONObject();
+      eventData.put("exceptionMessage", e.getExceptionMessage());
+      eventData.put("loggingString", e.getLoggingString());
+      Logger.getLogger(MAIN_LOGGER).severe(eventData.toString());
+
+      // The error affected the user. Tell them about it.
+      outputAdapter.sendMessage(e.getExceptionMessage());
+    } catch (JavabuilderException | JavabuilderRuntimeException e) {
+      // The error affected the user and was caused by them. Tell them about it.
+      outputAdapter.sendMessage(e.getExceptionMessage());
```
sanchitmalhotra126, line 43: Not a huge deal, but would it be conceptually clearer to just have the catch block for `JavabuilderException | JavabuilderRuntimeException` but then only log if the exception is an instance of InternalServer(Runtime)Error? Might be easier to follow the hierarchy since InternalServer(Runtime)Errors are Javabuilder(Runtime)Exceptions

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
```diff
@@ -73,48 +79,54 @@ public String handleRequest(Map<String, String> lambdaInput, Context context) {
...
-        // Delete any leftover contents of the tmp folder from previous lambda invocations
-        Util.recursivelyClearDirectory(Paths.get(System.getProperty("java.io.tmpdir")));
-      } catch (IOException e) {
-        throw new InternalServerError(INTERNAL_EXCEPTION, e);
-      }
+      // Log the error
+      JSONObject eventData = new JSONObject();
+      eventData.put("exceptionMessage", error.getExceptionMessage());
+      eventData.put("loggingString", error.getLoggingString());
+      Logger.getLogger(MAIN_LOGGER).severe(eventData.toString());
```
sanchitmalhotra126, line 105: nit: seeing this in a couple of places, might be worth breaking it out into a method at some point (maybe adding a `String generateEventData()` to `JavabuilderThrowableProtocol`)? don't think it's required for this change though.

sanchitmalhotra126 (approved):
Looks good! Two small things but nothing blocking

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeBuilderWrapper.java
jmkulwik: That's a good idea. I'm going to merge as-is and then I may do a follow-up PR.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
jmkulwik: That's a good idea. I'm going to merge as-is and then I may do a follow-up PR.

PR #96, opened 2021-08-27
Rename internal server errors to more clearly describe what they are doing

jmkulwik:
This renames two exceptions to more accurately describe their function.

UserFacingException -> InternalServerError
InternalJavabuilderError -> InternalServerRuntimeError

These are both errors that are the functional equivalent of 500 server errors. The InternalServerRuntimeError is used when we cause an error from within student APIs. This is a RuntimeException so students won't get a compiler error from failing to handle the exception.

To consider: Should InternalServerError be in the `protocol` package even though the `lib` package is the only one that uses it?

Note: Both of these "Errors" actually implement `Exception` and `RuntimeException` rather than `Error` despite their name. That is because despite being conceptually Internal Server Errors, they act as Exceptions within the Java code. It seemed more important to think of these as 500 errors rather than Java Exceptions when choosing how to write code. Thus the name.

In addition to the above change, there were also 2 places that used the InternalServerRuntimeError that should have used the InternalServerError:
* LambdaRequestHandler, after recursively clearing the tmp directory
* Util's getAllJarPaths, if there is a URI exception

These have both been corrected


sanchitmalhotra126 (approved):
Nice! This is definitely clearer. About the naming - agreed that "Error" makes sense here. I was thinking a little about how all our named exceptions are organized now, and with this change we seem to have a clear separation where internally initiated exceptions are named "Errors" vs user-initiated exceptions are named "Exception" (UserInitiatedException, Media/Neighborhood/Theater/etc RuntimeException) which I think makes sense for me. 

If anything perhaps the naming of Javabuilder(Runtime)Exception could potentially use some rework? Only because I see a clear separation between user-facing errors that are in some way shown to the end user (Javabuilder & JavabuilderRuntimeException) and internal-facing errors (InternalFacingException) that are only reported to CloudWatch. "InternalFacingException" makes its purpose pretty self-explanatory while "JavabuilderException" isn't as clear? That being said it is documented, but since we are reconsidering exception names, it could use some discussion.

jmkulwik:
> If anything perhaps the naming of Javabuilder(Runtime)Exception could potentially use some rework? Only because I see a clear separation between user-facing errors that are in some way shown to the end user (Javabuilder & JavabuilderRuntimeException) and internal-facing errors (InternalFacingException) that are only reported to CloudWatch. "InternalFacingException" makes its purpose pretty self-explanatory while "JavabuilderException" isn't as clear? That being said it is documented, but since we are reconsidering exception names, it could use some discussion.

^ I like this suggestion. I've [started a conversation](https://codedotorg.slack.com/archives/C01EF4GJ9GE/p1630353495012100) on the csa-dev channel to get some suggestions. In the meantime, I'll merge this work so we can get the first set of renames out.

PR #97, opened 2021-08-30
Handle arbitrary message types in Javabuilder

sanchitmalhotra126:
Creates a way for routing messages from Javalab to any arbitrary handler based on a message type key. This change also includes a backwards-compatible fallback to treat any non-JSON messages as standard System.in messages, so current behavior doesn't break before Javalab is fully updated to send JSON messages and deployed to production. As a follow-up, the fallback code should be removed once the corresponding Javalab changes have been deployed to production.

Tested locally + unit. Tested against up-to-date staging Javalab to ensure no input behavior was broken.

https://codedotorg.atlassian.net/browse/CSA-730

### DIFF
--- a/org-code-javabuilder/protocol/src/main/java/org/code/protocol/InputHandler.java
+++ b/org-code-javabuilder/protocol/src/main/java/org/code/protocol/InputHandler.java
@@ -1,0 +1,62 @@
+package org.code.protocol;
+
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.Map;
+import java.util.Queue;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+/**
+ * [PROTOTYPE CODE]
+ *
+ * <p>Handles retrieving various types of JSON messages from the client. Currently expects JSON in
+ * the format:
+ *
+ * <p>{ "messageType": "<message type>", "message": "<message contents>" }
+ */
+public class InputHandler {
+  private final Map<InputMessageType, Queue<String>> inputQueues;
+  private final InputAdapter inputAdapter;
+
+  public InputHandler(InputAdapter inputAdapter) {
+    this.inputAdapter = inputAdapter;
+    this.inputQueues = new HashMap<>();
+  }
+
+  public String getNextMessageForType(InputMessageType type) {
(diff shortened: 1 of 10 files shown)


org-code-javabuilder/protocol/src/main/java/org/code/protocol/InputHandler.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Queue;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+/**
+ * Handles retrieving various types of JSON messages from the client. Expects JSON in the format:
+ *
+ * <p>{ "messageType": "<message type>", "message": "<message contents>" }
+ *
+ * <p>Currently if JSON is not in this format, then it is assumed that the message is a System.in
```
molly-moen, line 15: nice temporary measure!

org-code-javabuilder/protocol/src/main/java/org/code/protocol/InputHandler.java
```diff
@@ -0,0 +1,61 @@
...
+          nextMessageType = InputMessageType.valueOf(jsonMessage.getString(MESSAGE_TYPE_KEY));
+          message = jsonMessage.getString(MESSAGE_KEY);
+        } catch (JSONException | IllegalArgumentException e) {
+          // If the message is not JSON, then assume it is a System.in message.
+          // TODO: Remove this and instead throw exception once Javalab is updated to send JSON
+          nextMessageType = InputMessageType.SYSTEM_IN;
+          message = nextMessageData;
+        }
+
+        System.out.printf("Received message: %s of type: %s\n", message, nextMessageType);
```
molly-moen, line 49: remove this?

molly-moen (commented):
Looks good! I think you just have a leftover system.out :)

org-code-javabuilder/protocol/src/main/java/org/code/protocol/InputHandler.java
sanchitmalhotra126: Oh whoops! Good catch :)

jmkulwik (approved):
Nice work on this!

PR #99, opened 2021-08-31
Neighborhood: Increase "large grid" size and no infinite paint when specified in constructor

bencodeorg:
A couple of updates to the Painter constructor, per request from our curriculum team:
- Updates the default painter constructor to use infinite paint only if the grid is 20x20 or larger.
- Updates the parameterized painter constructor to always pay attention to the amount of paint students specify (this is currently ignored if the grid is larger than 16x16)

## Testing story

Tested manually, and added unit tests.

### DIFF
--- a/org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/Painter.java
+++ b/org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/Painter.java
@@ -4,12 +4,13 @@ import java.util.HashMap;
 import org.code.protocol.OutputAdapter;
 
 public class Painter {
-  private static final int LARGE_GRID_SIZE = 16;
+  private static final int LARGE_GRID_SIZE = 20;
   private static int lastId = 0;
   private int xLocation;
   private int yLocation;
   private Direction direction;
   private int remainingPaint;
+  private boolean hasInfinitePaint;
   private final Grid grid;
   private final String id;
   private final OutputAdapter outputAdapter;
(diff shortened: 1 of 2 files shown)


org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/Painter.java
```diff
@@ -4,12 +4,13 @@
...
 
 public class Painter {
-  private static final int LARGE_GRID_SIZE = 16;
+  private static final int LARGE_GRID_SIZE = 20;
   private static int lastId = 0;
   private int xLocation;
   private int yLocation;
   private Direction direction;
   private int remainingPaint;
+  private boolean hasInfinitePaint;
```
sanchitmalhotra126, line 13: nit: can be `final`

org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/Painter.java
```diff
@@ -22,6 +23,7 @@ public Painter() {
     this.remainingPaint = 0;
     World worldInstance = World.getInstance();
     this.grid = worldInstance.getGrid();
+    this.hasInfinitePaint = (this.grid.getSize() >= LARGE_GRID_SIZE) ? true : false;
```
sanchitmalhotra126, line 26: nit: can simplify to `this.hasInfinitePaint = this.grid.getSize() >= LARGE_GRID_SIZE;`

sanchitmalhotra126 (approved):
Looks good! Two questions which might be follow-ups and need @moneppo or Jamila's input - what should we return for `getMyPaint()` when we have infinite paint? And similarly should `takePaint()` work in the same way, or message something else to users?

PR #100, opened 2021-09-01
Upgrade lambdas to Ruby 2.7

molly-moen:
AWS Lambda no longer supports Ruby 2.5. Upgrade our routing and authorization lambdas to use ruby 2.7. I added instructions to the readmes of those lambdas as to how to install ruby 2.7 and run alongside ruby 2.5. Since the code-dot-org repo recommends setting ruby 2.5 as the global version, a user of Javabuilder can set ruby 2.7 as the local version in their javabuilder folder and both code-dot-org and javabuilder will work.


api-gateway-routes/README.md
```diff
@@ -3,3 +3,9 @@
 This lambda function defines the actions that happen when the Javabuilder API Gateway triggers the connect, disconnect, and default actions.
 
 To create a zip file to upload to the lambda, run [zip_for_aws](zip_for_aws.sh).
+
+# Prerequisites
+You need to install Ruby 2.7.4 to build and/or run this function. [rbenv](https://github.com/rbenv/rbenv) is a useful tool for installing Ruby versions
```
molly-moen, line 8: @jmkulwik mind checking if these instructions are ok for a windows setup?

sureshc:
Awesome! I'll try it out locally this morning.

api-gateway-routes/README.md
jmkulwik: It may be correct, but `zip_for-aws.sh` doesn't work on a windows setup - I just use my EC2 instance right now. For both of these, someone would likely need WSL.

sureshc: Should we remove the Read Me files for each Lambda and use a consolidated Read Me at the root of the project? The deploy shell script at the root of the project builds and deploys the 2 Ruby Lambdas (and the Java Lambda). We probably don’t need the `.zip_for_aws.sh` scripts inside each Lambda directory anymore either.

.gitignore
```diff
@@ -13,3 +13,6 @@
 
 # Ignore Gradle build output directory
 build
+
+# Ignore ruby configuration file
+.ruby-version
```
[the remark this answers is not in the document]
molly-moen: good call, I'll add the .ruby-version file

api-gateway-routes/README.md
molly-moen: I've moved this text to the main readme and deleted the `.zip_for_aws.sh` scripts.

jmkulwik (approved):
Nice work! LGTM!

sureshc:
Can you merge the latest from `main` into this feature branch to pick up #98 ? I'm not able to deploy successfully to a Javabuilder instance in the production AWS account with this feature branch yet.

sureshc:
>  merged latest main

Thanks! I successfully deployed an update to a Ruby Lambda. Sorry for the inconvenience. I should have just deployed to the Development AWS account, which would have worked without the fix in `main`.

sureshc (approved):
Thank you for taking this on! It was considerably more than the 1 line change I naively estimated.

PR #101, opened 2021-09-01
Remove input handler fallback for non-JSON messages

sanchitmalhotra126:
Follow-up for https://codedotorg.atlassian.net/browse/CSA-730 / https://github.com/code-dot-org/javabuilder/pull/97. This removes fallback logic and instead throws an exception if we get an invalid or non-JSON message from Javalab. NOTE: This change should not be merged or deployed until https://github.com/code-dot-org/code-dot-org/pull/42254 has been deployed to production.

Tested locally + unit.


molly-moen (approved):
🎉

sanchitmalhotra126:
Javalab change has been merged and deployed to production - safe to merge and deploy this now.

PR #103, opened 2021-09-03
Theater: add text support

molly-moen:
Add support for `drawText` in the theater. This requires adding a font configuration file as a lambda layer since lambdas do not come with font support by default. It also requires font files so we can support sans serif, serif and mono fonts in normal, bold, italic and bold-italic styles. The font files are packaged with the code so they work on both a local and a lambda deployment.

## Testing
Tested both locally and on my dev lambda. 

### Code
```
stage.drawText("Hello World", 50, 50, Color.BLUE, Font.SANS, FontStyle.ITALIC, 15, 0);
stage.drawText("Hello World Again", 100, 150, Color.GREEN, Font.SERIF, FontStyle.BOLD_ITALIC, 24, 45);
stage.drawText("Hello World!", 200, 150, Color.GREEN, Font.MONO, FontStyle.NORMAL, 10, 180);
```
### Screenshot
Screen Shot 2021-09-03 at 10 44 16 AM


org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
```diff
@@ -189,13 +192,50 @@ public void drawImage(Image image, int x, int y, int width, int height, double r
...
    * @param y the distance from the top of the image to draw the text.
-   * @param color the color to draw the text, using any CSS color string (e.g. #234 or green)
-   * @param font the name of the font to draw the text in
+   * @param color the color to draw the text.
+   * @param font the name of the font to draw the text in.
+   * @param fontStyle the name of the font style to draw the text in.
    * @param height the height of the text in pixels.
    * @param rotation the rotation or tilt of the text, in degrees
    */
   public void drawText(
```
molly-moen, line 201: @moneppo please take a look at api updates

template.yml
```diff
@@ -251,10 +251,19 @@ Resources:
       ContentUri: org-code-javabuilder/change_runtime_directory
       Description: Change Java runtime to launch from the writeable /tmp directory to enable student projects to write files more easily.
       LayerName: change-java-runtime-directory
+  AddFontConfigLayer:
```
molly-moen, line 254: @sureshc tagged you to look at template changes

org-code-javabuilder/media/src/main/java/org/code/media/FontHelper.java
```diff
@@ -0,0 +1,54 @@
...
+              fontToFilenamePrefix.get(font) + fontStyleToFilenameSuffix.get(fontStyle);
+          InputStream fileStream = this.getClass().getClassLoader().getResourceAsStream(filename);
+          this.fontMap.put(
+              filename, java.awt.Font.createFont(java.awt.Font.TRUETYPE_FONT, fileStream));
+        }
+      }
+    } catch (FontFormatException | IOException e) {
+      // throw an exception if we can't load a font, as we will hit a null pointer exception
+      // later if the user tries to use that font.
+      throw new InternalServerRuntimeError(InternalErrorKey.INTERNAL_RUNTIME_EXCEPTION);
```
jmkulwik, line 38: Also include the exception as the cause. i.e. 
`throw new InternalServerRuntimeError(InternalErrorKey.INTERNAL_RUNTIME_EXCEPTION, e);`

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
```diff
@@ -9,8 +9,9 @@
 import java.io.ByteArrayOutputStream;
 import java.io.FileNotFoundException;
 import java.util.HashMap;
-import org.code.media.AudioWriter;
+import org.code.media.*;
 import org.code.media.Color;
+import org.code.media.Font;
 import org.code.media.Image;
```
jmkulwik, line 15: If we're using a * import, do we also need Color, Font, and Image enumerated separately?

org-code-javabuilder/theater/src/test/java/org/code/theater/StageTest.java
```diff
@@ -61,7 +64,7 @@ public void tearDown() {
   void drawLineDefaultsToBlack() {
     s.removeStrokeColor();
     s.drawLine(0, 0, 100, 100);
-    verify(graphics).setColor(Color.BLACK);
+    verify(graphics).setColor(java.awt.Color.BLACK);
```
jmkulwik, line 67: Why did this need to change to java.awt.Color?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
```diff
@@ -91,6 +91,11 @@ public String handleRequest(Map<String, String> lambdaInput, Context context) {
             dashboardHostname,
             useNeighborhood);
 
+    // manually set font configuration file since there is no font configuration on a lambda.
+    java.util.Properties props = System.getProperties();
+    // /opt is the folder all layer files go into.
+    props.put("sun.awt.fontconfig", "/opt/fontconfig.properties");
```
jmkulwik, line 97: (possibly out of scope) - we also set the connection ID on a global properties object that we maintain (see `setConnectionId` above). Should that move to this model as well?

org-code-javabuilder/media/src/main/java/org/code/media/FontHelper.java
```diff
@@ -0,0 +1,54 @@
...
+    return this.fontMap.get(
+        fontToFilenamePrefix.get(font) + fontStyleToFilenameSuffix.get(fontStyle));
+  }
+
+  private void populateFontMap() {
+    try {
+      for (Font font : Font.values()) {
+        for (FontStyle fontStyle : FontStyle.values()) {
+          String filename =
+              fontToFilenamePrefix.get(font) + fontStyleToFilenameSuffix.get(fontStyle);
```
sanchitmalhotra126, line 29: nit: separate method for `String getFileName(Font font, FontStyle fontStyle)` since it's used in a few places?

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
molly-moen: yep, the reason is that Color, Font and Image are all defined in `java.awt.*`. By enumerating them separately here we use a nice Java feature in which the class that is specifically imported is the one used by default. So in `Stage` if we reference `Color` (or Font or Image) it will use the `org.code.media` one instead of the `java.awt` one.

org-code-javabuilder/theater/src/test/java/org/code/theater/StageTest.java
molly-moen: similarly to above, we now need both `java.awt.Color` and `org.code.media.Color`. I made `org.code.media.Color` the default, so we have to explicitly reference `java.awt.Color` if we want to use it.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
molly-moen: good question. I'm not sure how the two relate. This is overwriting a property on the jvm to tell it to look in that location for the fontconfig file, instead of the default (which on the lambda does not exist). How do we use the properties object for the connection id?

org-code-javabuilder/media/src/main/java/org/code/media/FontHelper.java
molly-moen: done, thanks for catching!

sanchitmalhotra126 (approved):
Looks good!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
jmkulwik: The properties object used for the connection id essentially sets a global variable that needs to be available across the various packages.

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
jmkulwik: 👍

org-code-javabuilder/theater/src/test/java/org/code/theater/StageTest.java
jmkulwik: 👍

jmkulwik (approved):
LGTM!

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
moneppo: LGTM!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
molly-moen: makes sense, I think these might be separate cases but we can take a look as a follow-up as to whether we can put the connection id in the system properties

org-code-javabuilder/media/src/main/java/org/code/media/FontHelper.java
molly-moen: done!

template.yml
sureshc: Maybe name this `FontConfigurationLayer`? CloudFormation templates are declarative, so we don't need to use the verb `Add`.

sureshc (approved):
Do we need to bump up the memory allocated to the BuildAndRun Lambda? Are we in danger of exceeding the 50MB Lambda package size?

molly-moen:
> Do we need to bump up the memory allocated to the BuildAndRun Lambda? Are we in danger of exceeding the 50MB Lambda package size?

@sureshc the size of the lambda with this addition is 35.4 mb up from 34.2 mb, so not a big increase

PR #104, opened 2021-09-03
Handle wav file creation when there is no audio output by the student's project

jmkulwik:
A wav file will be created in an un-readable state if it has no data in it. This adds a single sample of silent audio to the generated wav file so it can be read by the client. 

This fixes a bug on firefox where an error while reading an empty audio file prevents the visual file from displaying.

Additional details start here: https://codedotorg.slack.com/archives/C01EF4GJ9GE/p1630693486126100

### DIFF
--- a/org-code-javabuilder/lib/src/test/java/org/code/javabuilder/InputRedirectionStreamTest.java
+++ b/org-code-javabuilder/lib/src/test/java/org/code/javabuilder/InputRedirectionStreamTest.java
@@ -4,6 +4,7 @@ import static org.junit.jupiter.api.Assertions.*;
 import static org.mockito.Mockito.*;
 
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import org.code.protocol.InputHandler;
 import org.code.protocol.InputMessageType;
 import org.junit.jupiter.api.BeforeEach;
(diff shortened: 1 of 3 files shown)


org-code-javabuilder/lib/src/test/java/org/code/javabuilder/InputRedirectionStreamTest.java
```diff
@@ -4,6 +4,7 @@
 import static org.mockito.Mockito.*;
 
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
```
jmkulwik, line 7: All changes in this file were unrelated to this change. They are test fixes that are required for tests to pass on Windows due to the different line separator used on Windows.

org-code-javabuilder/media/src/main/java/org/code/media/AudioWriter.java
```diff
@@ -68,6 +68,12 @@ public void addDelay(double delaySeconds) {
    * closes output streams.
    */
   public void writeToAudioStreamAndClose() {
+    if (this.audioSamples.length == 0) {
+      // Add a silent audio sample so we can build a valid wav file.
+      // TODO: Send a "No audio" signal instead
```
jmkulwik, line 73: https://codedotorg.atlassian.net/browse/CSA-793

molly-moen (approved):
🎉

org-code-javabuilder/lib/src/test/java/org/code/javabuilder/InputRedirectionStreamTest.java
sanchitmalhotra126: Ah interesting! Separate discussion, but should we not be using `System.lineSeparator()` then?

PR #105, opened 2021-09-08
Grant CodeBuild permission to lookup Hosted Zone ID

sureshc:
Grant AWS CodeBuild permission to [lookup the Hosted Zone ID](https://github.com/code-dot-org/javabuilder/blob/69eb6272a5c926e29dab2022f0928ed8c243a2ff/deploy.sh#L9) from the base domain name (code.org vs dev-code.org) of the Javabuilder service that's being provisioned. This enables AWS CodePipeline / CodeBuild ability to provision Javabuilder environments the same way an Engineer can.

### DIFF
--- a/iam.yml
+++ b/iam.yml
@@ -190,6 +190,10 @@ Resources:
                 Action:
                   - "cloudformation:DescribeStacks"
                 Resource: '*'
+              - Effect: Allow
+                Action:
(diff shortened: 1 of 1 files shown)


sanchitmalhotra126 (approved):

Hamms (approved):
I don't have a ton of context here, but based on what I *do* know this LGTM!

PR #106, opened 2021-09-09
Playground API

moneppo:
This is the PR for providing feedback on the API for Playground. Specs:
https://docs.google.com/document/d/14S47uuVF-hzxYeiw4ap-WqlN4A8ctUOypPNTwRKGh6c/edit#
https://docs.google.com/document/d/1zVCVa2MRvMQqt1m13_KEP8vjWcf61Rd8N_u3wuuic04/edit#


org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * Adds a clickable item from the playground.
+     * 
+     * @param item the item to add. If the item is already in the playground,
+     *              this method does nothing.
+     * @param x the distance, in pixels, from the left side of the playground
+     * @param y the distance, in pixels, from the top of the playground
+     */
+    public void addItem(ClickableItem item, int x, int y);
```
sanchitmalhotra126, line 43: We need width and height as well right?

org-code-javabuilder/playground/src/main/java/org/code/playground/Playground.java
```diff
@@ -0,0 +1,5 @@
+package org.code.playground;
+
+public static final class Playground {
+  public static final Board board = new Board();
```
sanchitmalhotra126, line 4: Nice, I like this pattern - will be easier for implementation and mirrors Theater/Stage

org-code-javabuilder/playground/src/main/java/org/code/playground/ClickableItem.java
```diff
@@ -0,0 +1,20 @@
+package org.code.playground;
+
+import java.io.FileNotFoundException;
+
+public abstract class ClickableItem extends Item {
```
sanchitmalhotra126, line 5: Discussed on Slack a bit already but just wanted to add a few thoughts - we decided that `ClickableItem` would not extend `Image`/depend on the `media` package to avoid confusion and that definitely makes sense. My only concern is that if we do decide later on to support more complicated image manipulation here (e.g. manipulating individual pixels/colors) then it will be a little more effort to add that capability back in as we'll have to bring back `Image` in some way. So as long as we're only ever going to allow users to display an image and only manipulate its position and width/height, then this will work, but just wanted to call this out.

org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
```diff
@@ -0,0 +1,124 @@
...
+     * @param text the text to draw
+     * @param x the distance from the left side of the playground to draw the text.
+     * @param y the distance from the top of the playground to draw the text.
+     * @param color the color to draw the text.
+     * @param font the name of the font to draw the text in.
+     * @param fontStyle the name of the font style to draw the text in.
+     * @param height the height of the text in pixels.
+     * @param rotation the rotation or tilt of the text, in degrees
+     */
+    public drawText(String text, int x, int y, Color color, Font font, FontStyle fontStyle, int height, double rotation);
```
jmkulwik, line 83: How are we planning to avoid confusion relating to the media package? i.e. I know we had decided to not extend Image to make clickableImage, but we're still having them import Color, Font, and FontStyle from the media package to use drawText.

org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
```diff
@@ -0,0 +1,124 @@
...
+    /**
+     * Returns the height of the playground screen. This will always be 400.
+     * 
+     * @return the height of the playground in pixels.
+     */
+    public int getHeight();
+
+    /**
+     * Sets the background of the playground to the provided image. The image will
+     * be scaled to fit the full playground screen, which may distory the image if
```
jmkulwik, line 25: NIT: `distory` -> `distort`

org-code-javabuilder/playground/src/main/java/org/code/playground/ClickableItem.java
jmkulwik: Similarly, I had thought we are planning to allow students to draw shapes - is this something we decided against?

org-code-javabuilder/playground/src/main/java/org/code/playground/ClickableItem.java
```diff
@@ -0,0 +1,20 @@
...
+
+    /**
+     * Sets a sound to play when the user clicks on this item.
+     * 
+     * @param filename the name of the sound file from the asset manager to play
+     * @throws FileNotFoundException when the sound file cannot be found.
+     */
+    public void setClickSound(String filename) throws FileNotFoundException;
```
jmkulwik, line 13: To confirm: We don't want to allow students to build their own sounds, correct? Just play pre-made sounds? Would we ever want to allow student-built sounds?

moneppo:
No, I think for the same reasons we won’t allow image manipulation -
performance and ensuring no confusion between this and the theater API

org-code-javabuilder/playground/src/main/java/org/code/playground/ClickableItem.java
moneppo: No, I think for the same reasons we won’t allow image manipulation - performance and ensuring no confusion between this and the theater API.

org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
moneppo: I think I misspoke about avoid confusion with the media package. I don't think there will be confusion in using the utilities we've created in `org.code.media` - I think of it like our own version of `java.awt.Color`. The confusion to avoid is overlap in capability with Theater, so including the full `Image` class in Playground could cause that kind of confusion.

org-code-javabuilder/playground/src/main/java/org/code/playground/ClickableItem.java
moneppo: Right now the only capabilities we know are needed are showing images at a specific position/size, but I do see a world where we would want to add pixel manipulation. I think that this could be pretty straightforward to add by creating new constructors for `ClickableItem` and `Item` that takes an `Image` object. If we wanted to add the ability to manipulate the image of an already showing `Item`/`ClickableItem`, e.g. `updateImage()` or `setImage(Image img)`/`setImage(String filename)`. But actually implementing those methods would be more involved.

As for shape drawing, it was a nice-to-have so I omitted it. We can add APIs in the future for that as well.

jmkulwik: @sanchitmalhotra126, how would supporting or not supporting shape drawing affect how we design the API?

sanchitmalhotra126: It will depend on implementation but if we are able to reuse Theater code using the more backend approach, then it won't be too much work to support drawing shapes, especially if we already have to support drawing text. We would be able to reuse any drawing functionality that exists in Theater already (text, lines, shapes).

org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
molly-moen: should this be addClickableItem?

org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
```diff
@@ -0,0 +1,124 @@
...
+     */
+    public void addItem(ClickableItem item, int x, int y);
+
+    /**
+     * Removes the clickable item from the playground.
+     * 
+     * @param item the item to remove. If the item is not in the playground, this
+     *              method does nothing.
+     */
+    public void removeClickableItem(ClickableItem item);
```
molly-moen, line 54: Since `ClickableItem` extends `Item`, does it make sense to just have one `addItem` and one `removeItem` method that takes the generic `Item`?

moneppo: Can you easily delineate between the two in the implementation? I thought downcasting was a pain in Java...

molly-moen: That's fair, it would make it easier on us to have both. If a student passes in a `ClickableItem` to `addItem` should we not support clicking it?

org-code-javabuilder/playground/src/main/java/org/code/playground/Item.java
```diff
@@ -0,0 +1,22 @@
...
+public class Item {
+  
+   /**
+   * Creates an item that can be displayed in the Playground. An item simply consists of an image,
+   * referenced by the name of the image file in the asset manager.
+   *
+   * @param filename the string name of the color (case insensitive).
+   * @throws FileNotFoundException if the file specified is not the in the asset manager
+   */
+  public Item(String filename) throws FileNotFoundException;
```
molly-moen, line 14: should the x, y width and height values live on `Item` instead of being set when the Item is added to the board? How would a click event change the x, y, width or height? Also should we have a setFilename option?

moneppo: Just chatted with Jamila on these and I had incorrectly thought that putting them on the API call was preferable, but that's not the case. I'll be changing these soon!

org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
moneppo: Yes, anything passed into the `addItem` method, event if it's a `ClickableItem`, should not have any click events registered.

org-code-javabuilder/playground/src/main/java/org/code/playground/Item.java
sanchitmalhotra126: Ah interesting, my impression was that the only way to change an item's position/size was to remove it and add it with a different position/size, but with this change it seems like that's not the case?

moneppo: Yes, now students should be able to move and resize things in this API.

org-code-javabuilder/playground/src/main/java/org/code/playground/ClickableItem.java
jmkulwik: 👍

org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
jmkulwik: 👍

org-code-javabuilder/playground/src/main/java/org/code/playground/ClickableItem.java
jmkulwik: 👍

jmkulwik (commented):
LGTM. Deferring my sign-off to @molly-moen and @sanchitmalhotra126 as they've been working closely in this space.

org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
```diff
@@ -0,0 +1,120 @@
...
+
+    /**
+     * Adds a non-clickable item to the board. 
+     *
+     * @param item the item to add. If the item is already on the board,
+     *             this method does nothing.
+     */
+    public void addItem(Item item, int x, int y, int width, int height);
```
sanchitmalhotra126, line 57: addItem wouldn't need x, y, width, height now right?

org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
sanchitmalhotra126: Another text question - would we need a way to remove text, or is text permanent once drawn?

sanchitmalhotra126 (commented):
Looks pretty good to me - a couple final questions

org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
```diff
@@ -0,0 +1,120 @@
...
+     * Sets the background of the board to the provided image. The image will
+     * be scaled to fit the full board, which may distort the image if
+     * it is not square.
+     * 
+     * @param filename the name of the file from the asset manager to put in the
+     *                 background
+     * @throws FileNotFoundException if the file cannot be found in the asset
+     *                               manager
+     */
+    public void setBackgroundImage(String filename) throws FileNotFoundException;
```
molly-moen, line 33: will we ever want to support setting the background to a color?

org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
```diff
@@ -0,0 +1,120 @@
...
+     * 
+     * @throws PlaygroundRunningException if the run() method has already been
+     *                                    called.
+     */
+    public void run() throws PlaygroundRunningException;
+
+    /**
+     * Ends the game, plays the sound supplied, and stops program execution.
+     * @param endingSound the name of a sound file in the asset manager to play at the end of the game.
+     * @throws PlaygroundNotRunningException if the run() method has not been
```
molly-moen, line 107: should this also be thrown if exit is called twice or is that ok?

org-code-javabuilder/playground/src/main/java/org/code/playground/Item.java
```diff
@@ -0,0 +1,83 @@
...
+   * @throws FileNotFoundException if the file specified is not the in the asset manager
+   */
+  public Item(String filename, int x, int y, int width, int height) throws FileNotFoundException;
+  
+  /**
+   * Retrieve the filename for this item.
+   *
+   * @return the filename in the asset manager for the image associated with this item.
+   */
+  public String getFilename();
```
molly-moen, line 26: should there also be a setFilename function (either here or on `ClickableItem`)

org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
moneppo: I think it's fine to be a no-op if they somehow manage to call this twice.

org-code-javabuilder/playground/src/main/java/org/code/playground/TextItem.java
```diff
@@ -0,0 +1,140 @@
...
+     * @param text the text content
+     * @param x the distance from the left side of the board
+     * @param y the distance from the top of the board
+     * @param color the color to draw the text
+     * @param font the name of the font to draw the text
+     * @param fontStyle the style of the font
+     * @param height the height of the text in pixels
+     * @param rotation the rotation or tilt of the text, in degrees
+     */
+    public TextItem(String text, Color color, Font font, FontStyle fontStyle, int height, double rotation);
```
sanchitmalhotra126, line 14: Constructor should also take x, y, width, and height right?

org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
```diff
@@ -0,0 +1,93 @@
...
+     */
+    public void setBackgroundImage(String filename) throws FileNotFoundException;
+
+    /**
+     * Adds a clickable item to the board.
+     * 
+     * @param item the item to add. If the item is already on the board,
+     *             this method does nothing.
+     */
+    public void addClickableItem(ClickableItem item);
```
bencodeorg, line 41: Not sure what level of nitiness we're going for in these API reviews, but the argument should be `ClickableImage` I think? Looks like we've been waffling between `ClickableImage` and `ClickableItem`, not sure which name is final?

moneppo: Fixed! Should be `ClickableImage`.

org-code-javabuilder/playground/src/main/java/org/code/playground/TextItem.java
moneppo: x,y and yes, height yes, but not width as text width varies based on the length of the string.

org-code-javabuilder/playground/src/main/java/org/code/playground/ImageItem.java
```diff
@@ -0,0 +1,91 @@
...
+   */
+  public String getFilename();
+  
+  /**
+   * Sets the filename for this item.
+   *
+   * @param filename the filename in the asset manager for the image to associat with this item
+   * @throws FileNotFoundException if the file specified is not the in the asset manager
+   */
+  public void getFilename(String filename) throws FileNotFoundException;
```
molly-moen, line 34: setFilename?

org-code-javabuilder/playground/src/main/java/org/code/playground/TextItem.java
```diff
@@ -0,0 +1,140 @@
+ public class TextItem implements Item {
```
molly-moen, line 1: clarifying question: text is never clickable?

moneppo: Correct.

molly-moen (approved):
LGTM!

org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
```diff
@@ -0,0 +1,101 @@
...
+     */
+    public void removeItem(Item item);
+    
+     /**
+     * Plays a sound from the asset manager.
+     * 
+     * @param filename the name of the sound file from the asset manager to play
+     * @throws FileNotFoundException when the sound file cannot be found.
+     */
+    public void playSound(String filename) throws FileNotFoundException;
```
sanchitmalhotra126, line 73: Now that we're set on the frontend approach, I realized throwing FileNotFoundExceptions get a little tricky - since we don't actually need to load the asset file in Javabuilder, we don't know if the file exists or not, but I'm guessing we want to keep these exceptions around since we expect the students to explicitly catch them (like in theater)? If so, a couple ideas:
- load the files just to check if they exist but send back filenames. not sure about this since it would just inflate load times without any performance or feature benefits
- pre-cache the list of asset file names in javabuilder and check if the given filename matches one of them. this would only make the initial load slower, but i'm also not sure how this would work if students can upload files to the asset manager while the game is running

@molly-moen thoughts?

sanchitmalhotra126 (commented):
One question around FileNotFoundExceptions following our discussion today, but otherwise looks good!

org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
molly-moen: good call out. Another option would be for javalab to send a list of filenames on program run, which might also speed up our image/audio load times. However, we currently don't have a way to send initial data to the lambda outside of the authorizer payload, and this doesn't make a ton of sense to include there. But we could do it if need be or figure out another method. Otherwise I think the second option would also work. I feel like it might make sense to not allow files to be used that were loaded after the game started running, because in 'real Java' you can't add extra files on the fly since everything is compiled.

moneppo: Definitely agree that we should not allow the program to use assets uploaded while the program is running.

sanchitmalhotra126 (approved):
Per our discussion today, LGTM!

PR #107, opened 2021-09-14
Create and Import Playground package

bencodeorg:
This PR creates the Playground package (without no-op API). This is a cherry-pick of one of Molly's first commits on the Playground prototype.

image

## Testing Story

Tested that I was able to import the Playground package, and get some dummy messages that are included in this PR (see screenshot above).


org-code-javabuilder/protocol/src/main/java/org/code/protocol/ClientMessageType.java
```diff
@@ -13,5 +13,6 @@
   /** A message directed to the client in local development mode */
   DEBUG,
   /** A status message directed to the client-side terminal, generated by Javabuilder. */
-  STATUS
+  STATUS,
+  PLAYGROUND
```
jmkulwik, line 17: NIT: Add a comment describing this message type

org-code-javabuilder/playground/src/main/java/org/code/playground/PlaygroundSignalKey.java
```diff
@@ -0,0 +1,5 @@
+package org.code.playground;
+
+public enum PlaygroundSignalKey {
+  ADD_IMAGE
```
jmkulwik, line 4: NIT: add a comment to the key (similar to NeighborhoodSignalKey's comments)

jmkulwik (approved):
Couple nits to add comments. Otherwise, LGTM!

org-code-javabuilder/playground/src/main/java/org/code/playground/Playground.java
```diff
@@ -0,0 +1,23 @@
...
+  public void sendTestMessages() {
+    HashMap<String, String> details = new HashMap<>();
+    details.put("filename", "puppy_1.jpg");
+    details.put("width", "200");
+    details.put("height", "200");
+    details.put("x", "200");
+    details.put("y", "200");
+    details.put("id", "first_puppy");
+    this.outputAdapter.sendMessage(new PlaygroundMessage(PlaygroundSignalKey.ADD_IMAGE, details));
+  }
```
sanchitmalhotra126, line 22: nit: we can remove this test code (or remove it when you implement the no-op API)

org-code-javabuilder/playground/src/main/java/org/code/playground/PlaygroundSignalKey.java
sanchitmalhotra126: Also these keys may likely change as we start implementing, so we can remove for now

org-code-javabuilder/playground/src/main/java/org/code/playground/Playground.java
```diff
@@ -0,0 +1,23 @@
+package org.code.playground;
+
+import java.util.HashMap;
+import org.code.protocol.*;
+
+public class Playground {
```
sanchitmalhotra126, line 6: nit: I think in the updated version of this API, `Playground` is a static final class - ok with merging this since it'll be updated shortly after when you implement the no-op API but just a heads up

sanchitmalhotra126 (approved):
Few small things but otherwise LGTM!

org-code-javabuilder/playground/src/main/java/org/code/playground/PlaygroundSignalKey.java
bencodeorg: Added a comment that this doesn't actually do anything at the moment, will delete as we actually build :)

org-code-javabuilder/playground/src/main/java/org/code/playground/Playground.java
bencodeorg: Yeah I'll remove then, was imagining this PR as just a way to get something that is runnable, and will clean up with no-op API.

org-code-javabuilder/playground/src/main/java/org/code/playground/Playground.java
bencodeorg: 👍 good to know, thanks! I'm having a bit of a hard time following the discussion on that PR, I see this comment:

https://github.com/code-dot-org/javabuilder/commit/63bc14a55dc97f7fc26837c3420a9b8d0324c40b#commitcomment-54117605

Can you help me understand why we'd want one vs. the other?

sanchitmalhotra126: Yep sure - it's just so that the actual student facing API for viewing and manipulating the playground (e.g. `addImage`, `getWidth`, etc) can be part of a non-static class (which is now `Board.java`), while `Playground` simply owns a singleton instance of `Board`. This way, students can use the API in a pseudo-static manner without needing to instantiate a new `Board` (via something like `Board myBoard = Playground.board`) but the API can remain non-static, which makes internal implementation easier as `Board` will have to manage a lot of internal state and instance variables.

PR #109, opened 2021-09-16
No-op Playground API

bencodeorg:
Cleans up https://github.com/code-dot-org/javabuilder/pull/106/files into a version that compiles. Not many major changes, but tried to put changes into individual commits for changes that weren't as simple as adding a return statement or a method body.

## Testing story

Here are two files I used in Javalab to confirm I could execute each method in the Playground package, in case anyone wants to test locally. 

### PlaygroundTesting.java

```
import org.code.playground.*;
import org.code.media.*;
import java.io.FileNotFoundException;

public class PlaygroundTesting {
    try {
      ImageItem image_item = new ImageItem("filename", -1, -1, -1, -1);
      image_item.getFilename();
      image_item.setFilename("filename");
      image_item.setX(-1);
      image_item.getX();
      image_item.setY(-1);
      image_item.getY();
      image_item.setWidth(-1);
      image_item.getWidth();
      image_item.setHeight(-1);
      image_item.getHeight(); 
    } catch (FileNotFoundException e) {
      System.out.println("there was an error");
    }

    TextItem text_item = new TextItem("text", -1, -1, Color.BLACK, Font.MONO, FontStyle.NORMAL, -1, -1.0);
    TextItem text_item_other_constructor = new TextItem("text", -1, -1, Color.BLACK, Font.MONO, -1, -1.0);
    text_item.setText("text");
    text_item.getText();
    text_item.setColor(Color.BLACK);
    text_item.getColor();
    text_item.setFont(Font.MONO);
    text_item.getFont();
    text_item.setFontStyle(FontStyle.NORMAL);
    text_item.getFontStyle();
    text_item.setRotation(-1.0);
    text_item.getRotation();
    text_item.setX(-1);
    text_item.getX();
    text_item.setY(-1);
    text_item.getY();
    text_item.setHeight(-1);
    text_item.getHeight();

    // test playground and board
    Playground playground = new Playground();
    Board board = playground.board;

    board.getWidth();
    board.getHeight();
    try {
      board.setBackgroundImage("filename");
      board.playSound("sound");
    } catch (FileNotFoundException e) {}
    board.addItem(text_item);
    board.removeItem(text_item);

    try {
      board.run();
    } catch (PlaygroundException e) {}
    board.exit();
    try {
      board.exit("ending_sound");
    } catch (PlaygroundException e) {}
    catch (FileNotFoundException e) {}

    try {
      MyClickableImage my_clickable_image = new MyClickableImage("filename", -1, -1, -1, -1);
      board.addClickableImage(my_clickable_image);
      board.removeClickableImage(my_clickable_image);
    } catch (FileNotFoundException e) {}
  }
}
```

### MyClickableImage.java

```
import org.code.playground.ClickableImage;
import java.io.FileNotFoundException;

public class MyClickableImage extends ClickableImage {
  public MyClickableImage(String filename, int x, int y, int width, int height) throws FileNotFoundException {
    super(filename, x, y, width, height); 
  }

  public void onClick() {}
}
```


org-code-javabuilder/playground/src/main/java/org/code/playground/Item.java
```diff
@@ -0,0 +1,31 @@
+package org.code.playground;
+
+interface Item {
```
molly-moen, line 3: there's been some discussion about changing this interface to a base class [link](https://docs.google.com/document/d/1Moo2s5EXZRp5rMg1VW9jlOqs_GeMN5yjU8FJgoqOEMk/edit?disco=AAAAPONopfg). Since discussion is still ongoing this could be done as a follow-up, but just an fyi

bencodeorg: I saw that :) I'm inclined to do as a follow-up to keep this moving if that sounds ok!

molly-moen (approved):
LGTM!

org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
```diff
@@ -0,0 +1,94 @@
+package org.code.playground;
+
+import java.io.FileNotFoundException;
+
+public class Board {
```
sanchitmalhotra126, line 5: We could add a protected constructor, but not required - that can also be done when we start actually implementing this class.

sanchitmalhotra126 (approved):
LGTM! Like Molly mentioned a few things may change as per ongoing discussion in the dev doc, but this should be good to merge for now.

org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
bencodeorg: Added!

PR #110, opened 2021-09-17
Make Item abstract class instead of interface

bencodeorg:
Simplify Item and sub-classes by making Item an abstract class instead of an interface. Suggested/discussed in the [Playground dev doc](https://docs.google.com/document/d/1Moo2s5EXZRp5rMg1VW9jlOqs_GeMN5yjU8FJgoqOEMk/edit#).

## Testing story

Confirmed that my script to execute all methods ran as it did when I merged the original no-op PR ([see test code there if interested](https://github.com/code-dot-org/javabuilder/pull/109)).

### DIFF
--- a/org-code-javabuilder/playground/src/main/java/org/code/playground/Item.java
+++ b/org-code-javabuilder/playground/src/main/java/org/code/playground/Item.java
@@ -1,6 +1,6 @@
 package org.code.playground;
 
-abstract class Item {
+public abstract class Item {
   /**
    * Set the X position for the item.
    *
--- a/org-code-javabuilder/playground/src/main/java/org/code/playground/Item.java
+++ b/org-code-javabuilder/playground/src/main/java/org/code/playground/Item.java
@@ -1,31 +1,51 @@
 package org.code.playground;
 
-interface Item {
+public abstract class Item {
   /**
    * Set the X position for the item.
    *
    * @param x the distance, in pixels, from the left side of the board
    */
-  public void setX(int x);
+  public void setX(int x) {}
 
   /**
    * Get the X position for the item.
    *
    * @return the distance, in pixels, from the left side of the board
    */
-  public int getX();
+  public int getX() {
+    return -1;
+  }
 
   /**
    * Set the Y position for the item.
    *
    * @param y the distance, in pixels, from the top of the board
    */
-  public void setY(int y);
+  public void setY(int y) {}
 
   /**
    * Get the Y position for the item.
    *
    * @return the distance, in pixels, from the top of the board
    */
-  public int getY();
+  public int getY() {
+    return -1;
+  }
+
+  /**
+   * Set the height for the item.
+   *
+   * @param height the height of the item, in pixels
+   */
+  public void setHeight(int height) {}
+
+  /**
+   * Get the height for the item.
+   *
+   * @return the height of the item, in pixels
+   */
+  public int getHeight() {
+    return -1;
+  }
 }
--- a/org-code-javabuilder/playground/src/main/java/org/code/playground/Item.java
+++ b/org-code-javabuilder/playground/src/main/java/org/code/playground/Item.java
@@ -1,6 +1,6 @@
 package org.code.playground;
 
-public abstract class Item {
+abstract class Item {
   /**
    * Set the X position for the item.
    *
--- a/org-code-javabuilder/playground/src/main/java/org/code/playground/Item.java
+++ b/org-code-javabuilder/playground/src/main/java/org/code/playground/Item.java
@@ -1,6 +1,8 @@
 package org.code.playground;
 
 public abstract class Item {
+  Item(int x, int y, int height) {}
+
   /**
    * Set the X position for the item.
    *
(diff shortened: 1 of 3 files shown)


org-code-javabuilder/playground/src/main/java/org/code/playground/Item.java
```diff
@@ -1,31 +1,51 @@
 package org.code.playground;
 
-interface Item {
+public abstract class Item {
```
molly-moen: what's the reasoning for making this an abstract class?

molly-moen: (instead of a regular class that is)

bencodeorg: Could be totally misunderstanding things, but I thought abstract classes were ones that you do not want to be instantiated directly, but rather subclassed (and then instantiated via those subclasses). Isn't that what we want in this case?

bencodeorg: Also updated to not public b/c I don't think we want students to be able to subclass it directly?

molly-moen: yep you're right on abstract, I was thinking since it didn't have abstract methods it shouldn't be abstract but that's not the case. I think we do want students to be able to reference it like `Item x = new TextItem()`. Will that still work with not making it public?

bencodeorg: mmm not sure I understand that syntax, what does it mean to create an object and mark its type as it's parent? I believe I just tested `TextItem x = new TextItem()`?

molly-moen: It's a pretty common java pattern, it just means that it is some subclass of the parent (or the parent itself it it's not an abstract class).

bencodeorg: Got it, and if you mark the object as type `Item` you wouldn't have access to methods defined on the subclass, yeah? It looks like it works if I mark the class as `public abstract` rather than just `abstract`

molly-moen: yep!

bencodeorg: Alrighty thanks! Made Item public then :)

molly-moen (approved):
LGTM!

org-code-javabuilder/playground/src/main/java/org/code/playground/Item.java
sanchitmalhotra126, line 3: I think we'll want a constructor for `Item` that takes in x, y, and height

sanchitmalhotra126 (commented):
Nice! I think it just needs a constructor

org-code-javabuilder/playground/src/main/java/org/code/playground/Item.java
bencodeorg: For my understanding, can you elaborate? We would want students to be able to create ImageItems with no filename, or TextItems with no text?

sanchitmalhotra126: Sure thing - so because `Item` is declared abstract, it cannot be instantiated anyway. Students would still only ever instantiate `ImageItem/TextItem/ClickableImage` instead of `Item`s directly, but a constructor would allow `Item` to set some common properties that subclasses could reference. Instead of each subclass needing to set x, y, and height themselves, these would be set by the superclass which just saves some boilerplate. For example, `Item` would be implemented something like:
```
protected Item(int x, int y, int height) {
   this.x = x;
   this.y = y;
   this.height = height;
}

...

public getX() {
   return this.x;
}

public setX(int x) {
   this.x = x;
}

...
```

And then subclasses (for example, ImageItem) would be implemented like
```
public ImageItem(int x, int y, int width, int height, String imageUrl) {
   super(x, y, height);
   this.width = width;
   this.imageUrl = imageUrl;
}

...

public getX() {
   return super.getX();
}

public setX(int x) {
   super.setX(x);
}

...

public getWidth() {
   return this.width;
}
```

So ImageItem would only need to worry about explicitly getting and setting properties specific to it (in this case width and imageUrl) and defer to its parent class for everything else.

FWIW, this will come in handy later when we have these items send Playground updates; for each `set` method, we will need to have the class send an update message, and having `Item` be a base class allows `Item` to own the updating logic and subclasses can just call a superclass method, instead each of them needing to having updating logic as well.

bencodeorg: Great explanation, got it thank you!

bencodeorg:
@sanchitmalhotra126 how does the last commit look?

sanchitmalhotra126 (approved):
LGTM!

sanchitmalhotra126:
> @sanchitmalhotra126 how does the last commit look?

Looks good! Thanks for the updates

PR #112, opened 2021-09-20
Create Playground Message Handler

sanchitmalhotra126:
Creates a PlaygroundMessageHandler class that takes care of routing Playground messages. For now this is nothing more than a wrapper around `outputAdapter.sendMessage()` but it allows us to isolate message related logic to support other use cases (like message batching) down the road.

Spec: https://docs.google.com/document/d/1Moo2s5EXZRp5rMg1VW9jlOqs_GeMN5yjU8FJgoqOEMk/edit#bookmark=id.7kl6bmoljerq

JIRA: https://codedotorg.atlassian.net/browse/CSA-832

Testing: unit


molly-moen (approved):
🎉

PR #113, opened 2021-09-20
Add Item constructor, getters, and setters

bencodeorg:
Implements basic constructors, getters, and setters for `Item`, `ImageItem`, and `TextItem`.

## Testing story

I modified a project where I have each of these getters and setters being executed and saw them each change value when re-assigned. I did not add unit tests, as it doesn't look like we test getters and setters in our other mini-apps. Can add though if desired!

### DIFF
--- a/org-code-javabuilder/playground/src/main/java/org/code/playground/Item.java
+++ b/org-code-javabuilder/playground/src/main/java/org/code/playground/Item.java
@@ -1,14 +1,24 @@
 package org.code.playground;
 
 public abstract class Item {
-  Item(int x, int y, int height) {}
+  private int xLocation;
+  private int yLocation;
+  private int height;
+
+  Item(int x, int y, int height) {
+    this.xLocation = x;
+    this.yLocation = y;
+    this.height = height;
+  }
 
   /**
    * Set the X position for the item.
    *
    * @param x the distance, in pixels, from the left side of the board
    */
-  public void setX(int x) {}
+  public void setX(int x) {
+    this.xLocation = x;
+  }
 
   /**
    * Get the X position for the item.
@@ -16,7 +26,7 @@ public abstract class Item {
    * @return the distance, in pixels, from the left side of the board
    */
   public int getX() {
-    return -1;
+    return this.xLocation;
   }
 
   /**
@@ -24,7 +34,9 @@ public abstract class Item {
    *
    * @param y the distance, in pixels, from the top of the board
    */
-  public void setY(int y) {}
+  public void setY(int y) {
+    this.yLocation = y;
+  }
 
   /**
    * Get the Y position for the item.
@@ -32,7 +44,7 @@ public abstract class Item {
    * @return the distance, in pixels, from the top of the board
    */
   public int getY() {
-    return -1;
+    return this.yLocation;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


org-code-javabuilder/playground/src/main/java/org/code/playground/TextItem.java
```diff
@@ -42,85 +53,99 @@ public TextItem(
    */
   public TextItem(String text, int x, int y, Color color, Font font, int height, double rotation) {
     super(x, y, height);
+    this.text = text;
+    this.color = color;
+    this.font = font;
+    this.rotation = rotation;
```
sanchitmalhotra126, line 59: Might be good to default `this.fontStyle` here to `FontStyle.NORMAL` since it's not provided, to prevent some unfriendly null pointer exceptions?

sanchitmalhotra126 (commented):
One comment, otherwise looks good! Regarding testing - agreed that testing getters/setters is probably not necessary, but setting up the tests now could be useful for when we add more functionality we want to test later. If you want to hold off on tests until then though, that's also fine with me.

org-code-javabuilder/playground/src/main/java/org/code/playground/Item.java
```diff
@@ -1,53 +1,67 @@
 package org.code.playground;
 
 public abstract class Item {
-  Item(int x, int y, int height) {}
+  private int xLocation;
+  private int yLocation;
+  private int height;
+
+  Item(int x, int y, int height) {
```
molly-moen, line 8: should item also take a width?

molly-moen: or is there no width because `TextItem` doesn't have a width?

bencodeorg: Yeah, no width in `TextItem`, at least in spec. That could be wrong though?

org-code-javabuilder/playground/src/main/java/org/code/playground/TextItem.java
```diff
@@ -42,85 +53,99 @@ public TextItem(
    */
   public TextItem(String text, int x, int y, Color color, Font font, int height, double rotation) {
```
molly-moen, line 54: you can have this constructor call the more specific constructor and set the default font style like this
```
this(text, x, y, color, font, FontStyle.NORMAL, height, rotation)
```
then you won't need to repeat code

org-code-javabuilder/playground/src/main/java/org/code/playground/Item.java
molly-moen: I think that makes sense, since the width is determined by the text itself

molly-moen (commented):
Are you planning to add `ClickableItem` as a separate PR?

org-code-javabuilder/playground/src/main/java/org/code/playground/Item.java
sanchitmalhotra126: > I think that makes sense, since the width is determined by the text itself

Yeah, I believe this is intended

org-code-javabuilder/playground/src/main/java/org/code/playground/TextItem.java
bencodeorg: Nice, thanks!

molly-moen:
> > Are you planning to add `ClickableItem` as a separate PR?
> 
> @molly-moen is there anything to add to `ClickableItem` from a getter/setter/constructor perspective?

oh I guess not, nvm!

molly-moen (approved):
LGTM!

sanchitmalhotra126 (approved):

PR #115, opened 2021-09-21
Playground: implement add and remove item handlers

molly-moen:
Implement the add and remove handlers for all Item types (`ClickableImage`, `ImageItem`, `TextItem`). This includes a change to the api to remove the `addItem` method and replace it with `addImageItem` and `addTextItem`.

## Links
- [jira](https://codedotorg.atlassian.net/browse/CSA-834)

### DIFF
--- a/org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
+++ b/org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
@@ -17,8 +17,8 @@ public class Board {
   private boolean firstRunStarted;
   private boolean isRunning;
 
-  private HashMap<String, ClickableImage> clickableImages;
-  private HashMap<String, Item> items;
+  private final HashMap<String, ClickableImage> clickableImages;
+  private final HashMap<String, Item> items;
   private int nextItemIndex;
 
   protected Board() {
@@ -71,7 +71,8 @@ public class Board {
     if (this.clickableImages.containsKey(image.getId())) {
       return;
     }
-    HashMap<String, String> details = this.getImageItemDetails(image);
+    HashMap<String, String> details = image.getDetails();
+    this.addIndexToDetails(details);
     this.clickableImages.put(image.getId(), image);
     this.playgroundMessageHandler.sendMessage(
         new PlaygroundMessage(PlaygroundSignalKey.ADD_CLICKABLE_ITEM, details));
@@ -87,7 +88,7 @@ public class Board {
       return;
     }
     this.clickableImages.remove(image.getId());
-    HashMap<String, String> details = this.getRemoveItemDetails(image);
+    HashMap<String, String> details = image.getRemoveDetails();
     this.playgroundMessageHandler.sendMessage(
         new PlaygroundMessage(PlaygroundSignalKey.REMOVE_ITEM, details));
   }
@@ -99,13 +100,7 @@ public class Board {
    *     nothing.
    */
   public void addImageItem(ImageItem item) {
-    if (this.items.containsKey(item.getId())) {
-      return;
-    }
-    this.items.put(item.getId(), item);
-    HashMap<String, String> itemDetails = this.getImageItemDetails(item);
-    this.playgroundMessageHandler.sendMessage(
-        new PlaygroundMessage(PlaygroundSignalKey.ADD_IMAGE_ITEM, itemDetails));
+    this.addItemHelper(item, PlaygroundSignalKey.ADD_IMAGE_ITEM);
   }
 
   /**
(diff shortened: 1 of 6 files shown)


org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
```diff
@@ -146,4 +202,40 @@ public void exit() throws PlaygroundException {
...
+  }
+
+  private HashMap<String, String> getItemDetails(Item item) {
+    HashMap<String, String> details = new HashMap<>();
+    details.put("height", Integer.toString(item.getHeight()));
+    details.put("x", Integer.toString(item.getX()));
+    details.put("y", Integer.toString(item.getY()));
+    details.put("id", item.getId());
+    details.put("index", Integer.toString(this.nextItemIndex));
+    this.nextItemIndex++;
```
sanchitmalhotra126, line 232: Is this for z-indexing on the client?

org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
```diff
@@ -60,28 +67,77 @@ public void setBackgroundImage(String filename) throws FileNotFoundException {}
...
+  }
 
   /**
    * Removes the non-clickable item from the board.
    *
    * @param item the item to remove. If the image is not on the board, this method does nothing.
    */
-  public void removeItem(Item item) {}
+  public void removeItem(Item item) {
+    if (!(this.items.containsKey(item.getId()) || this.clickableImages.containsKey(item.getId()))) {
```
sanchitmalhotra126, line 133: Why are we checking `this.clickableImages` here as well? We're only removing from `this.items` right?

org-code-javabuilder/playground/src/test/java/org/code/playground/ClickableImageHelper.java
```diff
@@ -0,0 +1,27 @@
+package org.code.playground;
+
+import java.io.FileNotFoundException;
+
+public class ClickableImageHelper extends ClickableImage {
```
sanchitmalhotra126, line 5: NIce!

sanchitmalhotra126 (commented):
Looks good! A couple questions

org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
molly-moen: oh good point, I forgot we have `removeClickableImage` 🤦‍♀️

org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
```diff
@@ -146,4 +202,40 @@ public void exit() throws PlaygroundException {
...
+  private HashMap<String, String> getItemDetails(Item item) {
+    HashMap<String, String> details = new HashMap<>();
+    details.put("height", Integer.toString(item.getHeight()));
+    details.put("x", Integer.toString(item.getX()));
+    details.put("y", Integer.toString(item.getY()));
+    details.put("id", item.getId());
+    details.put("index", Integer.toString(this.nextItemIndex));
+    this.nextItemIndex++;
+    return details;
+  }
```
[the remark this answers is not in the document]
molly-moen: good call, I'll refactor

org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
```diff
@@ -60,28 +67,66 @@ public void setBackgroundImage(String filename) throws FileNotFoundException {}
...
-  public void addClickableImage(ClickableImage image) {}
+  public void addClickableImage(ClickableImage image) {
+    if (this.clickableImages.containsKey(image.getId())) {
+      return;
+    }
+    HashMap<String, String> details = image.getDetails();
+    this.addIndexToDetails(details);
+    this.clickableImages.put(image.getId(), image);
+    this.playgroundMessageHandler.sendMessage(
+        new PlaygroundMessage(PlaygroundSignalKey.ADD_CLICKABLE_ITEM, details));
```
sanchitmalhotra126, line 78: Could we reuse `addItemHelper` here too if we passed it the list to add to or an `isClickable` boolean maybe?

sanchitmalhotra126 (approved):
One small potential refactor otherwise LGTM!

org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
molly-moen: Since the `clickableImages` hash needs to have `<String, ClickableImage>` as the type instead of `<String, Item>`, it would end up requiring doing some if/else that doesn't seem super clean to me. I think I will leave it as is since it only ends up being 3 repeat lines of code.

sanchitmalhotra126: Ah good point, sounds good

PR #116, opened 2021-09-21
Downgrade image quality from assets

jmkulwik:
We display student-created images on a 400x400 canvas. We also downgrade the image quality when collecting images in a gif to pass back to the client. Therefore, it doesn't make sense to have high quality images be used during backend processing. This PR adjusts the images when they are imported so they maintain their aspect ratio, but display with a maximum width & height of 400px.

This is especially important for units where students do large amounts of pixel manipulation and iterate over every pixel in an image. Looking at https://studio.code.org/s/csa-u7-dev/lessons/1/levels/1, the program finishes in 37s with this change rather than 74s without it, resulting in a 2x improvement.

The image quality of the final result is visually similar due to the reasons explained above, but uses ~64% of the full-quality size.

With downgraded quality
https://drive.google.com/file/d/15So036TPygC0ZFF_OcB_s4e2PWNmgyXo/view?usp=sharing
theaterImage (2)

With full quality
https://drive.google.com/file/d/1yzJsrxvnzzh0NBP9e8gQYEx5K-7RcakL/view?usp=sharing
(Gif could not display inline because it was too big)

NOTE: This could break existing projects where a user hard-coded the width or height of an image. Very few users have started the Theater units yet and should generally be discouraged from hard-coding in this way, but this is something we should keep in mind as this is merged and deployed.


org-code-javabuilder/media/src/main/java/org/code/media/Image.java
```diff
@@ -156,17 +157,39 @@ public BufferedImage getBufferedImage() {
...
+    }
+    int targetHeight, targetWidth;
+    if (height > width) {
+      targetHeight = 400;
+      targetWidth = (int) ((double) width / height * 400);
+    } else {
+      targetWidth = 400;
+      targetHeight = (int) ((double) height / width * 400);
+    }
+    BufferedImage image = new BufferedImage(targetWidth, targetHeight, BufferedImage.TYPE_INT_RGB);
```
molly-moen, line 187: I think we might want this to be of type `TYPE_INT_ARGB` so we can maintain opacity

org-code-javabuilder/media/src/main/java/org/code/media/Image.java
```diff
@@ -156,17 +157,39 @@ public BufferedImage getBufferedImage() {
...
-      return image;
     } catch (IOException e) {
       throw new FileNotFoundException("Could not find file " + filename);
     }
+
+    // Resize the image to a max size of 400x400px while maintaining the aspect ratio.
+    // This allows us to save time when students run on pixel manipulation on an entire image.
+    int height = originalImage.getHeight();
+    int width = originalImage.getWidth();
+    if (width <= 400 && height <= 400) {
```
sanchitmalhotra126, line 176: nit: could we have constants for max width and max height in case we want to update these?

sanchitmalhotra126 (approved):
LGTM!

org-code-javabuilder/media/src/main/java/org/code/media/Image.java
jmkulwik: 👍

org-code-javabuilder/media/src/main/java/org/code/media/Image.java
jmkulwik: 👍

PR #117, opened 2021-09-21
Playground: Set background image

sanchitmalhotra126:
Sends a SET_BACKGROUND_IMAGE message with the background image file name.

I also created an ~~enum~~ class for `ClientMessageDetailKeys` that would contain common keys we use in the detail object that gets sent with client messages. I'm curious for feedback on this - I added this because, as we'll be introducing a lot of these for Playground, I felt like it would be good for us to have all our keys in one place and standardize the naming a little bit so it's easier to reference. But I'm not sure if it's overkill, as a lot of the keys will still be app-specific (for example all the playground x, y, width, height, etc keys). Still though, this can at least house some common keys like `filename`, `url`, and `id` for example. Additionally, if we think it would be helpful, we can also subclass `ClientMessageDetailKeys` for app-specific keys (for example `PlaygroundClientMessageDetailKeys` which can contain enums for X, Y, WIDTH, HEIGHT, etc)

Spec: https://docs.google.com/document/d/1Moo2s5EXZRp5rMg1VW9jlOqs_GeMN5yjU8FJgoqOEMk/edit#heading=h.yth48qvin8gl
JIRA: https://codedotorg.atlassian.net/browse/CSA-838

Tested locally with All The Things Playground level & unit tests

### DIFF
--- a/org-code-javabuilder/protocol/src/main/java/org/code/protocol/ClientMessageDetailKeys.java
+++ b/org-code-javabuilder/protocol/src/main/java/org/code/protocol/ClientMessageDetailKeys.java
@@ -1,6 +1,6 @@
 package org.code.protocol;
 
 /** Expected keys in the optional detail object of {@link ClientMessage}s */
-public final class ClientMessageDetailKeys {
+public class ClientMessageDetailKeys {
   public static final String FILENAME = "filename";
 }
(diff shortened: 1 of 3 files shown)


org-code-javabuilder/protocol/src/main/java/org/code/protocol/ClientMessageDetailKeys.java
```diff
@@ -0,0 +1,16 @@
+package org.code.protocol;
+
+/** Expected keys in the optional detail object of {@link ClientMessage}s */
+public enum ClientMessageDetailKeys {
```
molly-moen, line 4: I like the idea of having these in one place, but I wonder if it would be simpler to instead have a class of static strings, since we will only ever use these in their String value, not their enum value. Then we don't need to call `.toString()` every time we use them.

molly-moen (approved):

PR #118, opened 2021-09-21
Send message on playSound

bencodeorg:
Sends a message to Javalab to play a sound when `playSound` is called. Also implements the exit method where students can play a particular sound on exit from their program.

## Testing story

Added unit tests, and confirmed that I got the appropriate message if I called board.playSound('filename') directly (should this be allowed? I think so per [this slack conversation](https://codedotorg.slack.com/archives/C01EF4GJ9GE/p1632255162336100?thread_ts=1632253744.333100&cid=C01EF4GJ9GE))

### DIFF
--- a/org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
+++ b/org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
@@ -143,7 +143,7 @@ public class Board {
    */
   public void playSound(String filename) throws FileNotFoundException {
     HashMap<String, String> details = new HashMap<>();
-    details.put("filename", filename);
+    details.put(ClientMessageDetailKeys.FILENAME, filename);
 
     PlaygroundMessage playSoundMessage =
         new PlaygroundMessage(PlaygroundSignalKey.PLAY_SOUND, details);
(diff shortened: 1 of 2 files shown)


org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
```diff
@@ -141,7 +141,14 @@ public void removeItem(Item item) {
    * @param filename the name of the sound file from the asset manager to play
    * @throws FileNotFoundException when the sound file cannot be found.
    */
-  public void playSound(String filename) throws FileNotFoundException {}
+  public void playSound(String filename) throws FileNotFoundException {
+    HashMap<String, String> details = new HashMap<>();
+    details.put("filename", filename);
```
molly-moen, line 146: Sanchit added this new class with constants for sending messages which already has a FILENAME constant: https://github.com/code-dot-org/javabuilder/blob/main/org-code-javabuilder/protocol/src/main/java/org/code/protocol/ClientMessageDetailKeys.java

molly-moen (approved):
LGTM, just one nit about a constant

PR #120, opened 2021-09-22
Playground: handle updates

molly-moen:
Handle `Item` and all `Item` subclass updates by sending an output message whenever a setter is called. This will enable the frontend to receive update information and update the game appropriately. Also actually call `ClickableImage` click events when they occur.

I also refactored out the constants created in [this PR](https://github.com/code-dot-org/javabuilder/pull/115) into the new `ClientMessageDetailKeys` class.

## Links
- [jira](https://codedotorg.atlassian.net/browse/CSA-835)

## Testing Story
This isn't testable beyond some basic unit tests right now because we would need to test with 2-way communication to verify the click events are working. I added unit tests to verify the setters send the expected messages. It would be good to find a way to do this 2-way communication in our unit tests, but it is out of scope for now.

### DIFF
--- a/org-code-javabuilder/playground/src/main/java/org/code/playground/Item.java
+++ b/org-code-javabuilder/playground/src/main/java/org/code/playground/Item.java
@@ -8,6 +8,7 @@ public abstract class Item {
   private int yLocation;
   private int height;
   private final String id;
+  private final PlaygroundMessageHandler playgroundMessageHandler;
 
   private final String HEIGHT_KEY = "height";
   private final String X_KEY = "x";
@@ -19,6 +20,7 @@ public abstract class Item {
     this.yLocation = y;
     this.height = height;
     this.id = UUID.randomUUID().toString();
+    this.playgroundMessageHandler = PlaygroundMessageHandler.getInstance();
   }
 
   /**
@@ -28,6 +30,7 @@ public abstract class Item {
    */
   public void setX(int x) {
     this.xLocation = x;
+    this.sendChangeMessage(X_KEY, Integer.toString(x));
   }
 
   /**
@@ -46,6 +49,7 @@ public abstract class Item {
    */
   public void setY(int y) {
     this.yLocation = y;
+    this.sendChangeMessage(Y_KEY, Integer.toString(y));
   }
 
   /**
@@ -64,6 +68,7 @@ public abstract class Item {
    */
   public void setHeight(int height) {
     this.height = height;
+    this.sendChangeMessage(HEIGHT_KEY, Integer.toString(height));
   }
 
   /**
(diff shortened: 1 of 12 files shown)


org-code-javabuilder/playground/src/main/java/org/code/playground/Item.java
```diff
@@ -80,17 +81,34 @@ protected String getId() {
...
+    details.put(ClientMessageDetailKeys.X, Integer.toString(this.getX()));
+    details.put(ClientMessageDetailKeys.Y, Integer.toString(this.getY()));
     return details;
   }
 
   protected HashMap<String, String> getRemoveDetails() {
-    HashMap<String, String> details = new HashMap<>();
-    details.put(ID_KEY, this.getId());
+    HashMap<String, String> details = this.getIdDetails();
     return details;
```
sanchitmalhotra126, line 93: nit: simplify to `return this.getIdDetails()`?

org-code-javabuilder/playground/src/test/java/org/code/playground/BoardTest.java
```diff
@@ -28,6 +26,13 @@ public void setUp() {
     messageCaptor = ArgumentCaptor.forClass(PlaygroundMessage.class);
     inputHandler = mock(InputHandler.class);
     unitUnderTest = new Board(playgroundMessageHandler, inputHandler);
+    GlobalProtocol.create(
+        mock(OutputAdapter.class),
+        mock(InputAdapter.class),
+        "",
+        "",
+        "",
+        mock(JavabuilderFileWriter.class));
```
sanchitmalhotra126, line 35: What is this used for?

org-code-javabuilder/playground/src/main/java/org/code/playground/TextItem.java
```diff
@@ -164,13 +160,25 @@ public double getRotation() {
...
-    details.put(ROTATION_KEY, Double.toString(this.getRotation()));
+    details.put(ClientMessageDetailKeys.TEXT, this.getText());
+    this.addColorToDetails(details);
+    details.put(ClientMessageDetailKeys.FONT, this.getFont().toString());
+    details.put(ClientMessageDetailKeys.FONT_STYLE, this.getFontStyle().toString());
+    details.put(ClientMessageDetailKeys.ROTATION, Double.toString(this.getRotation()));
     return details;
   }
+
+  private void setColorHelper(Color color) {
```
sanchitmalhotra126, line 171: Not exactly related to this change but since we split out color into its components here, should we do the same for `getColor` too? i.e. return a new color with the RGB values instead of original color object?

sanchitmalhotra126 (approved):
LGTM!

org-code-javabuilder/playground/src/test/java/org/code/playground/BoardTest.java
molly-moen: it's needed because `PlaygroundMessageHandler` calls `GlobalProtocol.getInstance().getOutputAdapter()`

org-code-javabuilder/playground/src/main/java/org/code/playground/TextItem.java
molly-moen: I'll ask curriculum. My thought was it would be confusing for students to have a different object returned by get, but since we lock in the rgb values it might be confusing to have the same object too

molly-moen: so it turns out we'll need to do some reworking of the color messages which I think would be best as a follow-up PR. [slack conversation](https://codedotorg.slack.com/archives/C01EF4GJ9GE/p1632255167336300) I'll leave this as-is for now

bencodeorg: Late to the party, but was just reading that slack thread...we want `textItem.getColor().setRed()` to set an item's color? Does that mean that `getColor()` would return the `textItem`?

bencodeorg: Oh nvm, it's returning the text item's color object, which we're setting to another color

bencodeorg: Wrapping my head around how we have colors set up and made a couple silly comments I just deleted, ignore them :).

molly-moen: sorry for the confusion that was not exactly the correct syntax in the thread. It would be something like `textItem.getColor().setRed(255)`. The Color object has red, green and blue values. So we would be setting the red value to some new value, and want the text color to change

bencodeorg: Got it, should have looked at the Color class. Thanks!

PR #122, opened 2021-09-22
Playground: Throw FileNotFoundExceptions

sanchitmalhotra126:
Adds functionality to load all the file names in a student's project and check if a given file is part of the project, so we can verify and throw FileNotFoundExceptions in Javabuilder. Some of this work was already being done by AssetUrlGenerator, so I've added this functionality to that class, which I also renamed AssetFileHelper to avoid confusion.

Note: this has some overlap (specifically in the tests) with https://github.com/code-dot-org/javabuilder/pull/120 - I'll update this PR once that is merged.

Also note: on the client, FileNotFoundExceptions currently show up as UserInitiatedExceptions which are translated as generic runtime exceptions. We have an existing tech debt task to handle and translate FileNotFoundExceptions on the client which will make this more user-friendly (https://codedotorg.atlassian.net/browse/CSA-553)

JIRA: https://codedotorg.atlassian.net/browse/CSA-840
Spec: https://docs.google.com/document/d/1Moo2s5EXZRp5rMg1VW9jlOqs_GeMN5yjU8FJgoqOEMk/edit#bookmark=id.dl49xj5tnbcf

Testing: local all the things playground & unit

### DIFF
--- a/org-code-javabuilder/protocol/src/main/java/org/code/protocol/AssetFileHelper.java
+++ b/org-code-javabuilder/protocol/src/main/java/org/code/protocol/AssetFileHelper.java
@@ -1,5 +1,6 @@
 package org.code.protocol;
 
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.net.URI;
 import java.net.http.HttpClient;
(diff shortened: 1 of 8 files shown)


org-code-javabuilder/protocol/src/main/java/org/code/protocol/AssetFileHelper.java
```diff
@@ -63,6 +95,23 @@ private void loadStarterAssetsList() {
     }
   }
 
+  private void loadUserAssetsList() {
+    try {
+      final JSONArray jsonResponse =
+          new JSONArray(getRequest(this.generateUserAssetsListUrl(), this.httpClient));
+      this.userAssetFilenames =
+          jsonResponse
```
molly-moen, line 103: if there are no assets will this set userAssetFilenames to an empty list or throw an exception?

molly-moen (commented):
Looks good, just one behavior question

org-code-javabuilder/protocol/src/main/java/org/code/protocol/AssetFileHelper.java
sanchitmalhotra126: Good question - it should return an empty list so we should not hit a JSON exception here. I'll add a test case to validate.

molly-moen (approved):
LGTM!

PR #125, opened 2021-09-28
Playground: fix text color behavior

sanchitmalhotra126:
~~Updates TextItem to not store color directly and only maintain individual the RGB values (along with the respective setters).~~ Edit: after looking at this a little more, I changed this so TextItem stores a color object, but it is always copied into a new object when color is supplied via the constructor or setColor. This is done so that Color can still own how it deals with the individual RGB values internally (for example, the logic for sanitizing values to be [0, 255]). The setters also create a new color object based off of the supplied value and the other two existing values.

This comes from discussion around how color should behave across Playground and Theater - we eventually want color to be immutable, so we should only be able to update TextItem color via `setColor` or `setRed/Green/Blue`. `getColor` now returns a new color object constructed from the stored RGB values.

Slack thread for context: https://codedotorg.slack.com/archives/C01EF4GJ9GE/p1632255167336300

JIRA: https://codedotorg.atlassian.net/browse/CSA-856

Tested locally by verifying websocket messages & unit

### DIFF
--- a/org-code-javabuilder/playground/src/main/java/org/code/playground/TextItem.java
+++ b/org-code-javabuilder/playground/src/main/java/org/code/playground/TextItem.java
@@ -11,9 +11,7 @@ public class TextItem extends Item {
   private Font font;
   private FontStyle fontStyle;
   private double rotation;
-  private int colorRed;
-  private int colorGreen;
-  private int colorBlue;
+  private Color color;
 
   /**
    * Creates text item that can be placed the board.
@@ -38,7 +36,8 @@ public class TextItem extends Item {
       double rotation) {
     super(x, y, height);
     this.text = text;
-    this.setColorHelper(color);
+    // Copy to new color to store the values only, and not a reference to the original color object
+    this.color = new Color(color);
     this.font = font;
     this.fontStyle = fontStyle;
     this.rotation = rotation;
@@ -84,7 +83,8 @@ public class TextItem extends Item {
    * @param color the text color
    */
   public void setColor(Color color) {
-    this.setColorHelper(color);
+    // Copy to new color to store the values only, and not a reference to the original color object
+    this.color = new Color(color);
     HashMap<String, String> colorDetails = new HashMap<>();
     this.addColorToDetails(colorDetails);
     this.sendChangeMessage(colorDetails);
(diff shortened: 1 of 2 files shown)


org-code-javabuilder/playground/src/main/java/org/code/playground/TextItem.java
```diff
@@ -85,19 +83,58 @@ public String getText() {
...
+
   /**
    * Get the text color for the item.
    *
    * @return the text color for the item
    */
   public Color getColor() {
-    return this.color;
+    // TODO: wrapping this.color in a new Color object will not be necessary once Color is
+    // immutable.
```
jmkulwik, line 136: NIT: Grab a permalink to this TODO and add it to the task to make color immutable. (And similarly to other TODOs in related to making color immutable in this task). The task, for reference: https://codedotorg.atlassian.net/browse/CSA-865

org-code-javabuilder/playground/src/test/java/org/code/playground/TextItemTest.java
```diff
@@ -77,5 +84,35 @@ public void settersSendChangeMessages() {
...
         messages.get(4).getDetail().get(ClientMessageDetailKeys.ROTATION));
+    assertEquals(
+        Integer.toString(colorRed),
+        messages.get(5).getDetail().get(ClientMessageDetailKeys.COLOR_RED));
+    assertEquals(
+        Integer.toString(colorGreen),
+        messages.get(6).getDetail().get(ClientMessageDetailKeys.COLOR_GREEN));
+    assertEquals(
+        Integer.toString(colorBlue),
+        messages.get(7).getDetail().get(ClientMessageDetailKeys.COLOR_BLUE));
```
jmkulwik, line 95: This is starting to turn into a pretty big test. Consider splitting the color-related checks into their own test.

org-code-javabuilder/playground/src/main/java/org/code/playground/TextItem.java
```diff
@@ -39,7 +36,8 @@ public TextItem(
       double rotation) {
     super(x, y, height);
     this.text = text;
-    this.setColorHelper(color);
+    // Copy to new color to store the values only, and not a reference to the original color object
```
jmkulwik, line 39: Nice documentation

jmkulwik (approved):
Couple nits, but other than that, LGTM!

org-code-javabuilder/playground/src/main/java/org/code/playground/TextItem.java
sanchitmalhotra126: Oh yep sure - do we usually link to the PR itself or the permalink once it's merged?

org-code-javabuilder/playground/src/test/java/org/code/playground/TextItemTest.java
sanchitmalhotra126: Agreed - I'll go ahead and split the color checks into their own test.

org-code-javabuilder/playground/src/main/java/org/code/playground/TextItem.java
jmkulwik: I generally make a GitHub comment and link to the comment. Permalink would work too, but it's a lot of steps to remember.

PR #126, opened 2021-09-30
Remove unused neighborhood file

molly-moen:
In my tech debt investigation I found a neighborhood file we no longer use since we updated how we handle output. This PR deletes that file.


jmkulwik (approved):
Haha I remember this file. Thanks for cleaning it up!

PR #127, opened 2021-09-30
Playground: throw runtime exception if a message is attempted after end()

molly-moen:
We don't want to allow any messages to be sent from the playground after `board.end()` has been called. We previously handled this by ignoring these messages on the frontend. Instead, we want to throw a runtime exception on the backend if a message is attempted to be sent after the board has ended.

### DIFF
--- a/org-code-javabuilder/playground/src/main/java/org/code/playground/PlaygroundMessageHandler.java
+++ b/org-code-javabuilder/playground/src/main/java/org/code/playground/PlaygroundMessageHandler.java
@@ -25,8 +25,11 @@ class PlaygroundMessageHandler {
   }
 
   public void sendMessage(PlaygroundMessage message) {
-    // only send messages if playground has not ended.
-    if (!Playground.board.hasEnded()) {
+    // only send messages if playground has not ended. Otherwise throw a runtime exception.
+    if (Playground.board.hasEnded()) {
+      throw new PlaygroundRuntimeException(
+          PlaygroundExceptionKeys.INVALID_MESSAGE_PLAYGROUND_ENDED);
+    } else {
       this.outputAdapter.sendMessage(message);
     }
   }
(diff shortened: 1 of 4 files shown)


org-code-javabuilder/playground/src/main/java/org/code/playground/PlaygroundMessageHandler.java
```diff
@@ -25,6 +25,12 @@ private PlaygroundMessageHandler() {
   }
 
   public void sendMessage(PlaygroundMessage message) {
-    this.outputAdapter.sendMessage(message);
+    // only send messages if playground has not ended. Otherwise throw a runtime exception.
+    if (Playground.board.hasEnded()) {
```
bencodeorg, line 29: Tangent re: doing this for updates on items not yet added to the board -- couldn't we put a condition here that doesn't send the message if an item isn't on the board? Too hacky?

bencodeorg (approved):

org-code-javabuilder/playground/src/main/java/org/code/playground/PlaygroundMessageHandler.java
molly-moen: I think that's the only way we could handle that on the backend. We would need to expose which items are on the board to the message handler, but we already track that on the board so it wouldn't be too much extra work. It would require the message handler to do some parsing of the messages to see the id and message type.

sanchitmalhotra126: Just circling back on this - moving this to the back end definitely makes more sense! As a potential follow-up though, to make the relationship between `Board` and `PlaygroundMessageHandler` less tightly coupled, how do you feel about exposing a method on PlaygroundMessageHandler like `setMessagesEnabled(boolean enabled)` or `enableMessages()/disableMessages()` instead of reading the state directly here? `Board` could call those at the appropriate time (e.g. when `end()` is called) so that `PlaygroundMessageHandler` doesn't need to be aware of the playground state, and the state information can stay isolated to `Board`. This may also help us keep things less coupled as move more state logic back to `Board`.

PR #128, opened 2021-10-04
Playground: fast-click prevention

sanchitmalhotra126:
Javabuilder-side support for adding fast-click prevention. Javabuilder sends an UPDATE_COMPLETE message after processing the current set of updates so that Javalab can listen and accordingly re-enable click events. Also includes some minor cleanups to related classes/tests.

Related Javalab PR for adding fast-click prevention: https://github.com/code-dot-org/code-dot-org/pull/42816
JIRA: https://codedotorg.atlassian.net/browse/CSA-841

Testing: locally & unit

### DIFF
--- a/org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
+++ b/org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
@@ -173,8 +173,7 @@ public class Board {
       throw new PlaygroundException(PlaygroundExceptionKeys.PLAYGROUND_RUNNING);
     }
 
-    this.playgroundMessageHandler.sendMessage(
-        new PlaygroundMessage(PlaygroundSignalKey.RUN, new HashMap<>()));
+    this.playgroundMessageHandler.sendMessage(new PlaygroundMessage(PlaygroundSignalKey.RUN));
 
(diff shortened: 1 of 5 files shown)


org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
```diff
@@ -185,6 +184,11 @@ public void start() throws PlaygroundException {
       if (message != null) {
         this.handleClickEvent(message);
       }
+      if (this.isRunning) {
+        // Only need to send update complete if the game is still running
+        this.playgroundMessageHandler.sendMessage(
```
molly-moen, line 189: should this be sent if the message was null?

molly-moen (commented):
Looks good, just one question

org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
sanchitmalhotra126: Yeah, I was thinking the same - opted to do it here so that we always send back an update complete message even if somehow we got a blank playground message, so that click events can get re-enabled.

molly-moen (approved):

PR #129, opened 2021-10-04
Make color immutable

sanchitmalhotra126:
Makes `Color` immutable. This more closely aligns with the Java Color class and makes behavior consistent between Playground and Theater.

JIRA: https://codedotorg.atlassian.net/browse/CSA-865

Testing: local & unit

### DIFF
--- a/org-code-javabuilder/media/src/main/java/org/code/media/Color.java
+++ b/org-code-javabuilder/media/src/main/java/org/code/media/Color.java
@@ -4,10 +4,22 @@ import java.util.Map;
 
 public class Color {
 
-  private int red;
-  private int green;
-  private int blue;
-  private int alpha;
+  private final int red;
+  private final int green;
+  private final int blue;
+  private final int alpha;
+
+  public static Color copyWithRed(Color color, int redValue) {
+    return new Color(redValue, color.getGreen(), color.getBlue());
+  }
+
+  public static Color copyWithGreen(Color color, int greenValue) {
+    return new Color(color.getRed(), greenValue, color.getBlue());
+  }
+
+  public static Color copyWithBlue(Color color, int blueValue) {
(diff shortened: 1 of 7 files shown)


org-code-javabuilder/media/src/main/java/org/code/media/Color.java
```diff
@@ -24,7 +36,7 @@ public Color(String color) throws IllegalArgumentException {
     this.red = colorToCopy.getRed();
     this.green = colorToCopy.getGreen();
     this.blue = colorToCopy.getBlue();
-    this.setFullOpacity();
+    this.alpha = 255;
```
molly-moen, line 39: nit: create constant for 255

org-code-javabuilder/media/src/test/java/org/code/media/ColorTest.java
```diff
@@ -12,11 +12,38 @@ public void defaultAlphaIs255() {
...
-    assertEquals(255, c.getAlpha());
+  public void testCopyWithRedCreatesCopyWithNewRedValue() {
+    final int red = 100;
+    final int green = 150;
+    final int blue = 200;
+    final int newRed = 50;
+    final Color copied = Color.copyWithRed(new Color(red, green, blue), newRed);
+    assertEquals(newRed, copied.getRed());
+    assertEquals(green, copied.getGreen());
+    assertEquals(blue, copied.getBlue());
```
molly-moen, line 23: can you also add a check that `copied` does not equal (strict equality) the original color?

molly-moen (approved):
LGTM, just a couple nits

org-code-javabuilder/media/src/test/java/org/code/media/ColorTest.java
sanchitmalhotra126: Sure, done!

PR #130, opened 2021-10-05
Playground: Clean up message send logic

molly-moen:
Fix up the logic for sending playground messages in preparation for batching messages. The changes here make it so that items have methods to turn on/off change messages, which will be called when an item is added/removed from the board. By default items won't send change messages, so any initial changes won't send irrelevant messages to javalab. I also updated the message handler to have enable/disable message methods, so that board can turn off messages when `exit` is called. Currently we don't have a reason to turn messages back on, but I left the method there in case it's required in the future.

I renamed the exception message we throw when a message is sent after the board exits to align better with the new model. There will be a frontend change incoming to use this new message.

### DIFF
--- a/org-code-javabuilder/playground/src/main/java/org/code/playground/Item.java
+++ b/org-code-javabuilder/playground/src/main/java/org/code/playground/Item.java
@@ -95,20 +95,14 @@ public abstract class Item {
   }
 
   protected void sendChangeMessage(String key, String value) {
-    if (this.shouldSendMessages) {
-      HashMap<String, String> details = this.getIdDetails();
-      details.put(key, value);
-      this.playgroundMessageHandler.sendMessage(
-          new PlaygroundMessage(PlaygroundSignalKey.CHANGE_ITEM, details));
-    }
+    HashMap<String, String> details = this.getIdDetails();
+    details.put(key, value);
+    this.sendChangeMessageHelper(details);
   }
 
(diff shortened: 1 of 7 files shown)


org-code-javabuilder/playground/src/main/java/org/code/playground/Item.java
```diff
@@ -93,16 +95,28 @@ protected String getId() {
...
-    HashMap<String, String> details = this.getIdDetails();
-    details.put(key, value);
-    this.playgroundMessageHandler.sendMessage(
-        new PlaygroundMessage(PlaygroundSignalKey.CHANGE_ITEM, details));
+    if (this.shouldSendMessages) {
+      HashMap<String, String> details = this.getIdDetails();
+      details.put(key, value);
+      this.playgroundMessageHandler.sendMessage(
+          new PlaygroundMessage(PlaygroundSignalKey.CHANGE_ITEM, details));
+    }
```
sanchitmalhotra126, line 103: nit: minor potential cleanup - we can create a hashmap of just `key, value` and call `this.sendChangeMessage(details)` here so we only have to check `this.shouldSendMessages` in one method

sanchitmalhotra126 (approved):
Looks good, nice refactors!

org-code-javabuilder/playground/src/main/java/org/code/playground/Item.java
molly-moen: done!

PR #131, opened 2021-10-05
Theater images should rotate around the top left corner rather than center

sanchitmalhotra126:
Just need to update the transform anchor to the top left corner (0, 0) rather than the center.

JIRA: https://codedotorg.atlassian.net/browse/CSA-866

Tested locally on All The Things theater level

New rotation (around top left corner):

https://user-images.githubusercontent.com/85528507/136087553-5d56d9b9-97ed-452c-a27f-62d4f1d886ef.mov

Original rotation (around center):

https://user-images.githubusercontent.com/85528507/136087587-dae0a312-3fc2-4fb6-9358-50535af705dd.mov


jmkulwik (approved):
LGTM!

PR #132, opened 2021-10-05
Unify clean up and exit

molly-moen:
This PR fixes a small bug and prepares us for playground message batching. It includes the following changes:
- instead of sending an `EXIT` message before throwing an exception or exiting safely, instead add a `finally` block in `executeCodeBuilder` that sends an exit message in all cases (this is the backend part of the fix for an issue where we would see 2 exit messages in javalab if there was an exception)
- Add a new interface `MessageHandler` which `PlaygroundMessageHandler` implements, and register the `MessageHandler` with `GlobalProtocol`. Then, in the new `finally` block in `CodeBuilderWrapper`, call clean up the message handler(s). This will enable us to send any pending messages on exit for our new batch message strategy. We also can implement other message handlers with their own exit methods in the future if need be.

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeBuilderWrapper.java
+++ b/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeBuilderWrapper.java
@@ -61,6 +61,9 @@ public class CodeBuilderWrapper {
 
       // Additionally, these may have affected the user. For now, let's tell them about it.
       outputAdapter.sendMessage(error.getExceptionMessage());
+    } finally {
+      GlobalProtocol.getInstance().cleanUpResources();
+      this.outputAdapter.sendMessage(new StatusMessage(StatusMessageKey.EXITED));
     }
   }
 }
--- a/org-code-javabuilder/playground/src/main/java/org/code/playground/PlaygroundMessageHandler.java
+++ b/org-code-javabuilder/playground/src/main/java/org/code/playground/PlaygroundMessageHandler.java
@@ -1,9 +1,10 @@
 package org.code.playground;
 
 import org.code.protocol.GlobalProtocol;
+import org.code.protocol.MessageHandler;
 import org.code.protocol.OutputAdapter;
 
-class PlaygroundMessageHandler {
+class PlaygroundMessageHandler implements MessageHandler {
   private static PlaygroundMessageHandler instance;
   private boolean messagesEnabled;
 
@@ -18,6 +19,7 @@ class PlaygroundMessageHandler {
 
   private PlaygroundMessageHandler() {
     this(GlobalProtocol.getInstance().getOutputAdapter());
+    GlobalProtocol.getInstance().registerMessageHandler(this);
   }
 
   // Visible for testing only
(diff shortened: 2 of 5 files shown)


org-code-javabuilder/protocol/src/main/java/org/code/protocol/GlobalProtocol.java
```diff
@@ -16,6 +19,7 @@
   private final String dashboardHostname;
   private final String channelId;
   private final AssetFileHelper assetFileHelper;
+  private Set<MessageHandler> messageHandlers;
```
bencodeorg, line 22: Help me understand when we'll have multiple message handler?

molly-moen: If we ever need another one for neighborhood, theater, some other new feature in the future. This is mainly here to prevent hard coding playground-specific logic into protocol

org-code-javabuilder/playground/src/main/java/org/code/playground/PlaygroundMessageHandler.java
```diff
@@ -35,11 +37,18 @@ public void sendMessage(PlaygroundMessage message) {
...
   }
 
-  protected void enableMessages() {
+  @Override
+  public void exit() {
+    // TODO: Once batching is added, send any remaining messages here
+  }
+
+  @Override
+  public void enableMessages() {
```
bencodeorg, line 46: These need to be public now? Is it because the interface is public (if so, why does the interface need to be public?) Bear with me, just trying to get my Java know-how up :) Maybe general question, how much do you think about access modifiers when writing Java code? Are you not concerned about it sometimes (and questions like this one are too nitty in those cases)?

sanchitmalhotra126 (approved):
Looks good! I'm curious how you're thinking Neighborhood and Theater might use `MessageHandler` - are there specific cleanup steps for each, or would they also benefit from using `enable/disableMessages`?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeBuilderWrapper.java
```diff
@@ -61,6 +61,9 @@ public void executeCodeBuilder() {
 
       // Additionally, these may have affected the user. For now, let's tell them about it.
       outputAdapter.sendMessage(error.getExceptionMessage());
+    } finally {
+      GlobalProtocol.getInstance().cleanUpResources();
```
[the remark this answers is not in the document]
molly-moen: this is so that we send any messages if `start` is not called, or the program exits at an unexpected time. We want to show the initial state even if we don't call start, so we need to do this clean up once the program has exited.

molly-moen:
> Looks good! I'm curious how you're thinking Neighborhood and Theater might use `MessageHandler` - are there specific cleanup steps for each, or would they also benefit from using `enable/disableMessages`?

@sanchitmalhotra126 I don't think as of now neighborhood and theater would need `MessageHandler`, but I could see a future point where we want to do more batching or we add another mini app that needs its own handler, sorry for the unclear explanation. The main reason to add `MessageHandler` is to have an interface outside of `playground` that we can reference in `lib` and `protocol`. The reason we want it out there is so we can handle sending any pending messages if the program exits unexpectedly.

PR #133, opened 2021-10-06
Playground: batch messages

molly-moen:
This PR adds simple batching to playground. We collect playground messages in a queue, then send them all as a json blob when `sendBatchedMessages` is called. 
 We send batches of messages at the following times:
- when `start` is called
- after every click handler
- when `end` is called
- when the program exits

The blob is in this format:
```
{
  "detail": {
    "updates": [
      {
        "detail": {
          "filename": "puppy_1.jpg",
          "x": "0",
          "width": "120",
          "y": "0",
          "index": "0",
          "id": "1",
          "height": "120"
        },
        "type": "PLAYGROUND",
        "value": "ADD_CLICKABLE_ITEM"
      },
      {
        "detail": {
          "y": "200",
          "id": "1"
        },
        "type": "PLAYGROUND",
        "value": "CHANGE_ITEM"
      },
      {
        "type": "PLAYGROUND",
        "value": "RUN"
      }
    ]
  },
  "type": "PLAYGROUND",
  "value": "UPDATE"
}
```

## Potential Improvements
As a first step towards batching, no changes have been made to the individual message details or messages we send (we still send UPDATE_COMPLETE at the end of each click event, for example). There's a few ways we could optimize the batch of messages to decrease the size and limit the work javalab needs to do:
- consolidate all changes to the same item id into one update detail. For example, if an ADD_ITEM is followed by a CHANGE_ITEM, we can just do an ADD_ITEM event with the updated values. Or, if we change an item, then remove it, we can just remove it. 
- Determine if we still need UPDATE_COMPLETE messages. If we will always send the results of a click event in one blob, we could implicitly tell that an update is complete by the fact that we receive the blob.

## Deployment Strategy
There is an associated code-dot-org [PR](https://github.com/code-dot-org/code-dot-org/pull/42858) that accepts this new UPDATE event. We should deploy that change before this change is deployed to prevent interruptions in playground functionality. The code-dot-org deployment will not cause issues if it is done first, because it still accepts non-batched updates.

### DIFF
--- a/org-code-javabuilder/playground/src/main/java/org/code/playground/PlaygroundMessageHandler.java
+++ b/org-code-javabuilder/playground/src/main/java/org/code/playground/PlaygroundMessageHandler.java
@@ -49,8 +49,6 @@ class PlaygroundMessageHandler implements MessageHandler {
       return;
     }
     JSONArray messages = new JSONArray();
-    // copy existing queue so any new messages that come in during parsing
-    // will be handled in next batch
     for (PlaygroundMessage message : this.queuedMessages) {
       messages.put(new JSONObject(message.getFormattedMessage()));
     }


sanchitmalhotra126 (commented):
Nice, looks good! Just a couple behavior questions

org-code-javabuilder/playground/src/main/java/org/code/playground/PlaygroundMessageHandler.java
```diff
@@ -26,20 +33,42 @@ private PlaygroundMessageHandler() {
...
-      this.outputAdapter.sendMessage(message);
+      this.queuedMessages.add(message);
     }
   }
 
+  public void sendBatchedMessages() {
+    if (this.queuedMessages.isEmpty() || this.sendInProgress) {
+      return;
+    }
+    this.sendInProgress = true;
```
jmkulwik, line 53: Can `sendBatchedMessages` be called while `sendInProgress` is true? I think we run student code in a single thread and querying for input is a blocking operation, right?

org-code-javabuilder/playground/src/main/java/org/code/playground/Board.java
```diff
@@ -15,7 +15,6 @@
 
   private boolean firstRunStarted;
   private boolean isRunning;
-  private boolean hasEnded;
```
molly-moen, line 18: I removed the use of this in [this PR](https://github.com/code-dot-org/javabuilder/pull/130/files), cleaning it up now

PR #135, opened 2021-10-06
Playground: split up large batch messages

molly-moen:
Websocket messages can be at most 128kb, which is around 131,000 characters. It is very unlikely a batch of playground messages would be more than 131,00 characters, but in the rare case the batch gets too large, we want to split it up into multiple messages. This doesn't cover the scenario where a single message is greater than 128 kb, but this seems extremely unlikely and would indicate bad user behavior.


org-code-javabuilder/playground/src/main/java/org/code/playground/PlaygroundMessageHandler.java
```diff
@@ -13,6 +13,9 @@
   private static PlaygroundMessageHandler instance;
   private boolean messagesEnabled;
   private final Queue<PlaygroundMessage> queuedMessages;
+  // A websocket request can be up to 128 kb, which is around 131,000 characters.
+  // Make sure we don't get close to this limit.
+  private final int MAX_CHARACTERS_PER_MESSAGE = 120000;
```
sanchitmalhotra126, line 18: nit: static?

jmkulwik (approved):
Nice improvement

PR #136, opened 2021-10-11
Javabuilder: add run tests option

sanchitmalhotra126:
Adds an option to run tests on Javabuilder (as opposed to running the main method, which is currently the default behavior). This is done by introducing an "execution type" parameter that is passed from Javalab to Javabuilder. Currently, the types include `RUN` and `TEST`, but my thinking was this can expanded to include other uses (such as `COMPILE_ONLY`) and give us more flexibility in how we want to handle requests. Currently invoking Javabuilder with `TEST` will do nothing (running tests no-ops); actually finding and running test code will be a separate change.

Additionally - if no execution type is provided, we will temporarily default to `RUN` to make sure Javalab doesn't break. There will be a separate change to Javalab for sending the parameter, after which there will be a follow-up task to remove the fallback.

JIRA: https://codedotorg.atlassian.net/browse/CSA-930
Related Javalab PR for passing the execution type parameter: https://github.com/code-dot-org/code-dot-org/pull/42938

Testing: locally with All The Things Level

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaRunner.java
+++ b/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaRunner.java
@@ -30,7 +30,7 @@ public class JavaRunner {
    * @throws InternalFacingException When we hit an internal error after the user's code has
    *     finished executing.
    */
-  public void runCode() throws InternalFacingException, JavabuilderException {
+  public void runMain() throws InternalFacingException, JavabuilderException {
     // Include the user-facing api jars in the code we are loading so student code can access them.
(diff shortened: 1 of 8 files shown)


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaRunner.java
```diff
@@ -72,16 +72,19 @@ public void runCode() throws InternalFacingException, JavabuilderException {
     }
   }
 
+  public void runTests() {
+    // TODO: Find and run tests
```
sanchitmalhotra126, line 76: https://codedotorg.atlassian.net/browse/CSA-929

PR #138, opened 2021-10-15
Revert from Ruby 2.7.4 to 2.7.2

sureshc:
To enable building Javabuilder in AWS CodeBuild, revert from Ruby 2.7.4 to 2.7.2.

To use an [AWS CodeBuild Amazon Linux 2 container](https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-available.html) to build Javabuilder, we need to use the [same specific version](https://github.com/aws/aws-codebuild-docker-images/blob/a07b8a37906d4de35cf0c1aa6fa94dc1151114e6/al2/x86_64/standard/3.0/Dockerfile#L448-L450) of Ruby included in the CodeBuild container.

All developers need to install Ruby 2.7.2 on their local development environment via rbenv to be able to build and deploy Javabuilder from their development environment after this change is merged.:
`rbenv install 2.7.2`


molly-moen (approved):
Thanks!

PR #139, opened 2021-10-18
Javabuilder: add new compile code request type

sanchitmalhotra126:
Adds a new execution type (COMPILE_ONLY) to Javabuilder that instructs Javabuilder to only compile code. As per the backpack design, this requires a list of files to compile to be provided; if no list is provided, or there are no files matching the list, then Javabuilder throws a UserInitiatedException.

This also includes the work to remove the execution type fallback introduced earlier, now that Javalab is sending the execution type param to Javabuilder.

Tested locally on All The Things level + unit

JIRA: https://codedotorg.atlassian.net/browse/CSA-518
JIRA: https://codedotorg.atlassian.net/browse/CSA-950 [remove fallback code]

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketServer.java
+++ b/org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketServer.java
@@ -4,12 +4,10 @@ import static org.code.protocol.LoggerNames.MAIN_LOGGER;
 
 import java.io.IOException;
 import java.nio.ByteBuffer;
-import java.util.ArrayList;
 import java.util.Base64;
 import java.util.List;
 import java.util.logging.Handler;
 import java.util.logging.Logger;
-import java.util.stream.Collectors;
 import javax.websocket.OnClose;
 import javax.websocket.OnMessage;
 import javax.websocket.OnOpen;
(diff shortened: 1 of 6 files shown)


org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketServer.java
```diff
@@ -50,22 +53,27 @@ public void onOpen(Session session) {
...
+    if (options.has("compileList")) {
+      compileList =
+          options
+              .getJSONArray("compileList")
+              .toList()
+              .stream()
+              .map(filename -> (String) filename)
+              .collect(Collectors.toList());
+    } else {
+      compileList = new ArrayList<>();
```
molly-moen, line 75: it feels weird to send an empty list for running code when we actually want to compile everything. Would it make more sense to send `null` instead?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
```diff
@@ -55,14 +59,18 @@ public String handleRequest(Map<String, String> lambdaInput, Context context) {
...
-    final String executionTypeString = lambdaInput.get("executionType");
-    // TODO: in order to not break current behavior in Javalab, execution type defaults to 'RUN'.
-    // Once Javalab is sending the execution type parameter, remove this fallback
-    final ExecutionType executionType =
-        executionTypeString == null
-            ? ExecutionType.RUN
-            : ExecutionType.valueOf(executionTypeString);
+    final List<String> compileList;
+    if (options.has("compileList")) {
+      compileList =
```
molly-moen, line 64: could we simplify this logic by sending the compile list in a different format? Also it might be worth creating a helper method for this code since it's duplicated across the websocket server and the lambda request handler.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserProjectFiles.java
```diff
@@ -28,4 +29,20 @@ public void addTextFile(TextProjectFile textFile) {
   public List<TextProjectFile> getTextFiles() {
     return this.textFiles;
   }
+
+  public List<JavaProjectFile> getMatchingJavaFiles(List<String> filenames) {
+    return this.getMatchingProjectFiles(filenames, this.javaFiles);
+  }
+
+  public List<TextProjectFile> getMatchingTextFiles(List<String> filenames) {
```
molly-moen, line 37: do we need this method?

molly-moen (commented):
Looks good! Just a couple comments

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
sanchitmalhotra126: Hmm yeah, the logic here is just to convert the JSONArray into a list of Strings - is there a format that would be more friendly for conversion? Agreed on duplicated code - I'll pull this out into a helper.

org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketServer.java
molly-moen: I'd lean towards passing null and handling it on the method side, because it's a little clearer.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
molly-moen: yeah this might be the simplest option then

PR #140, opened 2021-10-18
Stop lambda if connection is lost

molly-moen:
Previously, if end user code was stopped via the Javalab 'stop' button, the user would not see any more code execution, but the lambda would continue executing code until it either finished or it timed out after 5 minutes. When stop is clicked in Javalab, we do close the websocket. Our websocket clean up also deletes the sqs queue created for code execution. We can utilize this clean up to determine if we should end code execution in the lambda as well, then stop the lambda. Here's how it works:
- Update our end user code execution to occur in a separate thread.
- Every 500 ms while end user code is running, check if the output or input adapter still has an active connection (the output adapter connects to the websocket, the input adapter to sqs). If either connection has been lost, we can safely end code execution, as we would not be able to communicate input or output, and it most likely means the user chose to end their program.
- If we discover that the input or output adapter cannot connect during code execution (for example, we try to receive a message and cannot) we mark the connection as inactive to catch in the next check by the `LambdaRequestHandler`. Previously we would throw an exception.
- We now ignore specific exceptions (mainly `InterruptedException`) that are caused by the thread being interrupted.


org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketOutputAdapter.java
```diff
@@ -22,14 +22,23 @@ public void sendMessage(ClientMessage message) {
...
+    return true;
+  }
+
   public void sendDebuggingMessage(ClientMessage message) {
     try {
       endpoint.sendText(message.getFormattedMessage());
     } catch (IOException e) {
       e.printStackTrace();
+    } catch (IllegalStateException e) {
+      // this happens if the endpoint has been closed. Fail silently.
```
sanchitmalhotra126, line 41: Do we need to set active connection to false here?

org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketServer.java
```diff
@@ -85,25 +86,37 @@ public void onOpen(Session session) {
...
             () -> {
-              CodeBuilderWrapper codeBuilderWrapper =
-                  new CodeBuilderWrapper(fileLoader, outputAdapter);
-              codeBuilderWrapper.executeCodeBuilder(executionType);
+              try {
+                this.codeExecutor.join();
+              } catch (InterruptedException e) {
+                // ignore this exception, it likely means the end user stopped
+                // their program.
+              }
```
sanchitmalhotra126, line 101: Just to make sure I understand - for the WebSocketServer case, we don't need to manually check if the connection is alive periodically because closing the websocket will automatically trigger an InterruptedException, right?

sanchitmalhotra126 (commented):
Looks good! One general question/idea - currently, the stop button simply closes the websocket connection. Do you think it would be worth Javalab sending back an explicit STOP message and then closing the connection? It might help us differentiate between intentional interruptions and unintended interruptions, in case that's useful information - but also may help make some of the logic here clearer - for example, I'd imagine we'd still need to periodically check if the input/output adapters are connected in order to determine if the connection has been suspended, but we could also safely stop the lambda on receiving a STOP message without needing to check.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
```diff
@@ -118,9 +123,27 @@ public String handleRequest(Map<String, String> lambdaInput, Context context) {
...
+      codeBuilderThread.start();
+      while (codeBuilderThread.isAlive()) {
+        // sleep for CHECK_THREAD_INTERVAL_MS, then check if we've lost the connection to the
+        // input or output adapter. This means we have lost connection to the end user (either
+        // because
+        // they terminated their program or some other issue), and we should stop executing their
+        // code.
+        Thread.sleep(CHECK_THREAD_INTERVAL_MS);
+        if (codeBuilderThread.isAlive()
+            && (!inputAdapter.hasActiveConnection() || !outputAdapter.hasActiveConnection())) {
```
jmkulwik, line 139: What would be a case where the inputAdapter would be inactive but the outputAdapter would be active? Is it worth always making both requests, or is it sufficient to ping only one service?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/InputRedirectionStream.java
```diff
@@ -36,8 +36,14 @@ public int read() {
       // The Java Lab console is an <input> element that uses the enter key to trigger onSubmit.
       // Rather than adding an arbitrary line separator from the client, we instead add the
       // separator here so we can use a line separator that Scanner will recognize.
-      final String messageWithNewline =
-          inputAdapter.getNextMessageForType(InputMessageType.SYSTEM_IN) + System.lineSeparator();
+      final String stringMessage = inputAdapter.getNextMessageForType(InputMessageType.SYSTEM_IN);
+      // a null message means we've lost connection to the input adapter and won't receive any more
```
jmkulwik, line 40: (for my own learning) What was causing this to happen?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
molly-moen: If we had some sort of aws issue where sqs was not available or the websocket wasn't available only one could be down.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
```diff
@@ -132,6 +155,15 @@ private void cleanUpResources(String connectionId, AmazonApiGatewayManagementApi
         new DeleteConnectionRequest().withConnectionId(connectionId);
     // Deleting the API Gateway connection should always be the last thing executed because the
     // delete action cleans up the AWS resources associated with this lambda
-    api.deleteConnection(deleteConnectionRequest);
+    try {
+      api.deleteConnection(deleteConnectionRequest);
+    } catch (GoneException e) {
```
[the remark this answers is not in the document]
molly-moen: yep, probably. If the student exited their browser or clicked stop before their program exited normally we would get a gone exception

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
```diff
@@ -118,9 +123,27 @@ public String handleRequest(Map<String, String> lambdaInput, Context context) {
...
+      CodeBuilderRunnable codeBuilderRunnable =
+          new CodeBuilderRunnable(userProjectFileLoader, outputAdapter, executionType, compileList);
+      Thread codeBuilderThread = new Thread(codeBuilderRunnable);
+      // start code build and execute in a thread
+      codeBuilderThread.start();
+      while (codeBuilderThread.isAlive()) {
+        // sleep for CHECK_THREAD_INTERVAL_MS, then check if we've lost the connection to the
+        // input or output adapter. This means we have lost connection to the end user (either
+        // because
+        // they terminated their program or some other issue), and we should stop executing their
```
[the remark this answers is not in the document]
molly-moen: done!

org-code-javabuilder/protocol/src/main/java/org/code/protocol/InputAdapter.java
```diff
@@ -3,4 +3,6 @@
 public interface InputAdapter {
   /** @return The next user input to the currently running program */
   String getNextMessage();
+
+  boolean hasActiveConnection();
```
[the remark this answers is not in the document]
molly-moen: done!

jmkulwik (approved):

PR #141, opened 2021-10-18
Theater: add blank Prompter

bencodeorg:
Adds blank Prompter class for use in Theater. Expected usage is something like `Theater.prompter.getImage('some string')`. More detail [here](https://docs.google.com/document/d/14t7lE6PiU4ffadCRteQyL754l8WxNqcz_31bqUIOXNc/edit#bookmark=id.1tsh653fql9c).

### DIFF
--- a/org-code-javabuilder/theater/src/main/java/org/code/theater/Prompter.java
+++ b/org-code-javabuilder/theater/src/main/java/org/code/theater/Prompter.java
@@ -1,0 +1,8 @@
+package org.code.theater;
+
+public class Prompter {
+  protected Prompter() {}
+
+  // When implemented, should ultimately return an Image object.
+  public void getImage() {}
+}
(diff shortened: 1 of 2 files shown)


org-code-javabuilder/theater/src/main/java/org/code/theater/Prompter.java
```diff
@@ -0,0 +1,8 @@
+package org.code.theater;
+
+public class Prompter {
+  protected Prompter() {}
+
+  // When implemented, should ultimately return an Image object.
+  public void getImage() {}
```
sanchitmalhotra126, line 7: Do you want to go ahead and have the method signature be the expected (`Image getImage(String)`) but just return `null`? That's typically how we've implemented no-op APIs in the past.

bencodeorg: Oops, hold on a sec. Looking back at spec, I may be misunderstanding what this method is supposed to do :).

PR #142, opened 2021-10-19
Prompter: send message when getImage is called

bencodeorg:
Sends a message ("GET_IMAGE") to Javalab when getImage is called, and includes the prompt that should be displayed in the `details`.

## Testing story

Added unit test, and tested manually that I saw these messages appearing when I called Theater.prompter.getImage("") from Javalab.

### DIFF
--- a/org-code-javabuilder/playground/build.gradle
+++ b/org-code-javabuilder/playground/build.gradle
@@ -16,7 +16,7 @@ java {
 
 dependencies {
     testImplementation 'org.junit.jupiter:junit-jupiter-api:5.6.0'
-    // https://mvnrepository.com/artifact/org.mockito/mockito-core
+    // https://mvnrepository.com/artifact/org.mockito/mockito-inline
     testImplementation group: 'org.mockito', name: 'mockito-inline', version: '3.9.0'
     testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.6.0'
     // https://mvnrepository.com/artifact/org.json/json
--- a/org-code-javabuilder/theater/src/main/java/org/code/theater/Prompter.java
+++ b/org-code-javabuilder/theater/src/main/java/org/code/theater/Prompter.java
@@ -1,11 +1,22 @@
 package org.code.theater;
 
+import java.util.HashMap;
 import org.code.media.Image;
+import org.code.protocol.GlobalProtocol;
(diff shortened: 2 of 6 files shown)


org-code-javabuilder/playground/build.gradle
```diff
@@ -16,7 +16,7 @@ java {
 
 dependencies {
     testImplementation 'org.junit.jupiter:junit-jupiter-api:5.6.0'
-    // https://mvnrepository.com/artifact/org.mockito/mockito-core
+    // https://mvnrepository.com/artifact/org.mockito/mockito-inline
```
sanchitmalhotra126, line 19: Thanks for updating these 👍

org-code-javabuilder/theater/src/main/java/org/code/theater/Prompter.java
```diff
@@ -1,11 +1,22 @@
...
+import org.code.protocol.GlobalProtocol;
+import org.code.protocol.OutputAdapter;
 
 public class Prompter {
-  protected Prompter() {}
+  private final OutputAdapter outputAdapter;
+
+  protected Prompter() {
+    this.outputAdapter = GlobalProtocol.getInstance().getOutputAdapter();
+  }
```
sanchitmalhotra126, line 13: nit: might want to have a second constructor that takes in `outputAdapter` as a parameter and call that here, for testing purposes (similar to `Board` and `Stage`) so you don't have to mock the global protocol instance

org-code-javabuilder/theater/src/main/java/org/code/theater/Prompter.java
```diff
@@ -1,11 +1,22 @@
...
 
   public Image getImage(String prompt) {
+    HashMap<String, String> getImageDetails = new HashMap<>();
+    getImageDetails.put("prompt", prompt);
```
sanchitmalhotra126, line 17: Can you add `"prompt"` to `ClientMessageDetailKeys` and reference here?

sanchitmalhotra126 (commented):
Couple small things, otherwise looks good!

org-code-javabuilder/theater/src/main/java/org/code/theater/Prompter.java
bencodeorg: That does make things a lot cleaner, thanks

PR #146, opened 2021-11-03
Warn prior to and preempt lambda timeout

bencodeorg:
We observed some weirdness occurring when a student project reached our build and run lambda timeout limit (5 minutes) -- more details in ticket below.

https://codedotorg.atlassian.net/browse/CSA-1004

Solution suggested by our infra team is to disconnect our build and run lambda before it times out, rather than try and figure out exactly why this behavior happens when a timeout occurs. The fix in this PR is to actively monitor how much time is remaining before hitting the timeout (checking every half second, using some logic that Molly added recently), send a warning status message when we're 15 seconds from our shutdown (25 seconds from the timeout limit), and at 10 seconds before timeout initiate closing the connection. 10 seconds is probably overkill, but I couldn't think of a great way to get a sense of how long closing the connection actually takes.

## Testing story

Tested via Postman, and saw a timeout warning message and a disconnect occur (no disconnect occurs currently after the lambda times out):

image

## Follow up

We should display this message as a warning to students that their program will terminate in 15 seconds -- I am less clear though on what should happen when a websocket connection closes (it looks like we just console log a message at the moment):

https://github.com/code-dot-org/code-dot-org/blob/adeb392d76ea6b1736f970723e42a5fa1be22f5b/apps/src/javalab/JavabuilderConnection.js#L170

Maybe we should always log to the student console something when the connection closes?


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
```diff
@@ -128,7 +130,10 @@ public String handleRequest(Map<String, String> lambdaInput, Context context) {
       Thread codeBuilderThread = new Thread(codeBuilderRunnable);
       // start code build and execute in a thread
       codeBuilderThread.start();
+
+      boolean timeoutWarningSent = false;
       while (codeBuilderThread.isAlive()) {
+
```
molly-moen, line 136: nit: remove extra line here?

molly-moen (approved):
LGTM! We should have a follow-up discussion on what to do if the websocket closes. Most of the time its due to normal operation and we wouldn't want a message, but it would be nice to have a message of some sort during abnormal operation.

molly-moen (approved):

PR #148, opened 2021-11-10
Allow Javabuilder output bucket to accept PUT requests

sanchitmalhotra126:
This will be needed for Javalab to upload images using signed URLs.

Tested on dev instance lambda + postman.

### DIFF
--- a/template.yml
+++ b/template.yml
@@ -292,7 +292,7 @@ Resources:
       BucketName: !If [IsDevCondition, !Sub "cdo-dev-${SubDomainName}-output", !Sub "cdo-${SubDomainName}-output"] 
       CorsConfiguration:
         CorsRules:
-          - AllowedMethods: [GET]
+          - AllowedMethods: [GET, PUT]
             AllowedOrigins: ['*']
             AllowedHeaders: ['*']
       BucketEncryption:


template.yml
```diff
@@ -292,7 +292,7 @@ Resources:
       BucketName: !If [IsDevCondition, !Sub "cdo-dev-${SubDomainName}-output", !Sub "cdo-${SubDomainName}-output"] 
       CorsConfiguration:
         CorsRules:
-          - AllowedMethods: [GET]
+          - AllowedMethods: [GET, PUT]
```
molly-moen, line 295: does this allow anyone to put or is there auth on this?

sanchitmalhotra126: As far as I understand, this should require the same level of permissions that are required for a GET request. Testing this in postman, if I try to make a PUT request to the location without any pre-signed info (just https://javabuilder-sanchit.dev-code.org/sessionId/file) I get an access denied, but with a pre-signed URL, I can upload.

molly-moen (approved):
LGTM, but I'll defer to infra on template details :)

template.yml
cat5inthecradle: Yep, this just tells the browser it's okay to attempt the request, but server side auth will still apply. https://docs.aws.amazon.com/AmazonS3/latest/userguide/cors.html

cat5inthecradle (approved):

sureshc (approved):

PR #149, opened 2021-11-10
Add functionality to Javabuilder to get a signed S3 URL for the image-upload bucket

sanchitmalhotra126:
This adds the functionality to Javabuilder to generate a signed upload URL and pass it back to Javalab as part of the GET_IMAGE message. Following this, Javalab would upload the image using the signed URL, and send back a message indicating that the upload is complete, after which Javabuilder would fetch the uploaded image.

I added this to `AWSFileWriter` since it shared a lot of the same components, and therefore renamed the class to `AWSFileManager` and the associated interface to `JavabuilderFileManager`. For that reason, a lot of the diff here is just in renaming - the relevant changes mostly reside in `Prompter` and `AWSFileManager`, and associated tests.

JIRA: https://codedotorg.atlassian.net/browse/CSA-1051

Testing: tested via deployed dev instance + postman. Confirmed that the upload URL is sent back with the GET_IMAGE payload, and, using postman, I could upload a file using the given URL.


org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalFileManager.java
```diff
@@ -27,4 +27,10 @@ public String writeToFile(String filename, byte[] inputBytes, String contentType
     }
     return String.format("http://localhost:8080/%s/%s", DIRECTORY, filename);
   }
+
+  @Override
+  public String getUploadUrl(String filename) throws JavabuilderException {
+    // TODO: Return upload URL for localhost uploads
```
sanchitmalhotra126, line 33: https://codedotorg.atlassian.net/browse/CSA-1058

org-code-javabuilder/theater/src/main/java/org/code/theater/Prompter.java
```diff
@@ -1,28 +1,43 @@
...
 import java.util.HashMap;
+import java.util.concurrent.atomic.AtomicInteger;
 import org.code.media.Image;
-import org.code.protocol.ClientMessageDetailKeys;
-import org.code.protocol.GlobalProtocol;
-import org.code.protocol.OutputAdapter;
+import org.code.protocol.*;
 
 public class Prompter {
+  private static final String FILE_NAME = "prompterImage-";
```
molly-moen, line 9: nit: `FILE_NAME_PREFIX` instead?

org-code-javabuilder/theater/src/main/java/org/code/theater/Prompter.java
```diff
@@ -1,28 +1,43 @@
...
+  private static final AtomicInteger FILE_INDEX = new AtomicInteger(0);
```
molly-moen, line 10: for my information: why do we need `AtomicInteger` over a regular counter? Is it for concurrency purposes?

molly-moen (approved):
LGTM! Just one nit and one question

org-code-javabuilder/theater/src/main/java/org/code/theater/Prompter.java
sanchitmalhotra126: Yeah, just an extra layer of concurrency protection. Practically speaking, this should all be single threaded anyway, but this just guarantees we don't have multiple files with the same name.

PR #150, opened 2021-11-12
Add functionality to Javabuilder to get the uploaded prompter image

sanchitmalhotra126:
This is the final piece of prompter uploading functionality - Javabuilder waits for a status message from Javalab, and then downloads the image and returns an Image object, or throws an exception accordingly. Note: this should only be merged after the related [Javalab change](https://github.com/code-dot-org/code-dot-org/pull/43535) has been deployed to production, so that Javalab actually sends the status messages.

JIRA (this change ended up being a little different than the JIRA description due to how the work broke down): https://codedotorg.atlassian.net/browse/CSA-1052
Javalab PR: https://github.com/code-dot-org/code-dot-org/pull/43535

Testing: tested via deployed dev instance and postman, in conjunction with localhost Javalab. After providing the upload URL to Javalab manually, I copied the outputted websocket messages to postman and confirmed that Javabuilder did consume the messages properly and take the correct action.

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalFileManager.java
+++ b/org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalFileManager.java
@@ -4,6 +4,7 @@ import static dev.javabuilder.LocalWebserverConstants.DIRECTORY;
 
 import java.io.File;
 import java.io.IOException;
+import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import org.code.javabuilder.InternalServerError;
@@ -31,6 +32,12 @@ public class LocalFileManager implements JavabuilderFileManager {
   @Override
   public String getUploadUrl(String filename) throws JavabuilderException {
     // TODO: Return upload URL for localhost uploads
+    return null;
+  }
+
+  @Override
+  public URL getFileUrl(String filename) {
+    // TODO: Return file URL for locally uploaded files
     return null;
   }
 }
--- a/org-code-javabuilder/media/src/main/java/org/code/media/Image.java
+++ b/org-code-javabuilder/media/src/main/java/org/code/media/Image.java
@@ -67,6 +67,19 @@ public class Image {
   }
 
   /**
+   * Creates a new Image from the given BufferedImage. Only meant for internal use.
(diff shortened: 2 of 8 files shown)


org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalFileManager.java
```diff
@@ -33,4 +34,10 @@ public String getUploadUrl(String filename) throws JavabuilderException {
     // TODO: Return upload URL for localhost uploads
     return null;
   }
+
+  @Override
+  public URL getFileUrl(String filename) {
+    // TODO: Return file URL for locally uploaded files
```
sanchitmalhotra126, line 40: https://codedotorg.atlassian.net/browse/CSA-1058

org-code-javabuilder/media/src/main/java/org/code/media/Image.java
```diff
@@ -151,6 +164,17 @@ public BufferedImage getBufferedImage() {
...
   }
 
+  /**
+   * Load the given image from a URL and return an Image object
+   *
+   * @param url to load the file from
+   * @return Image object
+   * @throws FileNotFoundException if the file is not found at the URL
+   */
+  public static Image fromUrl(URL url) throws FileNotFoundException {
```
molly-moen, line 174: Is there a way for us to protect this function better? I am slightly concerned about students being able to get images from arbitrary urls, although with import blocking I think this concern may go away (since they won't be able to access `URL`).

sanchitmalhotra126: Yeah agreed - I added this static method in an effort to protect it a little (as opposed to adding an `Image` constructor that takes a URL) but it is still fully public and accessible. I think for now I can add a different static method `fromMediaFile` or something that will internally fetch the uploaded image.

Looking at this though, I'm also thinking in the long run we could refactor how we access media content in general, especially as we consider different ways of storing Javabuilder content as per our infra discussion - maybe some `JavabuilderContentManager` that manages all the content Javabuilder needs to run (media, source code, etc).

molly-moen (approved):
LGTM!

PR #151, opened 2021-11-17
Enable prompter on localhost

sanchitmalhotra126:
This enables localhost file uploads in order to use Prompter on localhost. Specifically, this adds a `doPut` route to our existing localhost `HttpFileServer` as well as an `HttpFilter` that's necessary for adding CORS headers. 
One callout is that I changed the url patterns that our local file server matches from theater specific URLs to any URL at `localhost:8080/javabuilderfiles/*`. This is because we can't use wildcards to match partial url paths (can't do something like `prompterImage-*`) and all prompter images are suffixed with an index. We could in theory enumerate all prompter paths (current limit is 20) but we'd have to update this any time we change the limit, so I figured this would be simpler. The doPut and doGet methods also have internal validation and will reject requests with a 403 error if path is invalid (if it's not a GET request for a theater or prompter file, or if it's not a PUT request for a prompter file).

JIRA: https://codedotorg.atlassian.net/browse/CSA-1058

Tested locally with All the Things

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/dev/javabuilder/CORSFilter.java
+++ b/org-code-javabuilder/lib/src/main/java/dev/javabuilder/CORSFilter.java
@@ -1,0 +1,53 @@
+package dev.javabuilder;
+
+import static dev.javabuilder.LocalWebserverConstants.DIRECTORY;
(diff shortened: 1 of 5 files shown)


molly-moen (approved):
LGTM!

PR #152, opened 2021-11-17
Custom class loader for user-provided code

molly-moen:
This PR adds a custom class loader with which we load any user-provided code. This class loader allows us to lock down which packages and classes users can access within their code. The version of the class loader in this PR runs in silent mode; it only logs to our backend when it would block a class, instead of throwing an exception. This is so we can ensure we are not missing any valid classes that students are already using. It also allows us to get this first step in while we work on the frontend messaging.

In order to make log parsing easier, I've slightly updated how we log our custom messages to cloud watch. Previously we sent stringified, escaped json, which isn't handled nicely. Now, if we are sending json, we will send it as-is, so we can parse the logs easier.

As an example, previously a message would look like this:
```
{
    "level": "WARNING",
    "sessionMetadata": {
        "levelId": "672",
        "connectionId": "abc",
        "sessionId": "123",
        "channelId": "xyz"
    },
    "message": "{\"className\":\"java.io.FileWriter\",\"type\":\"invalidClass\"}"
}
```

Now it looks like this:
```
{
    "level": "WARNING",
    "sessionMetadata": {
        "levelId": "672",
        "connectionId": "abc",
        "sessionId": "123",
        "channelId": "xyz"
    },
    "message": {
        "className": "java.io.FileWriter",
        "type": "invalidClass"
    }
}
```

## Deployment Strategy
This PR can be deployed any time, as it does not change the user experience. After this, we should add the correct error handling in Javalab for this new error message. Once that is deployed (which will likely be post-hour of code), and we have verified that we have not missed any valid classes, we can turn this "on" by throwing an exception in the class loader for invalid classes.


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserClassLoader.java
```diff
@@ -0,0 +1,95 @@
...
+
+    // Log that we are going to throw an exception
+    JSONObject eventData = new JSONObject();
+    eventData.put("type", "invalidClass");
+    eventData.put("className", name);
+    Logger.getLogger(MAIN_LOGGER).warning(eventData.toString());
+    // For now, don't throw an exception, and instead go on to the approved class loader.
+    // We want to ensure we aren't blocking anything needed by users, so we are running this in
+    // silent mode.
+    // throw new ClassNotFoundException(name);
```
molly-moen, line 55: task [here](https://codedotorg.atlassian.net/browse/CSA-1076) for actually throwing an exception

breville:
Nice incremental strategy to manage risk!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserClassLoader.java
```diff
@@ -0,0 +1,95 @@
...
+    // Otherwise, throw an exception.
+    if (this.allowedClasses.contains(name)) {
+      return this.approvedClassLoader.loadClass(name);
+    }
+    // allow .<specific-class> usage from valid packages
+    for (int i = 0; i < this.allowedPackages.length; i++) {
+      if (name.startsWith(this.allowedPackages[i])) {
+        return this.approvedClassLoader.loadClass(name);
+      }
+    }
```
sanchitmalhotra126, line 45: Does this have any noticeable performance impact? It's a pretty small list so I wouldn't think so, but since this check happens for every non-user class I wasn't sure.

sanchitmalhotra126: Also should we be checking for the `.` explicitly here? It's probably unlikely, but could we end up matching some class in a different package like hypothetically `java.utils.SomeClass` since it starts with `java.util`?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaRunner.java
```diff
@@ -62,4 +70,16 @@ private void run(CodeRunner runner) throws JavabuilderException, InternalFacingE
...
+   * @param javaFiles List of java files to parse
+   * @return The class names of the given java files, as a list of Strings.
+   */
+  private List<String> parseClassNames(List<JavaProjectFile> javaFiles) {
+    List<String> classNames = new ArrayList<>();
+    for (JavaProjectFile file : javaFiles) {
+      classNames.add(file.getClassName());
+    }
+    return classNames;
+  }
```
sanchitmalhotra126, line 84: nit: could use `stream()` here

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/MainRunner.java
```diff
@@ -46,6 +46,18 @@ public void run(URLClassLoader urlClassLoader) throws JavabuilderException {
       if (e.getCause() instanceof FileNotFoundException) {
         throw new UserInitiatedException(UserInitiatedExceptionKey.FILE_NOT_FOUND, e.getCause());
       }
+      // NoClassDefFoundError is thrown by the class loader if the user attempts to use a disallowed
+      // class.
+      if (e.getCause() instanceof NoClassDefFoundError) {
```
sanchitmalhotra126, line 51: Should we handle this for tests too?

sanchitmalhotra126 (commented):
Nice! This looks really promising. A couple of questions but nothing major.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserClassLoader.java
molly-moen: I've updated this to have `.` at the end of the approved packages, and I changed the individual class list to a set for faster lookup.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/MainRunner.java
molly-moen: This is trickier than I thought, starting a thread [here](https://codedotorg.slack.com/archives/C01EF4GJ9GE/p1637281178354400). May warrant a separate PR.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaRunner.java
molly-moen: done!

sanchitmalhotra126 (approved):
LGTM! I'm good with shipping this and handling tests being a separate PR.

PR #153, opened 2021-11-19
Show invalid class message for tests

molly-moen:
This PR properly handles the upcoming invalid class exception thrown by the custom class loader when running tests (currently merged in silent mode). The exception will be thrown as a `NoClassDefFoundError`, but we want to surface it to the end user as a `ClassNotFoundException`, with a message that the class is not supported. The new error will look like this on the javalab side:
Screen Shot 2021-11-19 at 1 47 40 PM

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavabuilderTestExecutionListener.java
+++ b/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavabuilderTestExecutionListener.java
@@ -117,15 +117,21 @@ public class JavabuilderTestExecutionListener extends SummaryGeneratingListener 
    */
   private String getErrorMessage(Throwable throwable, String className) {
     final String errorLine = this.errorLine(throwable.getStackTrace(), className);
-    final String errorMessage;
+    String errorMessage;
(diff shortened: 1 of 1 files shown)


bencodeorg (approved):

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavabuilderTestExecutionListener.java
```diff
@@ -147,4 +151,28 @@ private String errorLine(StackTraceElement[] stackTrace, String className) {
     }
     return null;
   }
+
+  private String getThrowableDisplayName(Throwable throwable) {
+    // A NoClassDefFoundError occurs when a user uses a disallowed class. We should
+    // instead show a ClassNotFoundException.
+    if (throwable instanceof NoClassDefFoundError) {
+      return "ClassNotFoundException";
```
sanchitmalhotra126, line 159: nit: `ClassNotFoundException.class.getSimpleName()`?

sanchitmalhotra126 (approved):
LGTM

PR #155, opened 2021-12-01
Create a different lambda configuration for each mini-app type

jmkulwik:
This creates a specialized lambda configuration for each mini-app type that is tuned to the timeout and RAM needs for the mini-app. 

Specifically, it sets the following:
```
Mini-app     | RAM    | Timeout
-------------------------------
Theater      | 1769MB | 90s
Playground   | 512MB  | 300s (can be increased to 900s)
Neighborhood | 512MB  | 60s
Console      | 512MB  | 60s
```
_^ Note: Theater is currently set at 300s timeout since all mini-app types are currently using Theater and will continue to until after HOC week. Once the other app types have been transitioned, the Theater timeout will be decreased._

This PR is paired with a PR from the code-dot-org repo: https://github.com/code-dot-org/code-dot-org/pull/43871, but they can be merged in any order.

Details on how the mini-apps were tuned to these levels can be found here: https://docs.google.com/document/d/1cWpdw5vJkmgdcw4yLg1EO4vM5N3a8M4AtzroPK52j9E/edit#


molly-moen (approved):
LGTM!

template.yml
```diff
@@ -264,7 +267,7 @@ Resources:
       ContentUri: org-code-javabuilder/font_config.zip
       Description: Add a font configuration file to enable use of fonts.
       LayerName: font-configuration
-  BuildAndRunJavaProjectFunction:
+  BuildAndRunJavaTheaterProjectFunction:
```
sanchitmalhotra126, line 270: Not sure if this is possible

sanchitmalhotra126 (approved):
LGTM!

api-gateway-routes/api_gateway_proxy_function.rb
```diff
@@ -50,11 +50,44 @@ def on_connect(event, context)
...
+      function_name: ENV['BUILD_AND_RUN_THEATER_PROJECT_LAMBDA_ARN'],
+      invocation_type: 'Event',
+      payload: JSON.generate(payload)
+    })
+  elsif authorizer['mini_app_type'] == 'playground'
+    response = lambda_client.invoke({
+      function_name: ENV['BUILD_AND_RUN_PLAYGROUND_PROJECT_LAMBDA_ARN'],
+      invocation_type: 'Event',
+      payload: JSON.generate(payload)
+    })
```
cat5inthecradle, line 78: Seeing this much duplicated code makes me wonder, as a reader, if `invocation_type` and `payload` are also different based on the app type. Since we're just changing the function name, it might be clearer to set a `function_name` local variable in the if block, or pluck the correct one out of a `function_names` hash. There's some error handling to add for the unknown name scenario, but something like this?

```ruby
# build the function_names hash, mapping names to ARN's
# check to see if the mini_app_type is not in the hash, use 'playground' or throw 400

response = lambda_client.invoke({
  function_name: function_names[authorizer['mini_app_type'],
  invocation_type: 'Event',
  payload: JSON.generate(payload)
})
```

template.yml
```diff
@@ -276,8 +279,74 @@ Resources:
...
+      Description: Compile and execute a JavaLab Console project.
+      MemorySize: 512
+      Timeout: 60
+      Role:
+        Fn::ImportValue: JavabuilderLambdaExecutionRole
+      Environment:
+        Variables:
+          AWS_LAMBDA_EXEC_WRAPPER: /opt/change_runtime_directory
+          OUTPUT_BUCKET_NAME: !Ref OutputBucket
+          GET_OUTPUT_URL: !Sub "https://${OutputDomain}"
```
cat5inthecradle, line 335: This duplication is a little harder to avoid. Might be a good place for a [Cloudformation Mapping](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/mappings-section-structure.html), so you can signal to the reader that memory type and timeout are the interesting and likely-to-be-configured bits.

cat5inthecradle (approved):
Just some notes on readability, otherwise LGTM!

template.yml
jmkulwik: I also realized I might be able to use a pattern similar to this "each" pattern: https://github.com/code-dot-org/code-dot-org/blob/004deda4d0d5408904a76eaec5c1345f190abbde/aws/cloudformation/components/alarms.yml.erb#L3

jmkulwik: Hah I just realized that's a .yml.erb.

jmkulwik: Actually, that gives me an idea, is it bad practice to use that mapping for all the fields that are the same? i.e. then it's an "update in one place" change whenever we update one of the shared fields.

jmkulwik: Could look something like this:
```
Mappings: 
  SharedLambdaSettings: 
    Handler: org.code.javabuilder.LambdaRequestHandler::handleRequest
    Runtime: java11
    CodeUri: org-code-javabuilder/lib/build/distributions/lib.zip
    AutoPublishAlias: live
    ...
  UniqueLambdaSettings
    Playground:
      MemorySize: 512
      Timeout: 60
...
      Handler: !FindInMap [SharedLambdaSettings, Handler]
      Runtime: !FindInMap [SharedLambdaSettings, Runtime]
      CodeUri: !FindInMap [SharedLambdaSettings, CodeUri]
      AutoPublishAlias: !FindInMap [SharedLambdaSettings, AutoPublishAlias]
      MemorySize: !FindInMap [UniqueLambdaSettings, Playground, MemorySize]
      Timeout: !FindInMap [UniqueLambdaSettings, Playground, Timeout]
```

template.yml
```diff
@@ -185,7 +185,10 @@ Resources:
           # The Logical ID of the BuildAndRun Lambda Alias, which is generated by SAM because AutoPublishAlias is enabled
           # has a predictable format: <function‑LogicalId>Alias<alias‑name>
           # https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-specification-generated-resources-function.html#sam-specification-generated-resources-function-autopublishalias
-          BUILD_AND_RUN_PROJECT_LAMBDA_ARN: !Ref BuildAndRunJavaProjectFunctionAliaslive
+          BUILD_AND_RUN_THEATER_PROJECT_LAMBDA_ARN: !Ref BuildAndRunJavaTheaterProjectFunctionAliaslive
+          BUILD_AND_RUN_NEIGHBORHOOD_PROJECT_LAMBDA_ARN: !Ref BuildAndRunJavaNeighborhoodProjectFunctionAliaslive
+          BUILD_AND_RUN_PLAYGROUND_PROJECT_LAMBDA_ARN: !Ref BuildAndRunJavaPlaygroundProjectFunctionAliaslive
+          BUILD_AND_RUN_CONSOLE_PROJECT_LAMBDA_ARN: !Ref BuildAndRunJavaConsoleProjectFunctionAliaslive
```
jmkulwik, line 191: Note: I think we'll want to adjust the names here a bit to make them easier to use in mappings elsewhere. i.e. something like this https://github.com/code-dot-org/code-dot-org/blob/004deda4d0d5408904a76eaec5c1345f190abbde/aws/cloudformation/components/alarms.yml.erb#L24

sureshc:
MIght be worth dynamically provisioning the CloudWatch Alarms for Invocations and Severe Errors in this template, since we’ll have multiple Lambdas to monitor now. [Here’s an ERB](https://github.com/code-dot-org/code-dot-org/blob/staging-next/aws/cloudformation/components/alarms.yml.erb) that renders a snippet of CloudFormation that provisions CloudWatch Alarms. We may want to conditionally create the Alarms only when the Javabuilder stack is production (perhaps by checking the subdomain and basedomain = `javabuilderbeta.code.org`?)

template.yml.erb
```diff
@@ -0,0 +1,383 @@
...
+      ContentUri: org-code-javabuilder/font_config.zip
+      Description: Add a font configuration file to enable use of fonts.
+      LayerName: font-configuration
+<%{
+  Theater: {MemorySize: 1769, Timeout: 90},
+  Neighborhood: {MemorySize: 512, Timeout: 60},
+  Console: {MemorySize: 512, Timeout: 60},
+  Playground: {MemorySize: 512, Timeout: 300}
+}.each do |name, config| -%>
+  BuildAndRunJava<%=name%>ProjectFunction:
```
cat5inthecradle, line 280: One of my minor preferences about cloudformation yaml files is that I like to have a blank line between resources. This is one of those places where that can come in handy because currently I don't immediately know whether I'm still in the context of `FontConfigurationLayer.Properties` until I get to the line after the erb code.

template.yml
cat5inthecradle: resolving this thread because the ERB solution looks great. As much as I don't want to overcomplicate it by bringing in another layer, it comes out pretty clean here.

template.yml
```diff
@@ -42,6 +42,7 @@ Resources:
       AliasTarget:
         DNSName: !GetAtt DomainName.RegionalDomainName
         HostedZoneId: !GetAtt DomainName.RegionalHostedZoneId
+
```
jmkulwik, line 45: As a last step, I'm going to delete the template.yml file so there's one source of truth. Until then, leaving this here so folks can compare against the .erb file if needed.

deploy.sh
```diff
@@ -24,7 +24,7 @@ STACK=${SUB_DOMAIN}
 
 PROVISIONED_CONCURRENT_EXECUTIONS=${PROVISIONED_CONCURRENT_EXECUTIONS-'1'}
 RESERVED_CONCURRENT_EXECUTIONS=${RESERVED_CONCURRENT_EXECUTIONS-'3'}
-
+cat template.yml.erb | ruby -rerb -e "puts ERB.new(ARGF.read, nil, '-').result" > template.yml
```
sureshc, line 27: I think we can use the command line `erb` utility that's included with Ruby and set the trim mode to `-`?

``` bash
erb -T - template.yml.erb > template.yml
```

template.yml.erb
```diff
@@ -0,0 +1,543 @@
+AWSTemplateFormatVersion: 2010-09-09
```
sureshc, line 1: I think if we declare a constant array at the beginning of this template, we can use that in every template tag where we need the list of app types:

``` ruby
<%
  JAVALAB_APP_TYPES = %w(
   Theater
   Neighborhood
   Console
   Playground
  )
%>
```

sureshc: Or maybe declare one constant hash here instead with the list of app types, and store the configuration settings for each app type in the hash. We can then iterate over the keys in the hash when we just need the list of app types, and also access the specific Lambda config settings for each app type in the hash as well.

deploy.sh
jmkulwik: 👍

template.yml.erb
jmkulwik: 👍 added the array for now. Will look into the hash as a follow-up.

PR #158, opened 2021-12-07
Unify checks for .java extension in Javabuilder

sanchitmalhotra126:
Moves checks for the `.java` extension to a single util class.

JIRA: https://codedotorg.atlassian.net/browse/CSA-873

Testing: unit + checked locally that user experience is unchanged


molly-moen (approved):
🎉

PR #159, opened 2021-12-10
Add disk space usage logs

molly-moen:
Add some extra logging around our disk space usage in javabuilder. This is to diagnose no disk space errors.


org-code-javabuilder/protocol/src/main/java/org/code/protocol/LoggerConstants.java
```diff
@@ -0,0 +1,19 @@
+package org.code.protocol;
+
+public class LoggerConstants {
```
sanchitmalhotra126, line 3: Nice

sanchitmalhotra126 (approved):

PR #161, opened 2021-12-13
Clear entire temp directory after code execution on lambda

sanchitmalhotra126:
This change clears the entire temp directory after code execution finishes in `CodeBuilder`. Because we need lambda and local-specific logic (we don't want to clear the whole tmp directory on localhost), the deleting logic was moved into `JavabuilderFileManager`

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSFileManager.java
+++ b/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSFileManager.java
@@ -12,7 +12,6 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Date;
 import org.code.protocol.*;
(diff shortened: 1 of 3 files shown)


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSFileManager.java
```diff
@@ -101,6 +105,17 @@ public URL getFileUrl(String filename) throws FileNotFoundException {
     }
   }
 
+  @Override
+  public void cleanUpTempDirectory(File tempFolder) throws IOException {
+    // Delete any leftover contents of the tmp folder from previous lambda invocations
+    // We can ignore the tempFolder since we are clearing the entire directory
+    Path toClear = Paths.get(System.getProperty("java.io.tmpdir"));
+    LoggerUtils.sendDiskSpaceLogs();
```
molly-moen, line 113: nit: can we move the logs into the recursivelyClearDirectory method so we don't need to copy/paste the pattern a bunch of times? It will also make it easier to adjust the logs in the future.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeBuilder.java
```diff
@@ -92,12 +94,8 @@ public void close() throws InternalFacingException {
...
     if (this.tempFolder != null) {
       try {
-        // Recursively delete the temp folder
-        Path toClear = this.tempFolder.toPath();
-        LoggerUtils.sendDiskSpaceLogs();
-        Util.recursivelyClearDirectory(toClear);
-        LoggerUtils.sendClearedDirectoryLog(toClear);
-        LoggerUtils.sendDiskSpaceLogs();
+        // Clean up the temp folder and temp directory
+        this.fileManager.cleanUpTempDirectory(this.tempFolder);
```
molly-moen, line 98: nice solution!

molly-moen (approved):
LGTM, just one nit

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSFileManager.java
sanchitmalhotra126: Yep will do

PR #162, opened 2021-12-13
Log identifier for lambda containers

bencodeorg:
Adds a persistent identifier for lambda containers to our logging.

Jessie remembered reading that the request handler persists across the lifetime of a lambda container (~~we can't find the source of this information~~ check out initialization code section [here for more information](https://docs.aws.amazon.com/lambda/latest/dg/java-handler.html)), so that is where we implemented this change.

We did try to prove it out by the following though:
- repeated runs of the same program on a dev deploy return the same UUID (as verified via Postman)
- if one program is executing (ie, by putting a wait in such that the program does not complete its execution), executing another program returns a different UUID

This change should log the container ID any time we log to cloudwatch.

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
+++ b/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
@@ -67,6 +67,8 @@ public class LambdaRequestHandler implements RequestHandler<Map<String, String>,
         JSONUtils.booleanFromJSONObjectMember(options, "useNeighborhood");
     final List<String> compileList = JSONUtils.listFromJSONObjectMember(options, "compileList");
 
+    // Generate an identifier that persists across uses of the same lambda container
+    // if one has not yet been created. Used for logging purposes.
     if (LambdaRequestHandler.LAMBDA_ID.equals("")) {
       LambdaRequestHandler.LAMBDA_ID = UUID.randomUUID().toString();
     }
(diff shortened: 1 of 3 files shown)


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
```diff
@@ -34,6 +36,7 @@
   private static final int CHECK_THREAD_INTERVAL_MS = 500;
   private static final int TIMEOUT_WARNING_MS = 20000;
   private static final int TIMEOUT_CLEANUP_BUFFER_MS = 5000;
+  private static String LAMBDA_ID = "";
```
sanchitmalhotra126, line 39: nit: initialize to `null` and check if null before assigning?

sanchitmalhotra126 (approved):
Looks good! One small nit

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
molly-moen: per the [docs](https://docs.aws.amazon.com/lambda/latest/dg/java-handler.html)(search for "initialization code" on that page), the way to docs recommend initializing a static variable would be:
```
private static String LAMBDA_ID = UUID.randomUUID().toString();
```
This way works too, but doing it above the handler is a bit cleaner.

molly-moen: looks like alternatively we could create a constructor and initialize things there as well.

bencodeorg: hah well that is probably where jessie remembered this from :). Simplified the logic, thanks for the info

molly-moen (approved):
🎉

PR #164, opened 2021-12-13
Reuse fonts across Lambda Invocations

molly-moen:
Lambdas can re-use objects across invocations if they are created as static variables ([documentation](https://docs.aws.amazon.com/lambda/latest/dg/java-handler.html)). We load our fonts for theater and playground from file, which creates temporary files that can cause available disk space to decrease across runs. This change only loads the fonts once and re-uses the generated object across lambda invocations.

To do this I created a new `CachedResources` object, which we can add cache-able resources to as we see fit. This object is accessible via `GlobalProtocol`.


org-code-javabuilder/protocol/src/main/java/org/code/protocol/CachedResources.java
```diff
@@ -0,0 +1,49 @@
...
+        "LiberationSans-BoldItalic.ttf",
+        "LiberationSans-Italic.ttf",
+        "LiberationSans-Regular.ttf",
+        "LiberationSerif-Bold.ttf",
+        "LiberationSerif-BoldItalic.ttf",
+        "LiberationSerif-Italic.ttf",
+        "LiberationSerif-Regular.ttf"
+      };
+
+  public Map<String, Font> getFontMap() {
```
jmkulwik, line 30: Rather than creating the font map here, could we make it in FontHelper and then set it here? I'm thinking CachedResources could end up with a lot of imports and logic that are related to the item it is caching rather than to the caching itself.

org-code-javabuilder/protocol/src/main/java/org/code/protocol/CachedResources.java
```diff
@@ -0,0 +1,49 @@
...
+import java.io.InputStream;
+import java.util.HashMap;
+import java.util.Map;
+
+// Resources we want to use across multiple invocations of Javabuilder.
+public class CachedResources {
+  private Map<String, Font> fontMap;
+
+  private static final String FONT_FOLDER_NAME = "fonts";
+  private static final String[] FONT_FILE_NAMES =
```
bencodeorg, line 14: These constants and `getFontMap()` feel like they might still belong in `FontHelper` (or something else new with font in the name), then we keep this object as part of this new `Cached Resources` class?

bencodeorg (commented):
Looks good! Just an organizational thought

org-code-javabuilder/protocol/src/main/java/org/code/protocol/CachedResources.java
molly-moen: we can't have a reference to `media` from `protocol` because `media` references `protocol`, and trying to do the opposite causes a circular reference error. It does end up with some code duplication across the two modules, I tried to keep this as limited in scope as possible. Let me know if you have ideas for a better pattern.

org-code-javabuilder/protocol/src/main/java/org/code/protocol/CachedResources.java
molly-moen: see my comment on Jessie's comment, we have a limitation with not being able to reference media from protocol.

org-code-javabuilder/protocol/src/main/java/org/code/protocol/CachedResources.java
molly-moen: we could pass in the map and file names as parameters, but I think the files themselves will need to live in `protocol` since we are importing them here. So it makes more sense to have the filenames live in this class.

org-code-javabuilder/playground/src/test/java/org/code/playground/ImageItemTest.java
```diff
@@ -28,7 +28,8 @@ public void setUp() {
         "",
         "",
         "",
-        mock(JavabuilderFileManager.class));
+        mock(JavabuilderFileManager.class),
+        null);
```
sanchitmalhotra126, line 32: Idea for later - might be worth making a TestGlobalProtocol that can be constructed using a builder so we don't have to update all these tests whenever we make a change to `GlobalProtocol.create()`?

sanchitmalhotra126 (commented):
Nice! Agree with Jessie that FontHelper should probably still own the font-specific logic, but this approach looks good to me. Edit - just read your comments about dependencies; disregard this suggestion :)

org-code-javabuilder/protocol/src/main/java/org/code/protocol/CachedResources.java
sanchitmalhotra126: How about a third class like `FontLoader` that only takes care of loading fonts into memory? We'd still have Font-specific logic in `protocol` but it would at least be isolated to its own class?

molly-moen: yeah I can refactor out to a separate class inside protocol.

org-code-javabuilder/playground/src/test/java/org/code/playground/ImageItemTest.java
molly-moen: that's a good idea, I can pick that up as a follow-up

molly-moen: I'm going to leave it as-is for now because I think the best way to do this would be to have some TestUtils package that has helpers for our tests, which feels too heavy for this specific situation. Otherwise we would need to import a test class into other packages which isn't a great pattern. For now it's not too onerous to update this in multiple places.

jmkulwik (approved):
LGTM!

org-code-javabuilder/media/src/main/java/org/code/media/FontHelper.java
```diff
@@ -1,42 +1,20 @@
...
-import org.code.protocol.InternalServerRuntimeError;
+import org.code.protocol.GlobalProtocol;
 
 public class FontHelper {
   private final Map<String, java.awt.Font> fontMap;
 
   public FontHelper() {
-    this.fontMap = new HashMap<>();
-    this.populateFontMap();
+    this.fontMap = GlobalProtocol.getInstance().getCachedResources().getFontMap();
```
jamescodeorg, line 11: It seems a bit unexpected to call GlobalProtocol here.  I'm wondering if it would be better to cache the whole FontHelper class and get an instance via GlobalProtocol in Stage (where it looks like we're already getting some items off of GlobalProtocol).

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
```diff
@@ -33,7 +33,9 @@
   private static final int CHECK_THREAD_INTERVAL_MS = 500;
   private static final int TIMEOUT_WARNING_MS = 20000;
   private static final int TIMEOUT_CLEANUP_BUFFER_MS = 5000;
+  private static final CachedResources cachedResources = new CachedResources();
```
jamescodeorg, line 36: Just to check my understanding -- in order for something to be re-used across Lambda invocations, does it have to be a static _in this class_?  Or can statics declared elsewhere also be reused?

org-code-javabuilder/protocol/src/main/java/org/code/protocol/CachedResources.java
jamescodeorg: This seems like a bit of a yellow flag.  Is the problem that Font things aren't really part of the `protocol` but we'd like to put it in GlobalProtocol for convenience?  If that's the case, did we consider having CachedResources just be separate (and not part of GlobalProtocol)?  (Side note: this is the kind of thing that Spring dependency injection is supposed to make easy.  Not advocating that we switch to it, but I initially found it hard to understand why we would want to use DI and this feels like one of the target use cases.)

jamescodeorg (commented):
I was pinged about this review and took a quick look.  A ton has changed since I last looked at this code and I'm not sure I understand how everything fits together anymore so please ignore my comments if they don't make sense / aren't helpful.

org-code-javabuilder/protocol/src/main/java/org/code/protocol/CachedResources.java
molly-moen: The problem is that we need to load the fonts from the `LambdaRequestHandler` as a static variable, then access them from within `media`. Right now our way of sharing resources we create with student code is to put them in the global protocol instance. What do you mean by having CachedResources be separate?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
molly-moen: it does have to be static in this class specifically

org-code-javabuilder/media/src/main/java/org/code/media/FontHelper.java
molly-moen: the reason I have it here is we need the `FontHelper` to have access to media's `Font` and `FontStyle` classes, which students also need access to. However, students don't have access to `protocol`. I can instead pass in the font map as a constructor parameter so we don't need to reference global protocol here.

org-code-javabuilder/protocol/src/main/java/org/code/protocol/CachedResources.java
molly-moen: ah do you mean we get a static instance of CachedResources separate from our static instance of GlobalProtocol? That might be cleaner

sanchitmalhotra126 (approved):
LGTM!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
```diff
@@ -34,6 +34,12 @@
   private static final int TIMEOUT_WARNING_MS = 20000;
   private static final int TIMEOUT_CLEANUP_BUFFER_MS = 5000;
   private static final String LAMBDA_ID = UUID.randomUUID().toString();
+
+  public LambdaRequestHandler() {
+    // create CachedResources once for the entire container
```
jmkulwik, line 39: Quick nit: Can you add a little more detail here about resource reuse - and perhaps a link to the lambda docs describing it?

jmkulwik (approved):
Nice update. LGTM

org-code-javabuilder/protocol/src/main/java/org/code/protocol/CachedResources.java
```diff
@@ -0,0 +1,28 @@
...
+  private static CachedResources cachedResourcesInstance;
+  private Map<String, Font> fontMap;
+
+  public static void create() {
+    CachedResources.cachedResourcesInstance = new CachedResources();
+  }
+
+  public static CachedResources getInstance() {
+    if (CachedResources.cachedResourcesInstance == null) {
+      throw new InternalServerRuntimeError(InternalErrorKey.INTERNAL_EXCEPTION);
```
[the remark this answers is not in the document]
molly-moen: done!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
molly-moen: done!

PR #165, opened 2021-12-13
Reorganize disk space logs

sanchitmalhotra126:
Reorganizes how we log disk space slightly. Now, only one disk space log report (type: `"diskSpaceUtilization"`) is sent at the start of each session. However, we still do also log before and after clearing the temp directory at the start and end of the session, but these have more specific metric types (ex. `"diskSpaceUtilization-startSession-afterClear"`). This way, our existing metrics and graphs will still work but now just go off of the report published at the beginning of the session, and the other metrics are available to us for further analysis.

Tested on dev lambda.


molly-moen (approved):
Nice!

PR #166, opened 2021-12-13
Add warning log for unavailable aws resources

molly-moen:
Add a warning log if we hit a `QueueDoesNotExistException` from SQS or a `GoneException` from ApiGateway. These are only a warning because we can hit them if the user terminates their program early via the stop button. However, we want to log them for now to see if we are getting these exceptions when we do not expect them.


org-code-javabuilder/protocol/src/main/java/org/code/protocol/LoggerUtils.java
```diff
@@ -33,6 +33,18 @@
     }
   }
 
+  /**
+   * Log a warning-level message for the given exception
+   *
+   * @param e
+   */
+  public static void sendWarningForException(Exception e) {
```
sanchitmalhotra126, line 41: Nice! There are a couple spots in `CodeBuilderRunnable` where I think we could use this, but we can consolidate that later.

sanchitmalhotra126 (approved):
LGTM! I think there will be some merge conflicts with the threading change since `hasActiveConnection` no longer exists, but should be good otherwise!

org-code-javabuilder/protocol/src/main/java/org/code/protocol/LoggerUtils.java
molly-moen: Realized this PR is sort of moot after your change, since we now log for gone exceptions. I think I'll abandon this one, but we can add this warning helper later if it would help!

molly-moen:
Closing because [this PR](https://github.com/code-dot-org/javabuilder/pull/172) now means we already log a warning if the connection goes away.

PR #163, opened 2021-12-13
Add progress updates for theater video generation

sanchitmalhotra126:
This change adds progress messages to theater that output after a set amount of time has passed (currently set to 5 seconds). For this to appear on the client, this will require a separate Javalab change as well (https://github.com/code-dot-org/code-dot-org/pull/44068).

Screen Shot 2021-12-10 at 12 23 30 PM

Testing: tested locally & unit

PR #169, opened 2021-12-15
Added error message when a theater project's output is longer than 120s

jmkulwik:
This adds on to a [change](https://github.com/code-dot-org/javabuilder/pull/163) that logs a message to the user after every 5s of theater output generated. 

Here, we hook in to the onPause method. If a project will add more than 120s to its output, it will trigger an error.

Currently, the error is not translatable. This is because we're trying to get this out as an urgent fix for the disk space issue. As a follow-up, we should add a translated version of this error on the client side.

This change also adds back in the "Generating results..." message - this will allow us to merge and deploy the Javabuilder changes prior to the dashboard changes. As a follow-up, we should remove this message.

Error output:
image

Success output:
image

Note: The above output includes the message to the user after every 5s of Theater output generated. This has not been merged into production yet. Currently, the output would look like this:

Error output:
image

Success output:
image

### DIFF
--- a/org-code-javabuilder/theater/src/main/java/org/code/theater/TheaterProgressPublisher.java
+++ b/org-code-javabuilder/theater/src/main/java/org/code/theater/TheaterProgressPublisher.java
@@ -9,6 +9,7 @@ import org.code.protocol.*;
  */
 class TheaterProgressPublisher {
   private static final double UPDATE_TIME_S = 5.0;
+  private static final double MAX_TIME_S = 120.0;
 
   private final OutputAdapter outputAdapter;
   private double pauseTimeSeconds;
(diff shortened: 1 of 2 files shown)


bencodeorg (approved):
LGTM -- do you think longer term we should keep track of the length of the generated video elsewhere (or rename `TheaterProgressPublisher`)? Seems like a little odd of a place to manage how long a video can be.

org-code-javabuilder/theater/src/main/java/org/code/theater/TheaterProgressPublisher.java
```diff
@@ -26,6 +27,13 @@ public TheaterProgressPublisher() {
 
   public void onPause(double seconds) {
     this.pauseTimeSeconds += seconds;
+    if (this.pauseTimeSeconds > MAX_TIME_S) {
+      String message =
```
molly-moen, line 31: (something to maybe make a future task for): should we make this translatable?

molly-moen (approved):
LGTM!

org-code-javabuilder/theater/src/main/java/org/code/theater/TheaterProgressPublisher.java
jmkulwik: Good call. Here's the task! https://codedotorg.atlassian.net/browse/CSA-1108

PR #171, opened 2021-12-16
Limit size of theater gifs

molly-moen:
Limit theater gifs to 30 mb in size, and ensure we clean up theater resources if we terminate the program due to a too-long or too-large gif.

I also noticed when we write the gif and audio to file we have a stream that's not closed, so added a close to that as well.

As a follow-up to this, also ensure we call `Theater.stage.close()` every time a theater mini-app project is run.

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSFileManager.java
+++ b/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSFileManager.java
@@ -1,7 +1,5 @@
 package org.code.javabuilder;
 
-import static org.code.protocol.LoggerNames.MAIN_LOGGER;
-
 import com.amazonaws.AbortedException;
 import com.amazonaws.HttpMethod;
 import com.amazonaws.SdkClientException;
@@ -16,9 +14,7 @@ import java.net.MalformedURLException;
 import java.net.URL;
 import java.nio.file.Paths;
 import java.util.Date;
-import java.util.logging.Logger;
 import org.code.protocol.*;
-import org.json.JSONObject;
 
 public class AWSFileManager implements JavabuilderFileManager {
   private final String outputBucketName;
@@ -72,16 +68,6 @@ public class AWSFileManager implements JavabuilderFileManager {
     }
 
     this.writes++;
-    try {
-      inputStream.close();
-    } catch (IOException e) {
-      // We could not close the stream. This won't affect the user but may affect us, so we log it
-      // as a severe error.
-      JSONObject eventData = new JSONObject();
-      eventData.put(LoggerConstants.EXCEPTION_MESSAGE, e.getMessage());
-      eventData.put(LoggerConstants.TYPE, "streamCloseError");
-      Logger.getLogger(MAIN_LOGGER).severe(eventData.toString());
-    }
     return this.getOutputURL + "/" + filePath;
   }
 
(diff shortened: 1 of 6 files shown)


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSFileManager.java
```diff
@@ -68,6 +72,16 @@ public String writeToFile(String filename, byte[] inputBytes, String contentType
     }
 
     this.writes++;
+    try {
+      inputStream.close();
```
sanchitmalhotra126, line 76: So I think we don't actually need to worry about this for ByteArrayInputStreams? According to the docs closing has no effect: https://docs.oracle.com/javase/7/docs/api/java/io/ByteArrayInputStream.html#close()

org-code-javabuilder/theater/src/main/java/org/code/theater/GifWriter.java
```diff
@@ -48,6 +50,13 @@ public GifWriter createGifWriter(ByteArrayOutputStream out) {
    */
   public void writeToGif(BufferedImage img, int delay) {
     try {
+      if (this.imageOutputStream.length() > MAX_STREAM_LENGTH_BYTES) {
+        Theater.stage.close();
+        // TODO: Make this translatable: https://codedotorg.atlassian.net/browse/CSA-1108
+        String message =
+            "Your video is too large. Please decrease the number of frames in your video and try again.";
+        throw new RuntimeException(message);
```
sanchitmalhotra126, line 58: InternalException or UserInitiatedException?

org-code-javabuilder/theater/src/main/java/org/code/theater/GifWriter.java
```diff
@@ -48,6 +50,13 @@ public GifWriter createGifWriter(ByteArrayOutputStream out) {
    */
   public void writeToGif(BufferedImage img, int delay) {
     try {
+      if (this.imageOutputStream.length() > MAX_STREAM_LENGTH_BYTES) {
+        Theater.stage.close();
```
sanchitmalhotra126, line 54: It does seem a bit weird for GifWriter to call Theater.stage directly, but totally fine with this for now. I think once we have a cleanup notifier in place, we can hook into that as well, but this is fine for now.

sanchitmalhotra126 (approved):
Nice! For making sure `Theater.stage.close()` is called on exit, we should be able to hook into the listener I'm adding in the threading PR

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSFileManager.java
molly-moen: oh good call, I'll remove this

org-code-javabuilder/theater/src/main/java/org/code/theater/GifWriter.java
molly-moen: I think I'll keep this as a runtime exception until we add the translation piece on the frontend, if we do an InternalException I think we'll lose the message.

org-code-javabuilder/theater/src/main/java/org/code/theater/GifWriter.java
molly-moen: yeah I'll add a comment to clean this up once we have a notifier

jmkulwik (approved):
LGTM!

PR #172, opened 2021-12-16
Remove threading from code execution and ensure cleanup happens

sanchitmalhotra126:
This PR introduces a couple changes in an effort to mitigate some of the live site issues we've been seeing this week. Because of how the issue organically evolved throughout the week, the size of this PR was inflated a bit and now includes a couple different logical components, but hopefully some commentary will make it easier to follow :)

### Changes summary:
1. Necessary setup and cleanup steps are moved out of `CodeBuilder` and into a wrapper class that manages its execution, called `CodeExecutionManager`. This is in an effort to ensure that cleanup always happens even if code execution is interrupted for some reason (unexpected exception thrown, etc).
2. Code execution is now synchronous (i.e. no longer in a thread). `CodeExecutionManager` calls `run()` on `CodeBuilderRunnable` directly rather than spinning up a separate thread. Threading was removed because we noticed some potential unwanted possibilities of threads still running after the handler had returned, and because in our current system, our interrupt mechanism was not working the way we intended. The caveat here is that we still need a way to terminate execution if a connection has closed (ex. student clicking the Stop button) - our best solution for now is that if code tries to send or receive messages via the Input/OutputAdapters when the connection has closed, we throw an exception and return early. This doesn't do anything for long running projects that never send/receive messages; however, we at least will block on executing code until its complete rather than returning from the handler early, so it should hopefully ensure that a container is only executing one set of code at a time.
3. We expose a proxy for objects to listen for lifecycle events, specifically just when execution has ended. This is useful (and necessary) for objects like Stage which open streams (GifWriter and AudioWriter) and need to have them closed. This was already somewhat mitigated by previous changes that called `stage.close()` if we hit a storage or time limit while writing Theater content, but this ensures that we'll always clean up Theater resources (if used) when exiting, even if the student's code doesn't call `play()`
4. Listening for timeout still does happen in a separate thread (in LambdaRequestHandler). This thread checks the timeout threshold in the same way, and if a timeout is pending, it alerts the CodeExecutionManager to clean up early. Because our only mechanism to disrupt code currently is an adapter throwing an exception, execution could still keep going until we time out, but this at least ensures we do our clean up steps before then.

### Scenario walkthrough
- Theater project uses up too many resources (storage / time).
  - GifWriter/TheaterProgressPublisher throw RuntimeExceptions. These are caught in `CodeBuilderRunnable` and sent to the user. `run()` returns as normal and `CodeExecutionManager` performs `onPostExecute`, which notifies `Stage` to clean up resources 
- Theater project never calls `play()`
  - Project finishes execution normally. In `onPostExecute()`, `Stage` is asked to `close()` and because `play()` was never called, it does its clean up then. 
- Project times out (long execution time or waiting for input)
  - The thread running in `LambdaRequestHandler` first sends a warning, and then, when the buffer threshold is reached, notifies `CodeExecutionManager` to clean up resources. `CodeExecutionManager` does post execution steps and the lambda performs its cleanup. The container times out and reports a timeout.
- Project is stopped in the middle of execution
  - If the project sends or receives any more messages, the adapters throw an exception, which is caught by `CodeBuilderRunnable`. Seeing that it's a CONNECTION_TERMINATED error, it returns early (doesn't send the EXITED message). `CodeExecutionManager` performs cleanup. If the project does not send/receive any more messages, code execution continues as normal, blocking execution on the lambda. Once execution is finished, cleanup is performed, and the handler returns.

Testing: dev instance of lambda with Postman


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeBuilderRunnable.java
```diff
@@ -56,17 +59,14 @@ private void executeCodeBuilder() {
...
+        Logger.getLogger(MAIN_LOGGER).warning(e.getLoggingString());
         return;
       }
-      JSONObject eventData = new JSONObject();
-      eventData.put(LoggerConstants.EXCEPTION_MESSAGE, e.getExceptionMessage());
-      eventData.put(LoggerConstants.LOGGING_STRING, e.getLoggingString());
-      eventData.put(LoggerConstants.CAUSE, e.getCause());
       // The error was caused by us (essentially an HTTP 5xx error). Log it so we can fix it.
-      Logger.getLogger(MAIN_LOGGER).severe(eventData.toString());
+      LoggerUtils.logError(e.getExceptionMessage(), e.getLoggingString(), e.getCause());
```
molly-moen, line 69: nice utility function :)

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeBuilderRunnable.java
```diff
@@ -86,16 +86,12 @@ private void executeCodeBuilder() {
...
-      Logger.getLogger(MAIN_LOGGER).severe(eventData.toString());
+      LoggerUtils.logError(error);
 
       // Additionally, these may have affected the user. For now, let's tell them about it.
       outputAdapter.sendMessage(error.getExceptionMessage());
-    } finally {
-      GlobalProtocol.getInstance().cleanUpResources();
-      this.outputAdapter.sendMessage(new StatusMessage(StatusMessageKey.EXITED));
     }
+    // Send exit message if we haven't already returned
```
molly-moen, line 94: nit: this comment is a little confusing. Maybe add why we don't need to send an exit message for cases when we return early?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
```diff
@@ -105,8 +105,17 @@ public String handleRequest(Map<String, String> lambdaInput, Context context) {
     final AmazonS3 s3Client = AmazonS3ClientBuilder.standard().build();
     final AWSFileManager fileManager =
         new AWSFileManager(s3Client, outputBucketName, javabuilderSessionId, getOutputUrl, context);
+    final LifecycleNotifier lifecycleNotifier = new LifecycleNotifier();
+    // TODO: Should GlobalProtocol.create() happen in CodeExecutionManager#onPreExecute so it can
+    // also be destroyed in #postExecute?
```
molly-moen, line 110: should this comment still be here? What clean up do we need to do on the protocol instance besides the cleanUpResources we already call?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
```diff
@@ -145,45 +151,58 @@ public String handleRequest(Map<String, String> lambdaInput, Context context) {
...
+        new CodeExecutionManager(
+            userProjectFileLoader,
+            GlobalProtocol.getInstance().getInputHandler(),
+            outputAdapter,
+            executionType,
+            compileList,
+            fileManager,
+            lifecycleNotifier);
+
+    final Thread timeoutNotifierThread =
```
molly-moen, line 164: nit: can we factor this out into a separate class/function?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeExecutionManager.java
```diff
@@ -0,0 +1,145 @@
...
+    this.overrideOutputStream = new OutputPrintStream(this.outputAdapter);
+    System.setOut(this.overrideOutputStream);
+    System.setIn(this.overrideInputStream);
+  }
+
+  /**
+   * Post-execution steps: 1) Notify listeners, 2) clean up global resources, 3) clear temporary
+   * folder, 4) close custom in/out streams, 5) Replace System.in/out with original in/out
+   */
+  private void onPostExecute() throws InternalServerError {
```
molly-moen, line 122: will there be any issues if onPostExecute() is called twice? Not sure if it's possible since we only call `requestEarlyExit` on a timeout, but want to double check.

org-code-javabuilder/theater/src/main/java/org/code/theater/GifWriter.java
```diff
@@ -51,8 +51,6 @@ public GifWriter createGifWriter(ByteArrayOutputStream out) {
   public void writeToGif(BufferedImage img, int delay) {
     try {
       if (this.imageOutputStream.length() > MAX_STREAM_LENGTH_BYTES) {
-        // TODO: Remove this once we have a clean up notifier for stage.
-        Theater.stage.close();
```
molly-moen, line 55: 🎉

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
```diff
@@ -421,7 +422,7 @@ public void play() {
       this.gifWriter.writeToGif(this.image, 0);
       this.audioWriter.writeToAudioStream();
       // We must call close before write so that the streams are flushed.
```
molly-moen, line 424: nit: update this comment to reflect new method name

molly-moen (commented):
This looks awesome! Just a few minor comments

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
sanchitmalhotra126: Good catch, I think this was left around from an earlier idea - I was thinking maybe we needed to null out the GlobalProtocol instance when cleaning up, but not sure if that's necessary anymore

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeExecutionManager.java
sanchitmalhotra126: So ideally `this.executionInProgress` should guard against that, but I'm realizing I never set it to false if execution completes as normal, so I'll fix that

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeExecutionManager.java
```diff
@@ -0,0 +1,145 @@
...
+      this.fileManager.cleanUpTempDirectory(this.tempFolder);
+      LoggerUtils.sendDiskSpaceUpdate(SessionTime.END_SESSION, ClearStatus.AFTER_CLEAR);
+      // Close custom input/output streams
+      this.overrideInputStream.close();
+      this.overrideOutputStream.close();
+    } catch (IOException e) {
+      throw new InternalServerError(InternalErrorKey.INTERNAL_EXCEPTION, e);
+    }
+    // Replace System in/out with original System in/out
+    System.setIn(this.systemInputStream);
```
bencodeorg, line 142: Should these be in `finally`?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
```diff
@@ -145,45 +151,58 @@ public String handleRequest(Map<String, String> lambdaInput, Context context) {
...
+                      && !timeoutWarningSent) {
+                    outputAdapter.sendMessage(new StatusMessage(StatusMessageKey.TIMEOUT_WARNING));
+                    timeoutWarningSent = true;
+                  }
+
+                  if (context.getRemainingTimeInMillis() < TIMEOUT_CLEANUP_BUFFER_MS) {
+                    outputAdapter.sendMessage(new StatusMessage(StatusMessageKey.TIMEOUT));
+                    // Tell the execution manager to clean up early
+                    codeExecutionManager.requestEarlyExit();
+                    // Clean up Lambda resources
```
bencodeorg, line 182: Nit: I think this cleans up stuff outside of lambda (maybe "AWS resources"?)

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSOutputAdapter.java
```diff
@@ -63,13 +38,10 @@ public void sendDebuggingMessage(ClientMessage message) {
...
 
   private void sendMessageHelper(PostToConnectionRequest post) {
-    if (!this.hasActiveConnection) {
-      return;
-    }
     try {
       this.api.postToConnection(post);
     } catch (GoneException e) {
-      this.hasActiveConnection = false;
+      throw new InternalServerRuntimeError(InternalErrorKey.CONNECTION_TERMINATED, e);
```
bencodeorg, line 44: General question -- how does this PR affect our current reporting of "SEVERE" errors? Will this be logged as severe?

bencodeorg (commented):
Impressive work! Mostly questions for my understanding :)

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
sanchitmalhotra126: Reworded this comment, but gonna leave this TODO here because I think we can still move some common setup steps (like creating GlobalProtocol) down into CodeExecutionManager so its shared by lambda and websocketserver.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeExecutionManager.java
sanchitmalhotra126: Added a test case to verify this as well.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeExecutionManager.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+  }
+
+  /**
+   * Request the code execution to cleanup early (e.g. in the case of an impending timeout). Code
+   * execution may still run, but this allows for post execution steps to still occur before the
+   * environment is shut down.
+   */
+  public void requestEarlyExit() {
+    if (!this.executionInProgress) {
```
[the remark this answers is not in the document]
sanchitmalhotra126: Yeah this is just a safeguard because `requestEarlyExit` is a public method - if we somehow accidentally call this from elsewhere before we call `execute()` this ensures that we don't do post execution steps. Very unlikely scenario, but just to be safe :)

org-code-javabuilder/protocol/src/main/java/org/code/protocol/LifecycleNotifier.java
```diff
@@ -0,0 +1,25 @@
+package org.code.protocol;
+
+import java.util.HashSet;
+import java.util.Set;
+
+/** Notifies delegates of code execution lifecycle events. */
+public class LifecycleNotifier implements LifecycleListener {
```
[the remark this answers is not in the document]
sanchitmalhotra126: The LifecycleListener interface? I added that as part of this PR, but it's modeled off of other listener/proxy interface patterns i've seen.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeExecutionManager.java
sanchitmalhotra126: Hmm, yep probably, good call

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSOutputAdapter.java
sanchitmalhotra126: This particular error will not because it is caught in CodeBuilderRunnable and logged as a warning, but in general our SEVERE error reporting should not change - we should be reporting the same way for all internal exceptions.

org-code-javabuilder/protocol/src/main/java/org/code/protocol/LifecycleNotifier.java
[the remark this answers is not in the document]
bencodeorg: Oh oops, missed that :) There are a couple existing interfaces with the same name (but different methods, which I did not notice) when I googled. Carry on

molly-moen (approved):
Looks great!

PR #173, opened 2021-12-16
No retries in build and run lambda

bencodeorg:
Lambda by default will retry twice on timeout or error -- this sets it such that we do not retry. This will be particularly useful if we start using `System.exit` (which will error) after all projects.

Deployment might be a little tricky, as on my dev deploy this setting did not stick until I fully deleted/recreated my dev deploy.

A proposal would be to deploy this change, then change the setting manually. If we do this, we should check the next time that we deploy to make sure that the setting sticks.

### DIFF
--- a/template.yml
+++ b/template.yml
@@ -284,7 +284,9 @@ Resources:
         ProvisionedConcurrentExecutions: !Ref ProvisionedConcurrentExecutions
       Description: Compile and execute a JavaLab project.
       MemorySize: 1769
-      Timeout: 300
+      Timeout: 30
+      EventInvokeConfig:
+        MaximumRetryAttempts: 0
       Role:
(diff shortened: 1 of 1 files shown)


template.yml
```diff
@@ -284,7 +284,9 @@ Resources:
         ProvisionedConcurrentExecutions: !Ref ProvisionedConcurrentExecutions
       Description: Compile and execute a JavaLab project.
       MemorySize: 1769
-      Timeout: 300
+      Timeout: 30
```
molly-moen, line 287: this should still be 300 right?

bencodeorg: yes 😅 , thanks

molly-moen (approved):
LGTM!

jmkulwik (approved):
LGTM!

PR #174, opened 2021-12-17
Exit the JVM if the disk use exceeds 50% of available resources

jmkulwik:
Calling System.exit forces the JVM to exit and release all its resources. This means that a project that leaves a bunch of resources locked will be forced to release all its locks. 

Calling System.exit results in the maximum possible disk space being available the next time the lambda runs. 

However, calling System.exit has some side effects: The lambda runtime get a bit longer during the next execution. When tested on a 20s project, the project runtime extended to 28s. Additionally AWS tracks a System.exit() as an error in the dashboard. Finally, System.exit() triggers a lambda retry.

As part of this work, we will need to temporarily disable our alarms that check for errors on the lambdas. This change will intentionally increase the error rate.

Exiting results in following AWS message:
RequestId: <ID> Error: Runtime exited with error: exit status 50
Runtime.ExitError

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
+++ b/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
@@ -14,7 +14,9 @@ import com.amazonaws.services.s3.AmazonS3;
 import com.amazonaws.services.s3.AmazonS3ClientBuilder;
 import com.amazonaws.services.sqs.AmazonSQS;
 import com.amazonaws.services.sqs.AmazonSQSClientBuilder;
+import java.io.File;
 import java.io.IOException;
+import java.nio.file.Paths;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
@@ -183,6 +185,12 @@ public class LambdaRequestHandler implements RequestHandler<Map<String, String>,
       // program.
     } finally {
       cleanUpResources(connectionId, api);
+      File f = Paths.get(System.getProperty("java.io.tmpdir")).toFile();
+      if ((double) f.getUsableSpace() / f.getTotalSpace() < 0.5) {
+        // The current project holds a lock on too many resources. Force the JVM to quit in
+        // order to release the resources for the next use of the container.
+        System.exit(1);
+      }
     }
     return "done";
   }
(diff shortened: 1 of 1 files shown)


bencodeorg (approved):
LGTM -- should probably try to get this out at the same time (or after) the timeout change? Or should we wait and see the isolated impact of Molly's project size limit

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
```diff
@@ -183,6 +185,12 @@ public String handleRequest(Map<String, String> lambdaInput, Context context) {
       // program.
     } finally {
       cleanUpResources(connectionId, api);
+      File f = Paths.get(System.getProperty("java.io.tmpdir")).toFile();
+      if ((double) f.getUsableSpace() / f.getTotalSpace() < 0.5) {
+        // The current project holds a lock on too many resources. Force the JVM to quit in
+        // order to release the resources for the next use of the container.
+        System.exit(50);
```
molly-moen, line 192: can you add a comment as to what error code 50 means?

bencodeorg:
Will this lose our lambda identifier (and no longer re-use font files, but I think that's less of a worry)?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
jmkulwik: Done! Thanks

jmkulwik:
Answering a couple questions in one:
> should probably try to get this out at the same time (or after) the timeout change. Or should we wait and see the isolated impact of Molly's project size limit?

^ Agreed for the timeout change. As far as the isolated impact of Molly's change - we'll still be able to see it to some extent. But the signal will be the number of times we dip below 50% space rather than the number of times we dip below 100% space.

> Will this lose our lambda identifier (and no longer re-use font files, but I think that's less of a worry)?

^ Yes, it will. We can save that by making it an environment variable. That being said, exiting the JVM resets nearly everything about the container, so it's a little less important to keep the identifier between times when system.exit is called.

molly-moen (approved):
LGTM!

PR #176, opened 2022-01-03
Error if no java files on all compilations

bencodeorg:
Throw an error any time we try to compile a project with no java files. We currently do this only in "compile only" mode -- moves the check so it will happen any time we try to compile (whether the entire project, or when compiling a subset of files).

[Parallel PR on Javalab side](https://github.com/code-dot-org/code-dot-org/pull/44165) with user friendly error message. This can be merged without that Javalab piece in place, as it will just return a less pretty error message in the meantime.


sanchitmalhotra126 (approved):
LGTM!

molly-moen (approved):

PR #177, opened 2022-01-05
Update allowed classes

molly-moen:
Add a few more classes to our allowed class list. These classes encompass:
- primitive types that are wrapped in java.lang (such as `java.lang.Boolean`)
- a few more exceptions
- some helper packages such as `java.math`
- low-risk java utilities such as `StringBuilder`

## Future Work
After this update is merged we can turn the classloader fully on (it currently runs in silent mode).

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserClassLoader.java
+++ b/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserClassLoader.java
@@ -64,6 +64,7 @@ public class UserClassLoader extends URLClassLoader {
           "java.lang.ArithmeticException",
           "java.lang.ArrayIndexOutOfBoundsException",
           "java.lang.Boolean",
+          "java.lang.Byte",
           "java.lang.Character",
           "java.lang.CharSequence",
           "java.lang.Class",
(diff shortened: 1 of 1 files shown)


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserClassLoader.java
```diff
@@ -63,25 +63,43 @@ public UserClassLoader(URL[] urls, ClassLoader parent, List<String> userProvided
       Set.of(
           "java.lang.ArithmeticException",
           "java.lang.ArrayIndexOutOfBoundsException",
+          "java.lang.Boolean",
```
sanchitmalhotra126, line 66: small nit, not a blocker for this PR - do you think it might be better for us to get the class names programmatically (e.g. `Boolean.class.getName()`) for these rather than typing them out manually? just to help avoid any accidental errors

sanchitmalhotra126 (approved):
LGTM!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserClassLoader.java
molly-moen: that's a good idea, I think I'll do that as a follow-up to keep this PR as just the new classes we are allowing

PR #180, opened 2022-01-06
Revert "Revert "Remove threading from code execution and ensure cleanup happens""

sanchitmalhotra126:
Undoing the revert of https://github.com/code-dot-org/javabuilder/pull/172, which was previously reverted to avoid merging a big change before break.

### Changes summary (adapted from #172):
1. Necessary setup and cleanup steps are moved out of CodeBuilder and into a wrapper class that manages its execution, called CodeExecutionManager. This is in an effort to ensure that cleanup always happens even if code execution is interrupted for some reason (unexpected exception thrown, etc).
2. Code execution is now synchronous (i.e. no longer in a thread). CodeExecutionManager calls run() on CodeBuilderRunnable directly rather than spinning up a separate thread. Threading was removed because we noticed some potential unwanted possibilities of threads still running after the handler had returned, and because in our current system, our interrupt mechanism was not working the way we intended. The caveat here is that we still need a way to terminate execution if a connection has closed (ex. student clicking the Stop button) - our best solution for now is that if code tries to send or receive messages via the Input/OutputAdapters when the connection has closed, we throw an exception and return early. This doesn't do anything for long running projects that never send/receive messages; however, we at least will block on executing code until its complete rather than returning from the handler early, so it should hopefully ensure that a container is only executing one set of code at a time.
3. Listening for timeout still does happen in a separate thread (in LambdaRequestHandler). This thread checks the timeout threshold in the same way, and if a timeout is pending, it alerts the CodeExecutionManager to clean up early. Because our only mechanism to disrupt code currently is an adapter throwing an exception, execution could still keep going until we time out, but this at least ensures we do our clean up steps before then.


molly-moen (approved):
LGTM!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeBuilder.java
```diff
@@ -1,37 +1,27 @@
...
-import org.code.protocol.*;
-import org.code.protocol.LoggerUtils.ClearStatus;
-import org.code.protocol.LoggerUtils.SessionTime;
+import org.code.protocol.GlobalProtocol;
+import org.code.protocol.InternalErrorKey;
+import org.code.protocol.JavabuilderException;
+import org.code.protocol.OutputAdapter;
 
 /** The orchestrator for code compilation and execution. */
 public class CodeBuilder implements AutoCloseable {
```
jmkulwik, line 15: Do we still need to implement AutoCloseable if we aren't using the `close()` method?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeExecutionManager.java
```diff
@@ -0,0 +1,182 @@
...
+        inputHandler,
+        outputAdapter,
+        executionType,
+        compileList,
+        fileManager,
+        lifecycleNotifier,
+        new CodeBuilderRunnableFactory());
+  }
+
+  public CodeExecutionManager(
```
jmkulwik, line 69: Should this constructor be private?

org-code-javabuilder/protocol/src/main/java/org/code/protocol/LifecycleNotifier.java
```diff
@@ -0,0 +1,25 @@
+package org.code.protocol;
+
+import java.util.HashSet;
+import java.util.Set;
+
+/** Notifies delegates of code execution lifecycle events. */
+public class LifecycleNotifier implements LifecycleListener {
```
jmkulwik, line 7: Is it possible to pull the lifecycle changes into a separate PR? It's a pivotal pattern that will be adopted by Javabuilder going forward, and it's worth setting up as a separate PR that establishes the pattern as well as example usage in a well-documented and easily-referenced way.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeExecutionManager.java
```diff
@@ -0,0 +1,182 @@
...
+
+  /**
+   * Request the code execution to cleanup early (e.g. in the case of an impending timeout). Code
+   * execution may still run, but this allows for post execution steps to still occur before the
+   * environment is shut down.
+   */
+  public void requestEarlyExit() {
+    if (!this.executionInProgress) {
+      Logger.getLogger(MAIN_LOGGER)
+          .warning("Early exit requested while execution not in progress.");
```
jmkulwik, line 122: Do we need both this log and the one on line 158?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
```diff
@@ -110,8 +110,16 @@ public String handleRequest(Map<String, String> lambdaInput, Context context) {
     final AmazonS3 s3Client = AmazonS3ClientBuilder.standard().build();
     final AWSFileManager fileManager =
         new AWSFileManager(s3Client, outputBucketName, javabuilderSessionId, getOutputUrl, context);
+    final LifecycleNotifier lifecycleNotifier = new LifecycleNotifier();
+    // TODO: Move common setup steps into CodeExecutionManager#onPreExecute
```
jmkulwik, line 114: Is there a task for this TODO?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
```diff
@@ -195,9 +187,44 @@ public String handleRequest(Map<String, String> lambdaInput, Context context) {
         System.exit(LOW_DISK_SPACE_ERROR_CODE);
       }
     }
+
     return "done";
   }
 
+  private Thread createTimeoutThread(
```
jmkulwik, line 194: (perhaps out of scope) Would it make sense to add this to the common code path (i.e. the code path that's shared between LambdaRequestHandler, WebSocketServer, and the local execution)? That way this gets exercised both in local development as well as with a deployed version of Javabuilder.

org-code-javabuilder/protocol/src/main/java/org/code/protocol/LoggerUtils.java
```diff
@@ -61,6 +61,30 @@ public static void sendClearedDirectoryLog(Path p) {
     Logger.getLogger(MAIN_LOGGER).info(eventData.toString());
   }
 
+  public static void logError(JavabuilderException error) {
+    LoggerUtils.logError(error.getExceptionMessage(), error.getLoggingString(), error.getCause());
+  }
+
+  public static void logError(
```
jmkulwik, line 68: Nice add! This is another one that would be nice to pull into a separate PR. (Not as necessary since it's a smaller, more straightforward change)

jmkulwik (changes requested):
Unless it's a lot of extra work, I'd recommend pulling the Lifecycle change into its own PR. You're establishing a new, long-term pattern with it, and it'll be much easier to follow and document if it's separate from the rest of this work. 

A thought:
If both PRs need to be merged into main together, a good strategy is to use a PR chain where the first PR references main and the second PR references the first. 
i.e.
`main <- close-fix <- notifier-change`
^ Here, notifier-change would be merged into close-fix and then close-fix would be merged into main.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeBuilder.java
sanchitmalhotra126: Yep we still do since we use it in a try-with-resources block.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeExecutionManager.java
sanchitmalhotra126: Oh good call, this should be package-private (still needs to be referenced for testing)

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeExecutionManager.java
sanchitmalhotra126: I think so - we log here in case a consumer calls `requestEarlyExit()` before calling `execute()`, and log in `onPostExecute()` if we have requested an early exit while code was running and try to call `onPostExecute()` twice (once in `requestEarlyExit()` and again after `runnable.run()` has completed - we don't want the second call to execute)

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
sanchitmalhotra126: Don't think so, I'll create one

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
sanchitmalhotra126: I think to some extent yes - currently this thread uses lambda specific objects and values (Context, timeout values, etc) but we may be able to refactor and pull common logic into CodeExecutionManager. I'll add a note about this to the TODO above that mentions moving common steps into CodeExecutionManager.

org-code-javabuilder/protocol/src/main/java/org/code/protocol/LifecycleNotifier.java
sanchitmalhotra126: Hmm yeah, that makes sense. For this PR, it would mean reverting the behavior in `Stage` that now implements `LifecycleListener` but that shouldn't be too big of a change.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
sanchitmalhotra126: https://codedotorg.atlassian.net/browse/CSA-1146

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeBuilder.java
jmkulwik: Does it continue to need to be in a try-with-resources block?

sanchitmalhotra126: Hmm, yeah I guess not. I'll move it out.

sanchitmalhotra126:
> Unless it's a lot of extra work, I'd recommend pulling the Lifecycle change into its own PR. You're establishing a new, long-term pattern with it, and it'll be much easier to follow and document if it's separate from the rest of this work.
> 
> A thought: If both PRs need to be merged into main together, a good strategy is to use a PR chain where the first PR references main and the second PR references the first. i.e. `main <- close-fix <- notifier-change` ^ Here, notifier-change would be merged into close-fix and then close-fix would be merged into main.

I've moved the lifecycle changes out to a new PR, currently comparing against this current branch: https://github.com/code-dot-org/javabuilder/pull/185

bencodeorg:
@sanchitmalhotra126 which changes listed here are now in this PR?
https://github.com/code-dot-org/javabuilder/pull/172

bencodeorg:
I guess to add to Jessie's suggestion about branching off branches, a pattern I've seen is to make the base branch that will be merged into staging (in this case, `revert-blah-close-fix`) only have changes in it that have been reviewed in another branch (ie, you'd have another branch, say `remove-threading`, in parallel to `lifecycle-notifier`).

sanchitmalhotra126:
> @sanchitmalhotra126 which changes listed here are now in this PR? #172

Ah good question, I'll update the PR description. Should be everything except #3.

> I guess to add to Jessie's suggestion about branching off branches, a pattern I've seen is to make the base branch that will be merged into staging (in this case, `revert-blah-close-fix`) only have changes in it that have been reviewed in another branch (ie, you'd have another branch, say `remove-threading`, in parallel to `lifecycle-notifier`). Maybe not worth it in this case, but just something for the future.

Ahh interesting, yeah that makes sense. Since this just started as a revert commit, I was only working off of one branch, but I'll keep that in mind!

org-code-javabuilder/lib/src/test/java/org/code/javabuilder/CodeBuilderTest.java
```diff
@@ -19,32 +16,18 @@
...
     fileManager = mock(JavabuilderFileManager.class);
+    tempFolder = mock(File.class);
     when(globalProtocol.getFileManager()).thenReturn(fileManager);
     userProjectFiles = mock(UserProjectFiles.class);
 
-    codeBuilder = new CodeBuilder(globalProtocol, userProjectFiles);
-  }
-
-  @Test
-  void replacesSystemIOAndClearsTempWhenCloseCalled() {
```
bencodeorg, line 35: Should something similar to this test be moved to `CodeExecutionManager`?

bencodeorg (commented):
Looks great! Just one question

jmkulwik (approved):
Nice! Thanks for splitting out that other PR.

org-code-javabuilder/lib/src/test/java/org/code/javabuilder/CodeBuilderTest.java
sanchitmalhotra126: Hmm good call, I'll add a test.

PR #181, opened 2022-01-07
Create test fixture for using GlobalProtocol in tests

sanchitmalhotra126:
This is a small quality of life change/prefactor for upcoming changes to https://github.com/code-dot-org/javabuilder/pull/180 which helps us limit the amount of file changes we need to make whenever the contract of `GlobalProtocol.create()` is updated. Now, whenever that changes, only `GlobalProtocolTestFactory` needs to be updated, and test classes can choose which mocks they want to provide to the create method. All other fields are mocked/set to empty strings automatically.

Testing: all unit tests pass.


jmkulwik (approved):
Nice change!

PR #182, opened 2022-01-07
Added translatable messages and fallback messages

jmkulwik:
This does a couple of cleanup fixes as well as a quality of life improvement:

1. Add translations for messages from Javabuilder indicating that a video is too long or too large.
1. Remove the Generating Results message as it's no longer needed.
1. Add support for fallback messages so errors can be added to Javabuilder without a DTP.

This is paired with https://github.com/code-dot-org/code-dot-org/pull/44266 from the code-dot-org repo. This PR should be merged second.

- Jira ticket: [CSA-1108](https://codedotorg.atlassian.net/browse/CSA-1108)

### DIFF
--- a/org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
+++ b/org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
@@ -416,7 +416,6 @@ public class Stage {
     if (this.hasPlayed) {
       throw new TheaterRuntimeException(ExceptionKeys.DUPLICATE_PLAY_COMMAND);
     } else {
-      this.outputAdapter.sendMessage(new StatusMessage(StatusMessageKey.GENERATING_RESULTS));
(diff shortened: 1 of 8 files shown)


org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
```diff
@@ -416,7 +416,6 @@ public void play() {
     if (this.hasPlayed) {
       throw new TheaterRuntimeException(ExceptionKeys.DUPLICATE_PLAY_COMMAND);
     } else {
-      this.outputAdapter.sendMessage(new StatusMessage(StatusMessageKey.GENERATING_RESULTS));
```
sanchitmalhotra126, line 419: This is removed because the front end is no longer displaying this, correct?

sanchitmalhotra126 (approved):
LGTM, nice change!

PR #183, opened 2022-01-07
Custom Class Loader: always catch NoClassDefFoundError

molly-moen:
We discovered we weren't always showing the nice error message when student code used an invalid class. This stemmed from the error sometimes being thrown when we load the main method (it seems invalid exceptions may cause loading the function to fail). This PR also catches the `NoClassDefFoundError` inside `findMainMethod` and replaces it with the nicely formatted error.


jmkulwik (approved):
Nice fix!

PR #184, opened 2022-01-10
Do not try and close non-existent queue on disconnect

bencodeorg:
We noticed elevated execution errors in API Gateway while monitoring our websocket connectivity test, which is currently happening on applab and teacher dashboard page loads in Code Studio.

Errors were traced to our `on_disconnect`, where we try to delete a queue that does not exist. Catch when this occurs during connectivity tests, and add logging if it occurs in other cases.

**Testing Story**

Tested manually via Postman that a connectivity test no longer generates errors, a real project execution also does not generate any additional logging, and that the new logging appears when a queue does not exist (ie, if I reverse the `if is_connectivity_test condition).

### DIFF
--- a/api-gateway-routes/api_gateway_proxy_function.rb
+++ b/api-gateway-routes/api_gateway_proxy_function.rb
@@ -20,13 +20,14 @@ def on_connect(event, context)
 
   request_context = event["requestContext"]
   # Add in some logging to make debugging easier
-  puts request_context
+  puts "CONNECT REQUEST CONTEXT: #{request_context}"
 
-  # Return early if this is the user connectivity test
   authorizer = request_context['authorizer']
-  if authorizer && authorizer['connectivityTest']
+  # Return early if this is the user connectivity test
+  if is_connectivity_test(authorizer)
     return { statusCode: 200, body: "connection successful" }
   end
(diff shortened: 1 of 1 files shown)


api-gateway-routes/api_gateway_proxy_function.rb
```diff
@@ -62,7 +64,24 @@ def on_connect(event, context)
...
   sqs = Aws::SQS::Client.new(region: get_region(context))
-  sqs.delete_queue(queue_url: get_sqs_url(event, context))
+
+  # Handle if queue does not exist,
+  # such as in case of connectivity test.
+  # The NonExistentQueue error is not documented in the AWS SDK,
+  # but was observed in errors accumulated from our connectivity tests.
+  begin
+    sqs.delete_queue(queue_url: get_sqs_url(event, context))
+  rescue Aws::SQS::Errors::NonExistentQueue => e
```
bencodeorg, line 74: I found a project rescuing this (undocumented) error as far back as 2015, so I'm thinking it's pretty stable:

https://github.com/ruby-shoryuken/shoryuken/issues/150

molly-moen (approved):

PR #185, opened 2022-01-10
Add lifecycle listener and notifier pattern, and integrate with Theater classes

sanchitmalhotra126:
Note: this is a change that was originally introduced in https://github.com/code-dot-org/javabuilder/pull/172 but split out for clarity.

This change introduces a lifecycle listener pattern that allows us to hook components in to code execution lifecycle events. The expected use is that components can implement the `LifecycleListener` interface and register themselves with `GlobalProtocol.registerLifecycleListener()` to receive updates. Updates are dispatched via the `LifecycleNotifier`, which is currently managed by `CodeExecutionManager`, as it manages the code execution lifecycle directly and can notify listeners accordingly. In this specific change, the pattern is implemented in Theater with `Stage` - Stage implements the interface and registers itself with GlobalProtocol. This ensures that Stage will always be notified with code execution has ended and can perform its resource cleanup steps.

Our current use cases only require an event for when code execution has ended, but if and when we need to hook into other lifecycle events, we can do so in the same way.

Testing: local development and dev lambda

### DIFF
--- a/org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
+++ b/org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
@@ -421,19 +421,23 @@ public class Stage implements LifecycleListener {
       this.progressPublisher.onPlay(this.audioWriter.getTotalAudioLength());
       this.gifWriter.writeToGif(this.image, 0);
       this.audioWriter.writeToAudioStream();
-      // We must call onExecutionEnded() before write so that the streams are flushed.
-      this.onExecutionEnded();
+      // We must call close() before write so that the streams are flushed.
+      this.close();
       this.writeImageAndAudioToFile();
       this.hasPlayed = true;
     }
+  }
+
+  @Override
+  public void onExecutionEnded() {
+    this.close();
   }
 
   /**
    * Clean up resources created by this instance. If onExecutionEnded or play has already been
    * called this method does nothing.
    */
-  @Override
-  public void onExecutionEnded() {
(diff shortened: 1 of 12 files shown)


molly-moen (approved):
Nice!

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
```diff
@@ -420,18 +421,19 @@ public void play() {
       this.progressPublisher.onPlay(this.audioWriter.getTotalAudioLength());
       this.gifWriter.writeToGif(this.image, 0);
       this.audioWriter.writeToAudioStream();
-      // We must call close before write so that the streams are flushed.
-      this.close();
+      // We must call onExecutionEnded() before write so that the streams are flushed.
+      this.onExecutionEnded();
```
bencodeorg, line 425: Maybe a nit (or misunderstanding on my part), but could we rename the contents of `onExecutionEnded()` into a helper that gets called both here and in `onExecutionEnded()`?

bencodeorg (approved):
Awesome! Much easier to read, thanks for splitting out

org-code-javabuilder/theater/src/main/java/org/code/theater/Stage.java
sanchitmalhotra126: Yep sure

jmkulwik (approved):
Thanks for splitting this out. Looks great!

PR #188, opened 2022-01-22
Use Stack-specific names for CloudWatch Alarms

sureshc:
Include subdomain name in CloudWatch Alarm Names to ensure each CloudWatch Alarm Name is unique within an AWS account. Without this change, it isn't possible to provision more than one Javabuilder CloudFormation Stack in a single AWS account. We informally have a 1:1 mapping between AWS accounts and base domain names (production --> `code.org` / development --> `dev-code.org`) so the subdomain name should be sufficient for ensuring uniqueness. Eliminate `_javabuilder` prefix for Alarm Name because by convention, the subdomain always includes `javabuilder` prefix (`javabuilderbeta.code.org`, `javabuilder-staging.code.org`, `javabuilder-suresh.dev-code.org`, etc.).

Example: javabuilder-suresh.dev-code.org and javabuilder-mrcunningham.dev-code.org have different Alarm Names:
image

### DIFF
--- a/template.yml.erb
+++ b/template.yml.erb
@@ -414,7 +414,7 @@ Resources:
   HighConcurrentExecutionsAlarm:
     Type: AWS::CloudWatch::Alarm
     Properties:
-      AlarmName: !Sub "${SubDomainName}_javabuilder_high_concurrent_executions"
+      AlarmName: !Sub "${SubDomainName}_high_concurrent_executions"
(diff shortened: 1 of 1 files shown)


jmkulwik (approved):
LGTM. Let me know when you merge so I can check that our dashboard is still working as expected

cat5inthecradle (approved):

PR #189, opened 2022-01-24
Add org.code.validation package

molly-moen:
Add a new org.code.validation package that will be used for validation specific code. Currently the package just contains a placeholder for the new Test Output Adapter.

# Links
- [CSA-410](https://codedotorg.atlassian.net/browse/JAVA-410)


org-code-javabuilder/validation/build.gradle
```diff
@@ -0,0 +1,19 @@
...
+
+repositories {
+    // Use JCenter for resolving dependencies.
+    jcenter()
+}
+
+dependencies {
+    implementation project(':protocol')
+    testImplementation 'org.junit.jupiter:junit-jupiter-api:5.7.0'
+    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.7.0'
```
jmkulwik, line 14: worth adding the fat-jar packaging step too?

org-code-javabuilder/validation/src/main/java/org/code/validation/TestOutputAdapter.java
```diff
@@ -0,0 +1,11 @@
+package org.code.validation;
+
+import org.code.protocol.ClientMessage;
+import org.code.protocol.OutputAdapter;
+
+public class TestOutputAdapter implements OutputAdapter {
```
jmkulwik, line 6: NIT: Given the java convention of naming the test directory "test" and our convention of using "test" in the names of test files, I'm trying to think if there's a good way to disambiguate this class from the rest of our test code. Perhaps "UserTestOutputAdapter" or something similar?

molly-moen: I think we would need a new word to encompass running student unit tests + levelbuilder tests. Currently we are also using "test" for `TestRunner`, which will run both of those tests.

org-code-javabuilder/validation/build.gradle
molly-moen: good call

molly-moen: done!

org-code-javabuilder/validation/src/main/java/org/code/validation/TestOutputAdapter.java
molly-moen: thinking more on this, `UserTestOutputAdapter` should work, we would probably also want to update `TestRunner` to `UserTestRunner` and not use "user" when referring just to student unit tests (maybe call them project tests or something?)

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/Util.java
```diff
@@ -15,7 +15,11 @@
 /** A set of static utility functions that are used in multiple locations */
 public class Util {
   private static final String[] JAR_NAMES = {
-    "neighborhood-full.jar", "theater-full.jar", "playground-full.jar", "studentlib-full.jar"
+    "neighborhood-full.jar",
+    "theater-full.jar",
+    "playground-full.jar",
+    "studentlib-full.jar",
+    "validation-full.jar"
```
jmkulwik, line 22: (May be outside the scope of this change) I remember there was some discussion on only including validation when running test code. Would we want to selectively include the jar?

molly-moen: I was planning to handle this as part of the class loader. Currently no one can actually use an org.code.validation package because it's not in the allow list (they can import it, but that won't help much)

org-code-javabuilder/validation/build.gradle
```diff
@@ -0,0 +1,34 @@
...
+}
+
+test {
+    useJUnitPlatform()
+}
+
+// Builds the Validation fat-jar (i.e. jar with all dependencies) and copies it to the resources
+// folder in the javabuilder sub-project (currently named 'lib'). This will allow student code to
+// access the Playground package.
```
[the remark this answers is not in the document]
molly-moen: thanks! fixed

jmkulwik (approved):
LGTM!

PR #190, opened 2022-01-25
Implement and use UserTestOutputAdapter

molly-moen:
Add a basic implementation of `UserTestOutputAdapter`. This will send on `SYSTEM_OUT` and `STATUS` messages in a project test run, and only `STATUS` messages in a validation run. It also has TODOs for sending on `TEST_RESULT` messages once we add that message type, and for parsing`NEIGHBORHOOD` messages in validation runs. This output adapter is now used when the execution type is `TEST`.

## Links
- [spec](https://docs.google.com/document/d/14_ka0o1YFhAoXFl-WiyukrfIvT4C0IYIVVYcFaEmV7w/edit#bookmark=id.3rhp19k1pe0b)
- [jira](https://codedotorg.atlassian.net/browse/JAVA-405)

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketServer.java
+++ b/org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketServer.java
@@ -16,7 +16,7 @@ import javax.websocket.Session;
 import javax.websocket.server.ServerEndpoint;
 import org.code.javabuilder.*;
 import org.code.protocol.*;
-import org.code.validation.TestOutputAdapter;
+import org.code.validation.UserTestOutputAdapter;
 import org.json.JSONObject;
 
 /**
@@ -79,7 +79,7 @@ public class WebSocketServer {
     outputAdapter = websocketOutputAdapter;
 
     if (executionType == ExecutionType.TEST) {
-      outputAdapter = new TestOutputAdapter(websocketOutputAdapter);
+      outputAdapter = new UserTestOutputAdapter(websocketOutputAdapter);
     }
     final LifecycleNotifier lifecycleNotifier = new LifecycleNotifier();
     GlobalProtocol.create(
--- a/org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketServer.java
+++ b/org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketServer.java
@@ -16,6 +16,7 @@ import javax.websocket.Session;
 import javax.websocket.server.ServerEndpoint;
 import org.code.javabuilder.*;
 import org.code.protocol.*;
+import org.code.validation.TestOutputAdapter;
 import org.json.JSONObject;
 
 /**
@@ -28,7 +29,8 @@ import org.json.JSONObject;
 @ServerEndpoint("/javabuilder")
 public class WebSocketServer {
   private WebSocketInputAdapter inputAdapter;
-  private WebSocketOutputAdapter outputAdapter;
+  private WebSocketOutputAdapter websocketOutputAdapter;
+  private OutputAdapter outputAdapter;
   private Handler logHandler;
   private Logger logger;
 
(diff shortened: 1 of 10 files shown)


org-code-javabuilder/validation/src/test/java/org/code/validation/ClientMessageHelper.java
```diff
@@ -0,0 +1,10 @@
+package org.code.validation;
+
+import org.code.protocol.ClientMessage;
+import org.code.protocol.ClientMessageType;
+
+public class ClientMessageHelper extends ClientMessage {
```
bencodeorg, line 6: For my understanding, is this just for tests? Or will it be something more generic than when we extend `ClientMessage` elsewhere? I would have expected `ValidationMessage` or something?

org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketServer.java
```diff
@@ -28,7 +29,8 @@
 @ServerEndpoint("/javabuilder")
 public class WebSocketServer {
   private WebSocketInputAdapter inputAdapter;
-  private WebSocketOutputAdapter outputAdapter;
+  private WebSocketOutputAdapter websocketOutputAdapter;
```
bencodeorg, line 32: Does this need to be an instance variable any more?

bencodeorg (commented):
Looks good! Just a couple nits

bencodeorg:
Since we're just getting started with this, can you clarify the definitions of "project test run" and a "validation run"? Just want to make sure I understand the basics :).

org-code-javabuilder/validation/src/main/java/org/code/validation/UserTestOutputAdapter.java
```diff
@@ -1,11 +1,43 @@
...
+    this.isValidation = false;
+    this.delegateOutputAdapter = delegateOutputAdapter;
+  }
+
   @Override
   public void sendMessage(ClientMessage message) {
     // TODO: write method
+    ClientMessageType messageType = message.getType();
+    switch (messageType) {
+        // TODO: when we have a TEST_RESULT type send that on to the delegate as well
```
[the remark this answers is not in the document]
molly-moen: yes they are within one sprint so I don't think they'll need Jira links

org-code-javabuilder/validation/src/test/java/org/code/validation/ClientMessageHelper.java
molly-moen: this is just for tests, so we don't need to reference classes like `SystemOutMessage` and `TheaterMessage`

org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketServer.java
molly-moen: I think if we leave it as an instance variable we get some reuse over multiple executions, I'm not sure if it's strictly necessary though.

org-code-javabuilder/validation/src/test/java/org/code/validation/ClientMessageHelper.java
bencodeorg: 👍 maybe a comment to that effect then?

bencodeorg (approved):

molly-moen:
> Since we're just getting started with this, can you clarify the definitions of "project test run" and a "validation run"? Just want to make sure I understand the basics :).

@bencodeorg the definition I'm using it project run is student provided tests, while validation run is levelbuilder provided tests. Open to suggestions if there's better terminology ideas.

bencodeorg:
> > Since we're just getting started with this, can you clarify the definitions of "project test run" and a "validation run"? Just want to make sure I understand the basics :).
> 
> @bencodeorg the definition I'm using it project run is student provided tests, while validation run is levelbuilder provided tests. Open to suggestions if there's better terminology ideas.

Ok that's what I thought :). I guess my confusion is the two are one execution (right? ie, one click of the Test button) and I think of "run" as synonymous with "execution"? I like the second wording you used ("student-provided tests", and "levelbuilder-provided tests")

org-code-javabuilder/validation/src/test/java/org/code/validation/ClientMessageHelper.java
molly-moen: done!

PR #191, opened 2022-01-25
Fix web socket server

molly-moen:
When we removed threading from Javabuilder we erroneously removed threading from `WebSocketServer`. `WebSocketServer` needs to run the student code in a thread so it can receive input messages while code is running. This change goes back to running student code in a thread.


bencodeorg:
Will this change result in local runs deviating further from deployed javabuilder? Or do we not really have a choice?

molly-moen:
> Will this change result in local runs deviating further from deployed javabuilder? Or do we not really have a choice?

They already run in a slightly different way, so it's not terribly different than it was before. We don't really have a choice here since everything for the websocket server happens on one process.

bencodeorg (approved):
Thanks for fixing!

PR #193, opened 2022-01-26
Add basics of Neighborhood Validation API

molly-moen:
Add the files and most of the implementation of the Neighborhood Validation API. This API will be used by Levelbuilders when they write validation tests. I ended up splitting `PainterLog` into two classes: `PainterLog`, which is as defined in the product spec, and `PainterTracker`, which will not be referenced by Levelbuilder code. `PainterTracker` will track painter events as they come in, and can generate a `PainterLog` which used for analysis of a static list of painter events. 

There are a couple methods not implemented in this PR, each has a link to the work item in which they will be implemented. These methods are:
- `NeighborhoodTestRunner.run`: we need to implement the logic to access the main method of student code outside of `MainRunner`
- `PainterTracker.trackEvent`: we have a separate work item for tracking neighborhood events, which will also include writing `NeighborhoodTracker`.

Tested with unit tests. This code is not used anywhere yet so it is very safe.

## Links
- [Product Spec](https://docs.google.com/document/d/1C1-_7vN2LRHDkkLdDDVrR9pRs2At1qGgUZv4Nzaa26o/edit#bookmark=id.1yyqkb5ybyy7)
- [Dev Doc](https://docs.google.com/document/d/14_ka0o1YFhAoXFl-WiyukrfIvT4C0IYIVVYcFaEmV7w/edit#bookmark=id.qn2gnkpw600b)
- [Jira](https://codedotorg.atlassian.net/browse/JAVA-411)


org-code-javabuilder/validation/src/main/java/org/code/validation/NeighborhoodLog.java
```diff
@@ -0,0 +1,96 @@
+package org.code.validation;
+
+public class NeighborhoodLog {
+  private PainterLog[] painterLogs;
+  private String[][] finalOutput;
```
jmkulwik, line 5: should these be `final`

org-code-javabuilder/validation/src/main/java/org/code/validation/EventTypeMapper.java
```diff
@@ -0,0 +1,34 @@
+package org.code.validation;
+
+import org.code.neighborhood.NeighborhoodSignalKey;
+
+public class EventTypeMapper {
+  // Map NeighborhoodSignalKey to EventType. This allows us to change NeighborhoodSignalKeys without
+  // needing to update the validation api.
```
jmkulwik, line 7: I'm a bit on the fence about whether it makes sense to just use the NeighborhoodSignalKey enum directly - it seems like these need to be tightly coupled anyway since this is testing neighborhood code. I'm curious if you can share more detail about why you decided to do it this way. 

Edit: now that I'm thinking through this, I'm realizing it's likely because EventType is user-facing (i.e. where the user is the levelbuilder) and NeighborhoodSignalKey is an internal implementation detail of Javabuilder. So we may change these independent of each other and we definitely don't want to have to go through and edit all the level sources if we change one of the NeighborhoodSignalKey .

org-code-javabuilder/validation/src/main/java/org/code/validation/NeighborhoodLog.java
```diff
@@ -0,0 +1,96 @@
+package org.code.validation;
+
+public class NeighborhoodLog {
```
jmkulwik, line 3: NIT: Would `NeighborhoodLogAnalyzer` be more accurate?

org-code-javabuilder/validation/src/main/java/org/code/validation/NeighborhoodLog.java
```diff
@@ -0,0 +1,96 @@
...
+  /**
+   * @param eventType
+   * @param times
+   * @return true if the given eventType occurred exactly "times" times across all painters, false
+   *     otherwise.
+   */
+  public boolean actionHappened(EventType eventType, int times) {
+    int actionCount = 0;
+    for (PainterLog painterLog : this.painterLogs) {
+      actionCount += painterLog.actionCount(eventType);
```
jmkulwik, line 44: Do we need to worry about infinite loops or large inputs? i.e. if times = 1000000 or the student code loops thousands of times? (Not sure that we do - curious for your thoughts)

org-code-javabuilder/validation/src/main/java/org/code/validation/NeighborhoodLog.java
```diff
@@ -0,0 +1,96 @@
...
+      for (int j = 0; j < this.finalOutput[0].length; j++) {
+        String expectedValue = expectedOutput[i][j];
+        String actualValue = this.finalOutput[i][j];
+        // values will have object equality if both are null
+        if (expectedValue == actualValue) {
+          continue;
+        }
+        if ((actualValue == null && expectedValue != null)
+            || (actualValue != null && expectedValue == null)
+            || !actualValue.equals(expectedValue)) {
```
jmkulwik, line 69: Is the second condition needed? i.e. Does `actualValue.equals(null)` produce `false`?

org-code-javabuilder/validation/src/main/java/org/code/validation/NeighborhoodLog.java
```diff
@@ -0,0 +1,96 @@
...
+    return true;
+  }
+  /**
+   * @param expectedOutput 2d array of booleans, where true means the cell has paint.
+   * @return true if paint of any color is in the cells set to true in expectedOutput, false
+   *     otherwise.
+   */
+  public boolean finalOutputContainsPaint(boolean[][] expectedOutput) {
+    if (expectedOutput.length != this.finalOutput.length
+        || expectedOutput[0].length != this.finalOutput[0].length) {
```
jmkulwik, line 83: Do we need this check for every internal array? (not sure if Java even allows 2D arrays with different sized internal arrays)

org-code-javabuilder/validation/src/main/java/org/code/validation/NeighborhoodLog.java
jmkulwik, line 3: Mind adding a comment at the top of the classes noting which are user-facing?

org-code-javabuilder/validation/src/main/java/org/code/validation/PainterEvent.java
```diff
@@ -0,0 +1,21 @@
+package org.code.validation;
+
+import java.util.Map;
+
+public class PainterEvent {
+  private EventType eventType;
+  private Map<String, String> details;
```
jmkulwik, line 7: NIT: can these be final?

org-code-javabuilder/validation/src/main/java/org/code/validation/PainterLog.java
```diff
@@ -0,0 +1,110 @@
...
+  public boolean didActionAtLeast(EventType eventType, int times) {
+    if (this.eventCounts.containsKey(eventType)) {
+      return eventCounts.get(eventType) >= times;
+    }
+    return false;
+  }
+
+  /**
+   * @param eventType
+   * @return the number of time the painter did the given eventType.
```
jmkulwik, line 66: NIT: time -> times

org-code-javabuilder/validation/src/main/java/org/code/validation/PainterLog.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.Map;
+
+public class PainterLog {
+  private String painterId;
+  private List<PainterEvent> events;
+  private Map<EventType, Integer> eventCounts;
+  private Position startingPosition;
+  private Position endingPosition;
+  private int startingPaintCount;
+  private int endingPaintCount;
```
jmkulwik, line 14: NIT: Can these be final?

jmkulwik: (same for a couple other classes)

org-code-javabuilder/validation/src/main/java/org/code/validation/PainterTracker.java
```diff
@@ -0,0 +1,37 @@
+package org.code.validation;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class PainterTracker {
```
bencodeorg, line 6: This feels to me like the `PainterLog` class, and the current `PainterLog` class more like ` PainterLogValidator` (esp if we moved all the getters here)?

org-code-javabuilder/validation/src/test/java/org/code/validation/NeighborhoodLogTest.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class NeighborhoodLogTest {
+  private NeighborhoodLog unitUnderTest;
+
+  @BeforeEach
+  public void setup() {
```
bencodeorg, line 14: Don't think it matters, but I think every other test file uses setUp()

org-code-javabuilder/validation/src/main/java/org/code/validation/EventTypeMapper.java
molly-moen: yep that was why. Also because EventType is a clearer name for levelbuilders to use instead of having them use NeighborhoodSignalKey.

bencodeorg (commented):
Looks good to me, just a couple naming/organizational questions

org-code-javabuilder/validation/src/main/java/org/code/validation/NeighborhoodLog.java
molly-moen: probably, this class will be used by Levelbuilders though so I'm inclined to keep the names shorter

org-code-javabuilder/validation/src/main/java/org/code/validation/NeighborhoodLog.java
molly-moen: I think in those cases a couple things could happen: the code could be slow, but finish, we could time out, or we could hit some out of memory error. If we want student code to still pass then we would want to find a workaround, but my gut feeling is that would be a case where their code wouldn't pass, as they likely wouldn't even be able to run their code without timing out.

org-code-javabuilder/validation/src/main/java/org/code/validation/NeighborhoodLog.java
molly-moen: you're right, I can remove that. I was trying to avoid null pointer exceptions but over-corrected.

org-code-javabuilder/validation/src/main/java/org/code/validation/NeighborhoodLog.java
molly-moen: good catch, yes Java does allow different sized internal arrays

org-code-javabuilder/validation/src/main/java/org/code/validation/NeighborhoodLog.java
molly-moen: will do

org-code-javabuilder/validation/src/main/java/org/code/validation/NeighborhoodLog.java
```diff
@@ -0,0 +1,96 @@
...
+
+  /**
+   * @param eventType
+   * @param times
+   * @return true if the given eventType occurred exactly "times" times for at least one painter,
+   *     false otherwise.
+   */
+  public boolean onePainterDidAction(EventType eventType, int times) {
+    for (PainterLog painterLog : this.painterLogs) {
+      if (painterLog.didActionExactly(eventType, times)) {
```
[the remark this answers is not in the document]
molly-moen: good question, there isn't. This plus your previous question makes me think we should rename `EventType` to `NeighborhoodActionType`. I'll check with Mike

org-code-javabuilder/validation/src/main/java/org/code/validation/PainterTracker.java
molly-moen: `PainterLog` is what will be used by levelbuilder validation code, while `PainterTracker` will be used to generate the log (which we can think of as an after the fact log, not a changing log). I'm going to comment these classes to make them clearer. I might also look into splitting into separate folders or something to make things super clear. That being said, we need to keep the getters in `PainterLog` so that levelbuilder validation code can access it.

org-code-javabuilder/validation/src/main/java/org/code/validation/NeighborhoodLog.java
molly-moen: I ended up splitting out the "support" files into a sub-package to make things clearer (I also added comments).

org-code-javabuilder/validation/src/main/java/org/code/validation/EventType.java
```diff
@@ -0,0 +1,14 @@
+package org.code.validation;
+
+public enum EventType {
```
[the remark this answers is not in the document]
molly-moen: renamed to `NeighborhoodActionType`

org-code-javabuilder/validation/src/main/java/org/code/validation/PainterTracker.java
molly-moen: I ended up splitting out the support code into a sub-package

org-code-javabuilder/validation/src/main/java/org/code/validation/NeighborhoodLog.java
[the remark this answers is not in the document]
molly-moen: renamed to `NeighborhoodActionType`

bencodeorg (approved):
Awesome!

molly-moen:
@bencodeorg fyi I moved the new test result message class into the support folder here since it was a quick change and the pr was open.

bencodeorg:
> @bencodeorg fyi I moved the new test result message class into the support folder here since it was a quick change and the pr was open.

Thanks!

jmkulwik (approved):
LGTM!

PR #195, opened 2022-01-28
Add UserTestResultMessage message type

bencodeorg:
Adds a new UserTestResultMessage type to send test results to the client.

## Testing story

Added a unit test to confirm that test result messages are always forwarded. Tested via Javalab that a simple standalone test got a message of type `TEST_RESULT` was received.

## Follow-up work

Handle test results in Javalab. If this is merged before the Javalab PR, test results temporarily would not show up in the Javalab console.


bencodeorg:
@molly-moen wondering if `TestResultMessage` should go in the new `support` sub-package? Kind of breaks with the pattern in other packages, but maybe a better pattern to use going forward? I personally would love to have a clearer delineation between what is our public facing API, and what is support stuff. Not sure if there's a java pattern for that (or is it just that we should be able to tell based on what is public vs not?)

molly-moen:
@bencodeorg yeah I moved `UserTestOutputAdapter` into the `support` package, so it would make sense to move this into that package too. That package isn't merged in yet though, so it can be done as a follow-up if you end up merging before me.

I agree about a cleaner delineation between support and non-support, maybe something to discuss further as a team? It would probably require doing some refactoring of the entire code base.

org-code-javabuilder/validation/src/main/java/org/code/validation/TestResultMessage.java
```diff
@@ -0,0 +1,11 @@
+package org.code.validation;
+
+import org.code.protocol.ClientMessage;
+import org.code.protocol.ClientMessageType;
+
+/** A message describing a test result directed to the client's terminal. */
+public class TestResultMessage extends ClientMessage {
```
molly-moen, line 7: sorry for the late change, but it might be clearer to name this `UserTestResultMessage` to align with `UserTestOutputAdapter`. I think we can keep the enum value as `TEST_RESULT` though.

molly-moen (approved):
LGTM!

bencodeorg:
> @bencodeorg yeah I moved `UserTestOutputAdapter` into the `support` package, so it would make sense to move this into that package too. That package isn't merged in yet though, so it can be done as a follow-up if you end up merging before me.
> 
> I agree about a cleaner delineation between support and non-support, maybe something to discuss further as a team? It would probably require doing some refactoring of the entire code base.

Yeah totally, feels like a larger conversation. I'll merge this and move this into support once we have the sub-package.

PR #196, opened 2022-01-29
Give tests access to org.code.validation

molly-moen:
For now, give all user-written and levelbuilder-written tests access to the org.code.validation package. Students aren't writing unit tests until post-pilot, and the validation package is hidden mostly to avoid confusion, not because it exposes any information. To do this, I added an 'elevated permissions' option to the user class loader, and gave all test runs elevated permissions.

## Links
- [jira](https://codedotorg.atlassian.net/browse/JAVA-422)

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaRunner.java
+++ b/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaRunner.java
@@ -43,14 +43,17 @@ public class JavaRunner {
    *     finished executing.
    */
   public void runMain() throws InternalFacingException, JavabuilderException {
-    this.run(this.mainRunner);
+    this.run(this.mainRunner, false);
   }
 
   public void runTests() throws JavabuilderException, InternalFacingException {
-    this.run(this.testRunner);
+    // Tests have more permissions than a regular run--as of now, all
+    // tests have access to the org.code.validation package.
+    this.run(this.testRunner, true);
   }
 
-  private void run(CodeRunner runner) throws JavabuilderException, InternalFacingException {
+  private void run(CodeRunner runner, boolean hasElevatedPermissions)
+      throws JavabuilderException, InternalFacingException {
     // Include the user-facing api jars in the code we are loading so student code can access them.
     URL[] classLoaderUrls = Util.getAllJarURLs(this.executableLocation);
 
@@ -58,7 +61,10 @@ public class JavaRunner {
     // packages/classes.
     UserClassLoader urlClassLoader =
         new UserClassLoader(
-            classLoaderUrls, JavaRunner.class.getClassLoader(), this.javaClassNames);
+            classLoaderUrls,
+            JavaRunner.class.getClassLoader(),
+            this.javaClassNames,
+            hasElevatedPermissions);
 
     runner.run(urlClassLoader);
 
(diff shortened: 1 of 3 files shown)


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaRunner.java
```diff
@@ -43,22 +43,28 @@ public JavaRunner(
    *     finished executing.
    */
   public void runMain() throws InternalFacingException, JavabuilderException {
-    this.run(this.mainRunner);
+    this.run(this.mainRunner, false);
   }
 
   public void runTests() throws JavabuilderException, InternalFacingException {
-    this.run(this.testRunner);
+    // Tests have more permissions than a regular run--as of now, all
```
jmkulwik, line 50: I'm not sure if this was a typo or intentional - you have two dashes (--)

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserClassLoader.java
```diff
@@ -41,11 +47,13 @@ public UserClassLoader(URL[] urls, ClassLoader parent, List<String> userProvided
...
       return this.approvedClassLoader.loadClass(name);
     }
-    // allow .<specific-class> usage from valid packages
-    for (int i = 0; i < this.allowedPackages.length; i++) {
-      if (name.startsWith(this.allowedPackages[i])) {
-        return this.approvedClassLoader.loadClass(name);
-      }
+    // allow .<specific-class> usage from allowed packages. If this code has elevated permissions,
+    // also check the
+    // elevated permissions allowed package list.
```
jmkulwik, line 52: NIT: Spacing. (I think the linter made some odd choices here around line breaks.)

jmkulwik (approved):
LGTM. Nice design with elevated permissions.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaRunner.java
```diff
@@ -43,22 +43,28 @@ public JavaRunner(
...
+    // tests have access to the org.code.validation package.
+    this.run(this.testRunner, true);
   }
 
-  private void run(CodeRunner runner) throws JavabuilderException, InternalFacingException {
+  private void run(CodeRunner runner, boolean hasElevatedPermissions)
```
sanchitmalhotra126, line 55: Nice solution! I wonder, if there's a chance we may have a more complicated permissions model down the road, would it be worth making this an enum? i.e. PermissionLevel.USER, PermissionLevel.VALIDATOR/TESTER or something like that?

molly-moen: good call, we will probably want user, validator and tester down the line so I'll update this (for now just with user and tester, we can add validator later once we are running validation code separate from project test code)

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaRunner.java
molly-moen: it was intentional, meant to be an [em dash](https://www.grammarly.com/blog/dash/), probably should be a colon instead for clarity

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserClassLoader.java
molly-moen: yep I think this was the linter, will fix

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaRunner.java
jmkulwik: 👍

sanchitmalhotra126 (approved):
Nice!

PR #197, opened 2022-02-02
Add new status messages for validation and tests

molly-moen:
We will have two categories of tests: validation tests and project tests. Instead of a single "running" status we want a separate status for each category of test. This PR adds the new enum values for those statuses, and moves the status message to `TestRunner`, which is where we will split out validation and project test runs. For now, treat all tests as project tests until we add the ability to send validation test code separately.

Note: until [this PR](https://github.com/code-dot-org/code-dot-org/pull/44638) is deployed, if we deploy this change we will not see any "running" status message for tests. It is ok if these changes come in out of order since tests are behind a feature flag.


org-code-javabuilder/lib/src/test/java/org/code/javabuilder/JavabuilderTestExecutionListenerTest.java
```diff
@@ -109,10 +97,10 @@ public void testExecutionFinishedSendsErrorMessageIfTestFails() {
     unitUnderTest.testPlanExecutionStarted(testPlan);
     unitUnderTest.executionFinished(testIdentifier, testExecutionResult);
 
-    // 3 calls: 1) execution started, 2) result, 3) error details
-    verify(outputAdapter, times(3)).sendMessage(messageCaptor.capture());
+    // 3 calls: 1) result, 2) error details
```
jmkulwik, line 100: NIT: 3 calls -> 2 calls

jmkulwik: (same below)

jmkulwik (approved):
LGTM

org-code-javabuilder/protocol/src/main/java/org/code/protocol/StatusMessageKey.java
```diff
@@ -9,5 +9,7 @@
   SENDING_VIDEO,
   TIMEOUT_WARNING,
   TIMEOUT,
-  EXITED
+  EXITED,
+  RUNNING_VALIDATION,
+  RUNNING_PROJECT_TESTS
```
sanchitmalhotra126, line 14: small nit: seems like these are grouped in rough order of execution, could we move these alongside RUNNING?

sanchitmalhotra126 (approved):

org-code-javabuilder/protocol/src/main/java/org/code/protocol/StatusMessageKey.java
molly-moen: done!

PR #201, opened 2022-02-03
Move shared exception message methods to helper

bencodeorg:
In preparation for [having a shorter stack trace for students](https://codedotorg.atlassian.net/browse/JAVA-409), moving shared message formatting helpers (currently copied exactly between `JavabuilderException` and `JavabuilderRuntimeException`) into their own util class. This change should introduce no behavior changes.

## Testing story

Tested manually in Javalab that a `JavabuilderException` (dividing by 0 in student code) and a `JavabuilderRuntimeException` (throwing `RuntimeException` in student code) both printed the long stack trace that is the current behavior.

### DIFF
--- a/org-code-javabuilder/protocol/src/main/java/org/code/protocol/JavabuilderThrowableMessageHelper.java
+++ b/org-code-javabuilder/protocol/src/main/java/org/code/protocol/JavabuilderThrowableMessageHelper.java
@@ -1,0 +1,36 @@
+package org.code.protocol;
+
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.util.HashMap;
+
+/* Helper for formatting exceptions that will be displayed to the user. */
+class JavabuilderThrowableMessageHelper {
+  private JavabuilderThrowableMessageHelper() {}
+
+  static JavabuilderThrowableMessage getExceptionMessage(
+      Throwable throwable, Enum key, String fallbackMessage) {
+    HashMap<String, String> detail = new HashMap<>();
+    detail.put(ClientMessageDetailKeys.CONNECTION_ID, Properties.getConnectionId());
+    if (throwable.getCause() != null) {
+      detail.put(ClientMessageDetailKeys.CAUSE, getLoggingString(throwable));
+      if (throwable.getCause().getMessage() != null) {
+        detail.put(ClientMessageDetailKeys.CAUSE_MESSAGE, throwable.getCause().getMessage());
+      }
+    }
+
+    if (fallbackMessage != null) {
+      detail.put(ClientMessageDetailKeys.FALLBACK_MESSAGE, fallbackMessage);
+    }
+
+    return new JavabuilderThrowableMessage(key, detail);
+  }
+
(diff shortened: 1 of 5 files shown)


org-code-javabuilder/protocol/src/main/java/org/code/protocol/JavabuilderThrowableMessageHelper.java
```diff
@@ -0,0 +1,36 @@
+package org.code.protocol;
+
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.util.HashMap;
+
+/* Helper for formatting exceptions that will be displayed to the user. */
+class JavabuilderThrowableMessageHelper {
+  private JavabuilderThrowableMessageHelper() {}
```
molly-moen, line 9: is this constructor needed?

bencodeorg: Just to signal that it's a helper class that shouldn't be instantiated.

molly-moen: Since all the methods are static I don't think we need this. If you did try to instantiate it, it would be a useless class since it doesn't have instance methods.

bencodeorg: Yeah, I just thought it was helpful to be explicit that it shouldn't be done. It seems like a pretty common pattern for util classes to do this from what I'm reading?

I guess as someone new to Java, I find myself looking through our repo often and looking for indicators of how something should be used from the way a class is structured (eg, has public constructor?). Maybe pretty obvious in this case,  I'm in favor of more explicit indicators of what something is for / not for (along the lines of doing `support` sub-packages)

bencodeorg: As I'm adding the shorter stack trace, I'm finding it might make sense to make it an instantiable class anyway, so may be moot. Curious for your thoughts though.

molly-moen: I read more on this and you're right, it is a good pattern to have a private constructor. I like this [suggestion](https://www.baeldung.com/java-sonar-hide-implicit-constructor#1-default-private-constructor) of having the private constructor throw an exception so it's clear why it is there. We'll probably want to adopt this pattern for our other util classes whether or not this one ends up needing to be instantiable or not.

jmkulwik: Oops - crossed wires. What Molly said :)

molly-moen (approved):
🎉 thanks for adopting a new pattern for util classes!

sanchitmalhotra126 (approved):
Nice cleanup!

PR #202, opened 2022-02-04
Neighborhood Validation API

molly-moen:
Revert-revert of [this PR](https://github.com/code-dot-org/javabuilder/pull/193) with an additional change to fix an issue it caused with neighborhood. See that PR for details on what it originally contained.

This version includes an additional change to fix an issue where neighborhood would only work once, then hang forever. The cause was that the new validation package now includes neighborhood, whereas before neighborhood was only referenced by itself. The neighborhood classes are now in two places in the jars that we include with student code: in neighborhood-full.jar and validation-full.jar. For some reason (still not sure on the exact why here) this was causing better reuse of resources in the `World` class. World has a static instance you get [here](https://github.com/code-dot-org/javabuilder/blob/d04709b34074f13045441a70fc6257e524199a82/org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/World.java#L40). Previously, this static instance got reset between every run (again, not sure on the how). With 2 inclusions of neighborhood, the static instance was reused, which caused painter to try to write to a previous instance of the `outputAdapter`, which was closed, which caused an exception so everything failed. The solution is to add clean-up to `World` using our handy `LifecycleListener` pattern.

We will want to do further clean up to make sure we aren't repeating this static variable reuse elsewhere. There is a [tech debt item](https://codedotorg.atlassian.net/browse/JAVA-435) to address this.


sanchitmalhotra126 (commented):
One thought - if I'm understanding correctly, the issue here was that the `World` instance had an out of date reference to `OutputAdapter` right? Is there any reason it needs to have that reference in the first place? As far as I can tell, it's only used in `Painter` via `worldInstance.getOutputAdapater()`. If we were to just use `GlobalProtocol.getInstance().getOutputAdapter()` in painter, would that also solve the issue?

That being said though, this cleanup is a good idea regardless - I'm just wondering if we can also get rid of World#getOutputAdapter entirely.

molly-moen:
@sanchitmalhotra126 yes that was the specific issue, although re-use of world could also cause a project to be referencing the wrong grid (for example, say a square is an obstacle when it's not), so this change is helpful for that case as well. We never saw that because the second run would always fail before a painter tried to move. For `World` holding a reference to output adapter, we could remove it, there isn't really a benefit to it being there as far as I can tell.

sanchitmalhotra126:
> @sanchitmalhotra126 yes that was the specific issue, although re-use of world could also cause a project to be referencing the wrong grid (for example, say a square is an obstacle when it's not), so this change is helpful for that case as well. We never saw that because the second run would always fail before a painter tried to move. For `World` holding a reference to output adapter, we could remove it, there isn't really a benefit to it being there as far as I can tell.

Ah yep, that makes sense

sanchitmalhotra126 (approved):

PR #205, opened 2022-02-05
Shorten stack trace when shown to users

bencodeorg:
Provide a shortened version of the stack trace when we surface it to end users. Our logs will still receive the full stack trace, as we log both the message sent to the user (`error.getExceptionMessage()`) as well as the full stack trace (`error.getLoggingString()`):

https://github.com/code-dot-org/javabuilder/blob/60b5e19a701bf86fba0672841755287673b7e1d8/org-code-javabuilder/protocol/src/main/java/org/code/protocol/LoggerUtils.java#L64-L66

## Testing story

Tested manually that errors thrown in student code (`RuntimeException` and an arithmetic exception on divide by zero) produced shorter stack traces like the following:

```
[EXCEPTION] Your code hit an exception while it was running.
Exception type: java.lang.RuntimeException
	 at Game.play(Game.java:69)
	 at Game.main(Game.java:15)
```

Instead of:

```
[EXCEPTION] Your code hit an exception while it was running.
org.code.javabuilder.UserInitiatedException: RUNTIME_ERROR
	at org.code.javabuilder.MainRunner.run(MainRunner.java:54)
	at org.code.javabuilder.JavaRunner.run(JavaRunner.java:63)
	at org.code.javabuilder.JavaRunner.runMain(JavaRunner.java:46)
	at org.code.javabuilder.CodeBuilder.runUserCode(CodeBuilder.java:70)
	at org.code.javabuilder.CodeBuilderRunnable.executeCodeBuilder(CodeBuilderRunnable.java:52)
	at org.code.javabuilder.CodeBuilderRunnable.run(CodeBuilderRunnable.java:38)
	at org.code.javabuilder.CodeExecutionManager.execute(CodeExecutionManager.java:103)
	at dev.javabuilder.WebSocketServer.lambda$onOpen$0(WebSocketServer.java:113)
	at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.code.javabuilder.MainRunner.run(MainRunner.java:31)
	... 8 more
Caused by: java.lang.RuntimeException
	at Game.play(Game.java:69)
	at Game.main(Game.java:15)
	... 13 more
```

### DIFF
--- a/org-code-javabuilder/protocol/src/main/java/org/code/protocol/JavabuilderThrowableMessageUtils.java
+++ b/org-code-javabuilder/protocol/src/main/java/org/code/protocol/JavabuilderThrowableMessageUtils.java
@@ -2,6 +2,7 @@ package org.code.protocol;
 
 import java.io.PrintWriter;
 import java.io.StringWriter;
+import java.lang.reflect.InvocationTargetException;
 import java.util.HashMap;
 
 /* Helper for formatting exceptions that will be displayed to the user. */
@@ -14,10 +15,12 @@ public final class JavabuilderThrowableMessageUtils {
       Throwable throwable, Enum key, String fallbackMessage) {
     HashMap<String, String> detail = new HashMap<>();
     detail.put(ClientMessageDetailKeys.CONNECTION_ID, Properties.getConnectionId());
-    if (throwable.getCause() != null) {
-      detail.put(ClientMessageDetailKeys.CAUSE, getLoggingString(throwable));
-      if (throwable.getCause().getMessage() != null) {
-        detail.put(ClientMessageDetailKeys.CAUSE_MESSAGE, throwable.getCause().getMessage());
+
+    Throwable cause = throwable.getCause();
+    if (cause != null) {
+      detail.put(ClientMessageDetailKeys.CAUSE, getUserFacingLoggingString(cause));
+      if (cause.getMessage() != null) {
+        detail.put(ClientMessageDetailKeys.CAUSE_MESSAGE, cause.getMessage());
       }
     }
 
@@ -34,5 +37,25 @@ public final class JavabuilderThrowableMessageUtils {
     PrintWriter printWriter = new PrintWriter(stringWriter);
     throwable.printStackTrace(printWriter);
     return stringWriter.toString();
+  }
+
+  /**
+   * @return A pretty shortened version of the exception and stack trace for sharing with end users.
+   *     Private, as this is meant for use as a helper in the context of a client-facing message.
+   */
+  private static String getUserFacingLoggingString(Throwable cause) {
+    if (cause.getClass() == InvocationTargetException.class) {
+      cause = cause.getCause();
+    }
+    StackTraceElement[] stackTrace = cause.getStackTrace();
+    String loggingString = cause.toString() + "\n";
+
+    for (StackTraceElement stackTraceElement : stackTrace) {
+      if (stackTraceElement.isNativeMethod()) {
+        break;
+      }
+      loggingString += "\t" + stackTraceElement.toString() + "\n";
+    }
+    return loggingString;
   }
 }


org-code-javabuilder/protocol/src/main/java/org/code/protocol/JavabuilderThrowableMessageUtils.java
```diff
@@ -35,4 +38,24 @@ public static String getLoggingString(Throwable throwable) {
...
+  /**
+   * @return A pretty shortened version of the exception and stack trace for sharing with end users.
+   *     Private, as this is meant for use as a helper in the context of a client-facing message.
+   */
+  private static String getUserFacingLoggingString(Throwable cause) {
+    if (cause.getClass() == InvocationTargetException.class) {
+      cause = cause.getCause();
+    }
+    StackTraceElement[] stackTrace = cause.getStackTrace();
+    String loggingString = "Exception type: " + cause.toString() + "\n";
```
bencodeorg, line 51: I don't recall what our expectation is around translation in error messages at the moment, I remembered this PR:

https://github.com/code-dot-org/javabuilder/pull/182/files

Do we do translation as a follow-up?

molly-moen: Ideally we'd like everything to be translatable. We do have the general issue that a stack trace has some English words in it (such as 'at'), but if we can make a translatable version of the "Exception type: " message that would be helpful. Now that we have the fallback message option I think we can add both this message as the "fallback" and also have a key with parameters for things like the exception name and stack trace, then follow-up with the code-dot-org side change to add the new translatable string.

bencodeorg: Got it, will update!

molly-moen (approved):
Looks good, just one nit!

org-code-javabuilder/protocol/src/main/java/org/code/protocol/JavabuilderThrowableMessageUtils.java
```diff
@@ -14,15 +15,32 @@ public static JavabuilderThrowableMessage getExceptionMessage(
...
+      detail.put(ClientMessageDetailKeys.STACK_TRACE, getUserFacingStackTraceString(cause));
+      detail.put(ClientMessageDetailKeys.EXCEPTION_MESSAGE, getExceptionMessageString(cause));
+      if (cause.getMessage() != null) {
+        detail.put(ClientMessageDetailKeys.CAUSE_MESSAGE, cause.getMessage());
       }
     }
 
+    // Caller providing a fallback message overrides the default generated in this method.
     if (fallbackMessage != null) {
-      detail.put(ClientMessageDetailKeys.FALLBACK_MESSAGE, fallbackMessage);
```
sanchitmalhotra126, line 25: nit: should this check happen earlier so that we don't generate a preferredFallbackMessage if we don't need to?

org-code-javabuilder/protocol/src/main/java/org/code/protocol/JavabuilderThrowableMessageUtils.java
```diff
@@ -35,4 +53,31 @@ public static String getLoggingString(Throwable throwable) {
...
+    loggingString += getUserFacingStackTraceString(cause);
+    return loggingString;
+  }
+
+  private static String getUserFacingStackTraceString(Throwable cause) {
+    String stackTraceString = "";
+
+    StackTraceElement[] stackTrace = cause.getStackTrace();
+    for (StackTraceElement stackTraceElement : stackTrace) {
+      if (stackTraceElement.isNativeMethod()) {
```
sanchitmalhotra126, line 72: nice! does checking `isNativeMethod` filter out all AWS calls automatically?

sanchitmalhotra126 (approved):
One nit, otherwise looks good!

org-code-javabuilder/protocol/src/main/java/org/code/protocol/JavabuilderThrowableMessageUtils.java
```diff
@@ -14,15 +15,32 @@ public static JavabuilderThrowableMessage getExceptionMessage(
       Throwable throwable, Enum key, String fallbackMessage) {
     HashMap<String, String> detail = new HashMap<>();
     detail.put(ClientMessageDetailKeys.CONNECTION_ID, Properties.getConnectionId());
-    if (throwable.getCause() != null) {
-      detail.put(ClientMessageDetailKeys.CAUSE, getLoggingString(throwable));
-      if (throwable.getCause().getMessage() != null) {
-        detail.put(ClientMessageDetailKeys.CAUSE_MESSAGE, throwable.getCause().getMessage());
+
+    Throwable cause = throwable.getCause();
+    if (cause != null && cause instanceof InvocationTargetException) {
```
jmkulwik, line 20: Mind adding a comment explaining why we do this?

PR #206, opened 2022-02-08
Add Validation protocol

molly-moen:
Add `ValidationProtocol` class, which for now just supports running the project's main method. Also implements `NeighborhoodTestRunner.run()`, which levelbuilders can use to run neighborhood projects. To make this change work I had to make a few other updates:
- Factor out the method that finds the main method in a project, as well as a convenience method for throwing an exception if the user tries to load an unallowed class.
- Add `ValidationRuntimeException` which will be thrown if there is an issue running the main method.
- Update `JavabuilderRuntimeException` to allow setting and getting of the fallback message, so we can show the fallback message to unit test results instead of sending over the whole exception message.

For now none of the unit test output is translatable. If we hit an exception while running the student main method we will either show them a "no main method" message or an "exception thrown" message, with the name of the exception if there is one. This will lead to some non-ideal messages for our internal messages for which we have specialized messages on the frontend (such as their painter did an invalid move). For now it will just say "A NeighborhoodRuntimeException was thrown in your main method". This will be improved as a [follow-up](https://codedotorg.atlassian.net/browse/JAVA-437), but was out of scope for this specific change.


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/TestRunner.java
```diff
@@ -68,4 +73,15 @@ public void run(URLClassLoader urlClassLoader) throws InternalServerError {
       throw new InternalServerError(InternalErrorKey.INTERNAL_EXCEPTION, e);
     }
   }
+
+  private void setUpForValidation(URLClassLoader urlClassLoader) {
+    Method mainMethod = null;
+    try {
+      mainMethod = Util.findMainMethod(urlClassLoader, this.javaFiles);
+    } catch (UserInitiatedException e) {
+      // User has no main method. This may be ok since we are only running tests, so we ignore this.
```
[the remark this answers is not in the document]
molly-moen: makese sense, I'll update findMainMethod to return null instead

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavabuilderTestExecutionListener.java
```diff
@@ -101,13 +102,20 @@ public void executionFinished(
    */
   private String getErrorMessage(Throwable throwable, String className) {
     final String errorLine = this.errorLine(throwable.getStackTrace(), className);
-    String errorMessage;
+    String errorMessage = null;
 
     // If there was an assertion failure, print the failure message. If an exception was thrown,
     // print the exception name.
```
sanchitmalhotra126, line 108: nit: mind updating the comment for the new logic?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/Util.java
```diff
@@ -2,18 +2,25 @@
...
+import java.util.List;
 import org.code.protocol.InternalErrorKey;
 import org.code.protocol.LoggerUtils;
 
 /** A set of static utility functions that are used in multiple locations */
 public class Util {
+  private Util() {
+    throw new UnsupportedOperationException("Instantiation of utility class is not allowed.");
+  }
+
```
sanchitmalhotra126, line 23: nice! can we make this a `final` class as well while we're here?

org-code-javabuilder/validation/src/main/java/org/code/validation/support/ValidationRuntimeException.java
```diff
@@ -0,0 +1,35 @@
...
+
+public class ValidationRuntimeException extends JavabuilderRuntimeException {
+  protected ValidationRuntimeException(ExceptionKey key) {
+    super(key);
+    this.setFallbackMessage(this.generateFallbackMessage(key, null));
+  }
+
+  protected ValidationRuntimeException(ExceptionKey key, Throwable cause) {
+    super(key, cause);
+    this.setFallbackMessage(this.generateFallbackMessage(key, cause));
```
sanchitmalhotra126, line 13: Could we just add a constructor to `JavabuilderRuntimeException` that takes `key`, `cause`, and `fallbackMessage` so we could do `super(key, cause, this.generateFallbackMessage(key, cause))`? That way we don't have to expose `setFallbackMessage` on JavabuilderRuntimeException?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/Util.java
```diff
@@ -2,18 +2,25 @@
```
jmkulwik, line 2: Should this be split into two separate util classes? The two sets of functions seem pretty different. (not a soft ask - curious if others have opinions)

org-code-javabuilder/validation/src/main/java/org/code/validation/support/ExceptionKey.java
```diff
@@ -0,0 +1,8 @@
+package org.code.validation.support;
+
+public enum ExceptionKey {
+  // The user's main method could not be run
+  ERROR_RUNNING_MAIN,
+  // The user has no main method and we tried to run main
+  NO_MAIN_METHOD
```
jmkulwik, line 7: Is this the same as the exception key for a missing main method when running code? How do we differentiate between the two?

org-code-javabuilder/validation/src/main/java/org/code/validation/support/ValidationRuntimeException.java
molly-moen: unfortunately we can't because you have to call super() before anything else. Since the method call for the parameter gets called before super() actually gets called you get a compilation error.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/Util.java
molly-moen: I debated this too. I saw two existing categories of util functions in this class before I added new ones: methods to deal with jars and a clean up method (`recursivelyClearDirectory`). Since the jar methods are used for compilation it made some sense to put the main method search functionality in here too, but I am open to suggestion. I think if we moved out the new methods it may also make sense to make a separate `JarUtil` class as well.

org-code-javabuilder/validation/src/main/java/org/code/validation/support/ValidationRuntimeException.java
sanchitmalhotra126: Ahh right, I wasn't sure if calling the method inline would work, but yeah makes sense.

org-code-javabuilder/validation/src/main/java/org/code/validation/support/ExceptionKey.java
molly-moen: right now we differentiate because `ValidationRuntimeException` is what gets sent this enum value, and it translates it into a separate error message. But it might make sense to make it distinct so that we can add errors on the Java Lab side when we add translation here.

molly-moen: updated to a different enum value

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/Util.java
molly-moen: done!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavabuilderTestExecutionListener.java
molly-moen: done!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/Util.java
sanchitmalhotra126: Yeah, it seems like this file may have been created at a time when we didn't have a lot of util classes, but given that we now have several specialized util classes, might be worth renaming and/or breaking this up? I'd be okay with that being a separate task though.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/Util.java
```diff
@@ -56,4 +63,53 @@ public static void recursivelyClearDirectory(Path directory) throws IOException
     Files.walk(directory).sorted(Comparator.reverseOrder()).map(Path::toFile).forEach(File::delete);
     LoggerUtils.sendClearedDirectoryLog(directory);
   }
+
+  public static void throwInvalidClassException(NoClassDefFoundError e)
```
sanchitmalhotra126, line 67: nit: could this be potentially a little more descriptive, like `convertAndThrowInvalidClassException`? that's a little long, so maybe not necessary, but I'll leave it up to you.

sanchitmalhotra126 (approved):
LGTM!

jmkulwik (approved):
LGTM!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/Util.java
molly-moen: I'll do this as a follow up

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/Util.java
molly-moen: done!

PR #207, opened 2022-02-09
Unit tests: use display name for class

molly-moen:
In unit tests results, use the display name value instead of the raw class name. If the unit test class doesn't have a display name, the display name function defaults to the class name. This allows clearer messaging for validation tests, and enables students to use display names on their own unit tests as well.

[Jira Ticket](https://codedotorg.atlassian.net/browse/JAVA-438)

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavabuilderTestExecutionListener.java
+++ b/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavabuilderTestExecutionListener.java
@@ -7,6 +7,7 @@ import org.junit.platform.engine.TestExecutionResult;
 import org.junit.platform.engine.TestSource;
 import org.junit.platform.engine.support.descriptor.MethodSource;
 import org.junit.platform.launcher.TestIdentifier;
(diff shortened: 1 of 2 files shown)


sanchitmalhotra126 (approved):

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavabuilderTestExecutionListener.java
```diff
@@ -24,12 +25,27 @@
...
     this.outputAdapter = outputAdapter;
   }
 
+  /**
+   * Called when the execution of the TestPlan has started, before any test has been executed.
+   *
+   * <p>See:
+   * https://junit.org/junit5/docs/5.0.0/api/org/junit/platform/launcher/listeners/SummaryGeneratingListener.html#testPlanExecutionStarted-org.junit.platform.launcher.TestPlan-
+   *
+   * @param testPlan describes the tree of tests about to be executed
```
jmkulwik, line 41: Thanks. That answered exactly the question I was wondering :)

PR #209, opened 2022-02-11
Refactor Util

molly-moen:
Our Util class in `lib` contained a wide variety of util functions. We also already had a separate `FileNameUtils` class in `lib` which only contained one function. I added a new `util` package under `lib` and refactored our util classes as follows:
- `FileNameUtils` -> `FileUtils`, and it now also contains `recursivelyClearDirectory` (previously in `Util`)
- new class `JarUtils` which contains all jar-related helpers
- new class `ProjectLoadUtils` which contains our helpers used in loading projects (find the main method and throw the correct exception in the case of an unallowed class)

As part of this work I also made all the util classes final and gave them private constructors that throw an exception if they are used.


sanchitmalhotra126 (approved):
Nice refactor!

molly-moen:
> LGTM!
> 
> Side note -- is this a pattern we should use in other packages too (eg, [protocol](https://github.com/code-dot-org/javabuilder/blob/main/org-code-javabuilder/protocol/src/main/java/org/code/protocol/JavabuilderThrowableMessageUtils.java))?
> 
> And if we had a `util` sub-package elsewhere, would is be parallel to `support` where we've added that? Maybe `support` makes more sense as a sub-package in the context of a student-facing package.

@bencodeorg good questions. I mostly made this a subpackage so that all the util classes were in one place and easily findable. If there's only one util class it feels less necessary but if we want to adopt that pattern we can. As to your second question, it would probably make sense to make it parallel, although if the util functions were only used in the support code then it would make sense for the package to be inside support.

PR #210, opened 2022-02-11
Use ClientMessageDetailKeys in Neighborhood

sanchitmalhotra126:
Quick prefactor for https://codedotorg.atlassian.net/browse/JAVA-412 - with this, PainterTracker can use the same keys to pull values out of the details objects.


molly-moen (approved):
Nice!

PR #211, opened 2022-02-12
Add ContentManager interface

bencodeorg:
Adds an (unused) `ContentManager` interface for the decoupling Javabuilder work to the `protocol` package. Straight copy from [the prototype branch](https://github.com/code-dot-org/javabuilder/pull/194/files).

The prototype added this interface to `org.code.javabuilder`, which is supposed to hold AWS-specific implementations, so moved to protocol.


org-code-javabuilder/protocol/src/main/java/org/code/protocol/ContentManager.java
```diff
@@ -0,0 +1,10 @@
+package org.code.protocol;
+
+public interface ContentManager {
```
sanchitmalhotra126, line 3: Nice! Can we add a quick javadoc here? something like "Manages content (such as code sources and assets) that are used and generated during a Javabuilder session"

sanchitmalhotra126 (approved):

PR #212, opened 2022-02-14
Create NeighborhoodTracker and implement PainterTracker

sanchitmalhotra126:
Creates/implements the NeighborhoodTracker and PainterTracker classes that track neighborhood events across a validation run. I made few small changes in the existing code as well so I just want to make sure there weren't any issues with these:
- Make `Grid#getSize()` public (instead of protected) for access in the validation package
- Make the `NeighborhoodSignalMessage` constructor public (instead of protected) so it could be instantiated in the validation package
- Add the amount of paint to the initialization details that are sent with the INITIALIZE_PAINTER message - Javalab wouldn't need this and should just ignore it, but it's needed for us to know how much paint the painter started out with.
- Make `NeighborhoodActionTypeMapper#convertNeighborhoodKeyToActionType` a static method.

JIRA: https://codedotorg.atlassian.net/browse/JAVA-412

Testing: unit tests


org-code-javabuilder/validation/src/main/java/org/code/validation/support/NeighborhoodTracker.java
```diff
@@ -0,0 +1,98 @@
...
+      final int paint = message.getDetail().getInt(PAINT);
+      final PainterTracker painterTracker = new PainterTracker(id, new Position(x, y), paint);
+      this.painterTrackers.put(id, painterTracker);
+      return;
+    }
+
+    if (!this.painterTrackers.containsKey(id) || !this.isInitialized) {
+      System.out.printf(
+          "Received a painter event for an uninitialized painter. Ignoring message: %s, %s\n",
+          message.getValue(), message.getDetail());
```
sanchitmalhotra126, line 58: Wanted to ask about this - should we throw an exception instead?

molly-moen: this would indicate something wrong with our initialization code--a student shouldn't be able to trigger painter events without first initializing a painter and calling events on it. I would maybe have this be an error log so we get notified something weird is happening.

sanchitmalhotra126: Cool makes sense. Do you think we should throw as well so the validation run stops, or should we continue with validation?

molly-moen (approved):
Looks great!

org-code-javabuilder/validation/src/main/java/org/code/validation/support/NeighborhoodTracker.java
molly-moen: Either way would be fine. If we had this bug on our end the student also would see weird behavior when they tried to run their code (because we would be trying to move/paint with an nonexistent painter) so I think we can just assume we won't see an event for a painter that doesn't exist. I would lean towards continue with validation in case things can pass, so that students don't see a bug on our end that doesn't impact them.

PR #213, opened 2022-02-15
Send test result messages as key/value instead of string

molly-moen:
In order to utilize our custom error messages (such as `Painter tried to move off the grid or into an obstacle`) in unit test results we need to send results as key/value pairs instead of plain strings. Previously we generated strings with check mark/x icons and success/fail messages, as well as basic exception messages. Now we send the messages in pieces and allow the frontend to display them how it chooses. For example, instead of the backend sending the string `✔ MyTestClass > myTest SUCCEEDED`, it will instead send an object:
```
{
  "status": "SUCCEEDED",
  "className": "MyTestClass",
  "methodName": "myTest",
  "isValidation": false 
}
```

For failed tests, we also send a status detail message which can include the assertion error, a custom error message key that is known by Java Lab, or an exception name and stack trace.

This work allows unit tests results to be mostly translatable, with the exception of stack traces (they will contain 'at') and assertion errors (they will be whatever is generated by JUnit).

This change comes with an associated [Java Lab change](https://github.com/code-dot-org/code-dot-org/pull/44841). We will want to merge the code.org changes first to avoid test results being empty for a long time.


jmkulwik (approved):
Nice work!

org-code-javabuilder/protocol/src/main/java/org/code/protocol/JavabuilderThrowableMessageUtils.java
```diff
@@ -67,7 +71,13 @@ private static String getDefaultFallbackMessageString(Throwable cause) {
     return loggingString;
   }
 
-  private static String getUserFacingStackTraceString(Throwable cause) {
+  /**
```
bencodeorg, line 74: Thanks for adding this :)

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavabuilderTestExecutionListener.java
```diff
@@ -93,66 +90,64 @@ public void executionFinished(
...
-    }
-    if (errorMessage == null) {
-      String additionalMessage = this.getAdditionalErrorMessage(throwable);
-      errorMessage =
-          String.format("\t%s thrown at %s\n", this.getThrowableDisplayName(throwable), errorLine);
-      if (additionalMessage != null) {
-        errorMessage = String.format("%s \t%s\n", errorMessage, additionalMessage);
-      }
+      errorDetails.putAll(exception.getExceptionDetails());
+      errorDetails.put(ClientMessageDetailKeys.TYPE, exception.getMessage());
```
bencodeorg, line 141: For my understanding, how is this different from what would be included in the `CAUSE_MESSAGE` key in the throwable util helper?

bencodeorg (commented):
Big improvement!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavabuilderTestExecutionListener.java
molly-moen: The 'message' is the enum key we assign to each `JavabuilderRuntimeException`. When we send the exception as a `JavabuilderThrowableMessage` the key gets sent separately from the details, so we need to add it in here so we can get it on the Java Lab side.

molly-moen: for example, the key is something like `FILE_NOT_FOUND`

bencodeorg: mmm not sure I'm following, I think I was asking a different question? I guess I'm confused because we already call `getMessage()` when you call `exception.getExceptionDetails()` and put it in the details hash under the `CAUSE_MESSAGE` key:

https://github.com/code-dot-org/javabuilder/blob/c05d467c9d4a8274383e5def7c6819881d3e8cf3/org-code-javabuilder/protocol/src/main/java/org/code/protocol/JavabuilderThrowableMessageUtils.java#L19-L54

Maybe the difference is that you're calling `getMessage()` on the `JavabuilderRuntimeException` itself, rather than its cause? Our exception hierarchy makes my brain hurt, so still trying to wrap (no pun intended) my head around it, so I'm probably just misunderstanding something...

molly-moen: Yep the difference is that one is on the cause, the other on the exception itself. The cause can either be any throwable or null, depending on if [this constructor](https://github.com/code-dot-org/javabuilder/blob/c05d467c9d4a8274383e5def7c6819881d3e8cf3/org-code-javabuilder/protocol/src/main/java/org/code/protocol/JavabuilderRuntimeException.java#L16) is used or not (if it's not, the cause will be null). The message for the `JavabuilderRuntimeException` will always a short constant value. For example, if it's a `NeighborhoodRuntimeException` it will always be one of [these](https://github.com/code-dot-org/javabuilder/blob/c05d467c9d4a8274383e5def7c6819881d3e8cf3/org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/ExceptionKeys.java#L3).

bencodeorg: Ahh got it got it, I was looking for an override for getMessage or something in JavabuilderRuntimeException, guess that's just default behavior for RuntimeException. Thanks for explaining!

image

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavabuilderTestExecutionListener.java
```diff
@@ -22,15 +23,14 @@
  * https://junit.org/junit5/docs/5.0.0/api/org/junit/platform/launcher/TestExecutionListener.html
  */
 public class JavabuilderTestExecutionListener extends SummaryGeneratingListener {
-  private static final String CHECK_MARK = "✔";
-  private static final String HEAVY_X = "✖";
-
   private final OutputAdapter outputAdapter;
   private TestPlan testPlan;
+  private boolean isValidation;
```
sanchitmalhotra126, line 28: nit: final?

org-code-javabuilder/lib/src/test/java/org/code/javabuilder/JavabuilderTestExecutionListenerTest.java
```diff
@@ -70,9 +69,12 @@ public void testExecutionFinishedSendsStatusMessage() {
 
     final ClientMessage message = messageCaptor.getAllValues().get(0);
     assertEquals(ClientMessageType.TEST_RESULT, message.getType());
-    assertTrue(message.getValue().contains(displayName));
-    assertTrue(message.getValue().contains(classDisplayName));
-    assertTrue(message.getValue().contains(TestExecutionResult.Status.SUCCESSFUL.toString()));
+
+    JSONObject messageDetail = message.getDetail();
+    assertTrue(messageDetail.getString("methodName").equals(displayName));
```
sanchitmalhotra126, line 74: nit: use ClientMessageDetailKeys.METHOD_NAME, etc for these tests??

org-code-javabuilder/validation/src/main/java/org/code/validation/support/UserTestResultSignalKey.java
```diff
@@ -0,0 +1,6 @@
+package org.code.validation.support;
+
+public enum UserTestResultSignalKey {
+  TEST_STATUS,
```
sanchitmalhotra126, line 4: nit: should this be TEST_RESULT? doesn't really matter, but just thinking of a case where we would need to send status messages as a test is happening separately from the test results

sanchitmalhotra126 (approved):
Nice! A few nits but nothing major

org-code-javabuilder/validation/src/main/java/org/code/validation/support/UserTestResultSignalKey.java
molly-moen: I'm going to keep it as is for 2 reasons: the java `TestExecutionResult` uses "status" to refer to success/failure/aborted, and I already merged the Java Lab change :)

org-code-javabuilder/lib/src/test/java/org/code/javabuilder/JavabuilderTestExecutionListenerTest.java
molly-moen: good call

PR #215, opened 2022-02-21
Implement local content manager

sanchitmalhotra126:
First part of JAVA-449. This sets up the LocalContentManager which fetches the project JSON payload from local storage and parses it. The actual parsing and reading is handled by a separate class, ProjectData.java, so it can be reused by AWSContentManager since the JSON-level operations will be the same. As of now, this content manager is not hooked up to anything so it should not result in any changes. Also, some of LocalContentManager's methods like generateAssetUploadUrl and writeToOutputFile are very similar or the same as the existing LocalFileManager methods.

JIRA: https://codedotorg.atlassian.net/browse/JAVA-449

Tested by manually seeding local storage with a JSON file and parsing sources and assets data via print statements, as well as unit tests.

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalContentManager.java
+++ b/org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalContentManager.java
@@ -36,11 +36,15 @@ public class LocalContentManager implements ContentManager, ProjectFileLoader {
   }
 
   @Override
-  public String generateAssetUrl(String filename) {
+  public String getAssetUrl(String filename) {
     try {
       this.loadProjectDataIfNeeded();
     } catch (InternalServerError e) {
-      // Log this error since this indicates an issue reading project data
+      // We should only hit this exception if we try to load an asset URL before source code has
+      // been loaded, which should only be in the the case of manual testing. Log this exception but
+      // don't throw to preserve the method contract.
+      // Note / TODO: Once we fully migrate away from Dashboard sources, we can remove the
+      // loadProjectDataIfNeeded() call here and this exception handling.
       LoggerUtils.logException(e);
       return null;
     }
@@ -51,7 +55,9 @@ public class LocalContentManager implements ContentManager, ProjectFileLoader {
   public String generateAssetUploadUrl(String filename) throws InternalServerError {
     this.loadProjectDataIfNeeded();
     final String uploadUrl = String.format(SERVER_URL_FORMAT, DIRECTORY, filename);
-    // Add new asset URL to our project data so it can be referenced later.
+    // The URL to GET this asset once it is uploaded will be the same as the upload
+    // URL since it points to the same location on the local file server. Add this
+    // URL to our project data so it can be referenced later.
     this.projectData.addNewAssetUrl(filename, uploadUrl);
     return uploadUrl;
   }
--- a/org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalContentManager.java
+++ b/org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalContentManager.java
@@ -13,6 +13,7 @@ import org.code.protocol.ContentManager;
 import org.code.protocol.InternalErrorKey;
 import org.code.protocol.JavabuilderException;
 import org.code.protocol.LoggerUtils;
+import org.json.JSONException;
 
 public class LocalContentManager implements ContentManager, ProjectFileLoader {
   private static final String SERVER_URL_FORMAT = "http://localhost:8080/%s/%s";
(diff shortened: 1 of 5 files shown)


org-code-javabuilder/lib/src/test/java/org/code/javabuilder/ProjectDataTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.Mockito.*;
+
+import org.code.protocol.InternalErrorKey;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentCaptor;
+
+// TODO: Test validation JSON when available
```
molly-moen, line 11: what info do you need for validation json? You can get sample validation json from all the things java lab level 12

org-code-javabuilder/protocol/src/main/java/org/code/protocol/ContentManager.java
```diff
@@ -1,14 +1,48 @@
...
+  /**
+   * Generates a URL pointing to the asset file referenced by the asset file name.
+   *
+   * @param filename asset file name to generate a URL for
+   * @return generated URL
+   */
+  String generateAssetUrl(String filename);
 
+  /**
+   * Generates a URL that a client may use to upload a specific asset named by the provided file
```
molly-moen, line 19: is the url a post or put? Would be good to be explicit in the comment so the contract is clear.

org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalContentManager.java
```diff
@@ -0,0 +1,94 @@
...
+      return null;
+    }
+    return this.projectData.getAssetUrl(filename);
+  }
+
+  @Override
+  public String generateAssetUploadUrl(String filename) throws InternalServerError {
+    this.loadProjectDataIfNeeded();
+    final String uploadUrl = String.format(SERVER_URL_FORMAT, DIRECTORY, filename);
+    // Add new asset URL to our project data so it can be referenced later.
```
molly-moen, line 54: I got a bit confused by us putting the upload url in the list of urls we use for get. Looks like in this case it's the same url, mind adding a comment explaining that?

molly-moen (commented):
Looks good! Just a couple questions/comments

org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalContentManager.java
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  @Override
+  public UserProjectFiles loadFiles() throws InternalServerError, UserInitiatedException {
+    this.loadProjectDataIfNeeded();
+    return this.projectData.getSources();
+  }
+
+  @Override
+  public String generateAssetUrl(String filename) {
```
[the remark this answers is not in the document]
sanchitmalhotra126: Good call - I named it like this to mirror the existing `generateAssetUrl` in GlobalProtocol, but yeah I think it makes more sense to rename it `getAssetUrl` here since it won't be exposed to other classes anyway, but only called via GlobalProtocol for now.

org-code-javabuilder/lib/src/test/java/org/code/javabuilder/ProjectDataTest.java
sanchitmalhotra126: Oh good to know! Wasn't sure if I could get validation data from a level yet, so I'll add test cases for that.

org-code-javabuilder/protocol/src/main/java/org/code/protocol/ContentManager.java
sanchitmalhotra126: It's a PUT - I'll update the comment.

org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalContentManager.java
sanchitmalhotra126: Yep, I'll add some more context in the comments. Since this is all on the local fileserver, the PUT url becomes the same as the GET url once the object is uploaded.

org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalContentManager.java
```diff
@@ -0,0 +1,94 @@
...
+    try {
+      this.loadProjectDataIfNeeded();
+    } catch (InternalServerError e) {
+      // Log this error since this indicates an issue reading project data
```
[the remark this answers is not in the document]
sanchitmalhotra126: Hmm yeah, so this is a little weird right now. I didn't add any exceptions to the method header here to be consistent with the existing contract for GlobalProtocol#generateAssetUrl. Ideally, once we fully migrate to using this class, the JSON file loading will happen in the constructor, or in some initialization step, before we call `generateAssetUrl`, so calling this method will be safe. For now however, we'd hit this exception if we somehow call `generateAssetUrl` before `loadFiles` is called, which would only happen if we call these methods out of order manually. I'll add to the comment to clarify this.

molly-moen (approved):
Nice!

org-code-javabuilder/lib/src/test/java/org/code/javabuilder/ProjectDataTest.java
sanchitmalhotra126: I'll go ahead and remove this TODO for now - I confirmed locally that validation JSON is present, but we can defer testing against expected output until we need to pull out validation code separately from source code.

PR #217, opened 2022-02-23
Implement AWSContentManager

sanchitmalhotra126:
Adds the Lambda implementation of the ContentManager interface. Not really much new content here - the project data getter methods (loadFiles, getAssetUrl, verifyAssetFileName) are exactly the same as LocalContentManager, and the generateAssetUploadUrl/writeToOutputFile methods are pretty much copies from the existing methods in AWSFileManager. The only difference is how AWSContentManager fetches the JSON file (in `loadProjectDataIfNeeded()`). This also includes a few minor fixes to LocalContentManager & LocalContentManagerTest.

Testing - tested by manually seeding S3 bucket with a JSON payload and confirming that all the correct data is read out and parsed + unit tests.

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSContentManager.java
+++ b/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSContentManager.java
@@ -28,7 +28,7 @@ public class AWSContentManager implements ContentManager, ProjectFileLoader {
   private final String bucketName;
   private final AmazonS3 s3Client;
   private final String javabuilderSessionId;
-  private final String getContentUrl;
+  private final String contentBucketUrl;
   private final Context context;
   private ProjectData projectData;
   private int writes;
@@ -38,22 +38,22 @@ public class AWSContentManager implements ContentManager, ProjectFileLoader {
       AmazonS3 s3Client,
       String bucketName,
       String javabuilderSessionId,
-      String getContentUrl,
+      String contentBucketUrl,
       Context context) {
-    this(s3Client, bucketName, javabuilderSessionId, getContentUrl, context, null);
+    this(s3Client, bucketName, javabuilderSessionId, contentBucketUrl, context, null);
   }
 
   AWSContentManager(
       AmazonS3 s3Client,
       String bucketName,
       String javabuilderSessionId,
-      String getContentUrl,
+      String contentBucketUrl,
       Context context,
       ProjectData projectData) {
     this.bucketName = bucketName;
     this.s3Client = s3Client;
     this.javabuilderSessionId = javabuilderSessionId;
-    this.getContentUrl = getContentUrl;
+    this.contentBucketUrl = contentBucketUrl;
     this.context = context;
     this.projectData = projectData;
     this.writes = 0;
@@ -96,8 +96,8 @@ public class AWSContentManager implements ContentManager, ProjectFileLoader {
               this.bucketName, key, new Date(expirationTimeMs), HttpMethod.PUT);
       this.uploads++;
       // Add the GET url for this file to the asset map so it can be referenced later.
-      this.projectData.addNewAssetUrl(filename, this.getContentUrl + "/" + key);
-      return this.getContentUrl + presignedUrl.getFile();
+      this.projectData.addNewAssetUrl(filename, this.contentBucketUrl + "/" + key);
+      return this.contentBucketUrl + presignedUrl.getFile();
     } catch (AbortedException e) {
       // this is most likely because the end user interrupted program execution. We can safely
       // ignore this.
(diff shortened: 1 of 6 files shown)


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSContentManager.java
```diff
@@ -0,0 +1,183 @@
...
+    } catch (AbortedException e) {
+      // this is most likely because the end user interrupted program execution. We can safely
+      // ignore this.
+    } catch (SdkClientException e) {
+      // We couldn't write to S3, send a message to the user and fail. The S3 SDK includes retries.
+      throw new InternalServerError(InternalErrorKey.INTERNAL_RUNTIME_EXCEPTION, e);
+    }
+
+    this.writes++;
+    return this.getContentUrl + "/" + filePath;
```
jmkulwik, line 134: What is the `getContentUrl`? I was assuming it was the URL used to GET content from S3, but it looks like we're also using it for PUTs too. If that's the case, perhaps just `contentUrl` would be clearer?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSContentManager.java
```diff
@@ -0,0 +1,183 @@
...
+      this.loadProjectDataIfNeeded();
+    } catch (InternalServerError e) {
+      throw new FileNotFoundException("Error loading data");
+    }
+    if (!this.projectData.doesAssetUrlExist(filename)) {
+      throw new FileNotFoundException(filename);
+    }
+  }
+
+  private String generateKey(String filename) {
```
jmkulwik, line 154: Can you add a comment here? I think I know what key it's generating, but I'm pretty sure me in a few weeks will have forgotten enough context to get confused :)

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSContentManager.java
sanchitmalhotra126: Yep good call - this is copied over from when we were only using the URL to GET theater files. I'll change to `contentUrl`

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSContentManager.java
```diff
@@ -0,0 +1,183 @@
...
+import org.code.protocol.ContentManager;
+import org.code.protocol.InternalErrorKey;
+import org.code.protocol.JavabuilderException;
+import org.code.protocol.LoggerUtils;
+
+public class AWSContentManager implements ContentManager, ProjectFileLoader {
+  // Temporary limit on writes per session until we can more fully limit usage.
+  private static final int WRITES_PER_SESSION = 2;
+  // Limit on the amount of S3 file uploads per session to avoid spamming
+  private static final int UPLOADS_PER_SESSION = 20;
```
jmkulwik, line 26: Can you add a bit more detail to these two comments - either here or by the equivalent exception in `UserInitiatedExceptionKey` (or both) - I was having a hard time differentiating why one or the other might be used until I realized that one gets used for prompter and the other gets used by theater. I also noticed that the exceptions don't get interpreted on the Java Lab side. Mind adding either a fallback message or a Java Lab message for those exceptions?

jmkulwik: (The exceptions are TOO_MANY_UPLOADS and TOO_MANY_WRITES, below)

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSContentManager.java
jmkulwik: Ahh, I'm also realizing I was reading all this in terms of getting the user sources from S3, and didn't realize these were updating existing functionality from when we added Theater and Prompter. `getContentUrl` does make a lot more sense now.

jmkulwik (approved):
LGTM! Just a few comment requests.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSContentManager.java
sanchitmalhotra126: Add comments and a fallback message for TOO_MANY_UPLOADS. For TOO_MANY_WRITES, I think it makes sense to display the standard error message for now ("Oops. We hit an error on our side...") because a student should only hit that error under very strange circumstances, such as Stage.play() is somehow called more than once per session without throwing and exception, or if they're calling AWS APIs directly from their code somehow (which shouldn't be possible with our custom class loader). In either case i'd imagine we'd want to investigate in more detail on our end.

PR #218, opened 2022-02-23
Integrate content manager, gated behind useDashboardSources flag

sanchitmalhotra126:
This change hooks up the ContentManager implementations to be used via GlobalProtocol. The logic to decide which code path to use (dashboard sources or S3/local sources) is abstracted behind two new classes, DelegatingAssetFileHelper and DelegatingFileManager and driven by the `useDashboardSources` flag that's sent from Javalab ([here](https://github.com/code-dot-org/code-dot-org/blob/d98116605c7ce08abbdc96ca6c0098df7dab9977/apps/src/javalab/JavabuilderConnection.js#L62)). This is so that external callers (Image, AudioUtils, Stage, Board, etc) don't have to change to any method contracts. Once we're fully migrated to the new code path, we can remove these abstractions and clean up the APIs a bit.

JIRA: https://codedotorg.atlassian.net/browse/JAVA-449

Testing: tested the following scenarios end-to-end and confirmed that projects execute successfully
- localhost Dashboard -> localhost Javabuilder (both using dashboard sources and not using dashboard sources)
- staging Dashboard -> deployed Javabuilder via Postman w/ prototype API Gateway (using dashboard sources)
- localhost Dashboard -> deployed Javabuilder w/ prototype API Gateway (not using dashboard sources). This worked successfully for Neighborhood, Console, Playground, and Theater projects that only use prompter images 🎉

### DIFF
--- a/org-code-javabuilder/protocol/src/main/java/org/code/protocol/DelegatingAssetFileHelper.java
+++ b/org-code-javabuilder/protocol/src/main/java/org/code/protocol/DelegatingAssetFileHelper.java
@@ -1,0 +1,40 @@
+package org.code.protocol;
+
+import java.io.FileNotFoundException;
+
+/**
+ * An {@link AssetFileHelper} that delegates to either the standard AssetFileHelper implementations
+ * of its methods, or to a {@link ContentManager}. Used to temporarily support two code paths.
+ */
+public class DelegatingAssetFileHelper extends AssetFileHelper {
+
+  private final ContentManager contentManager;
+  private final boolean useDashboardSources;
+
+  public DelegatingAssetFileHelper(
+      String dashboardHostname,
+      String channelId,
+      String levelId,
+      ContentManager contentManager,
+      boolean useDashboardSources) {
+    super(dashboardHostname, channelId, levelId);
+    this.contentManager = contentManager;
+    this.useDashboardSources = useDashboardSources;
+  }
+
+  @Override
+  public String generateAssetUrl(String filename) {
+    return this.useDashboardSources
+        ? super.generateAssetUrl(filename)
+        : this.contentManager.getAssetUrl(filename);
+  }
+
+  @Override
+  public void verifyAssetFilename(String filename) throws FileNotFoundException {
+    if (this.useDashboardSources) {
+      super.verifyAssetFilename(filename);
+    } else {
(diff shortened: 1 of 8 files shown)


org-code-javabuilder/protocol/src/main/java/org/code/protocol/DelegatingAssetFileHelper.java
```diff
@@ -0,0 +1,40 @@
+package org.code.protocol;
+
+import java.io.FileNotFoundException;
+
+/**
+ * An {@link AssetFileHelper} that delegates to either the standard AssetFileHelper implementations
+ * of its methods, or to a {@link ContentManager}. Used to temporarily support two code paths.
+ */
+public class DelegatingAssetFileHelper extends AssetFileHelper {
```
molly-moen, line 9: do we have a work item to remove this and `DelegatingFileManager` once they are no longer needed?

org-code-javabuilder/protocol/src/testFixtures/java/org/code/protocol/GlobalProtocolTestFactory.java
```diff
@@ -61,6 +65,16 @@ public Builder withLifecycleNotifier(LifecycleNotifier lifecycleNotifier) {
       return this;
     }
 
+    public Builder withContentManager(ContentManager contentManager) {
```
molly-moen, line 68: is this method used anywhere?

molly-moen (approved):
Nice solution!

org-code-javabuilder/protocol/src/main/java/org/code/protocol/DelegatingAssetFileHelper.java
sanchitmalhotra126: We have [this](https://codedotorg.atlassian.net/browse/JAVA-451) task to generally cleanup unused code after migrating - I'll mention it there.

org-code-javabuilder/protocol/src/testFixtures/java/org/code/protocol/GlobalProtocolTestFactory.java
sanchitmalhotra126: Not currently - just added it in case unit tests would need to mock this for any reason.

PR #219, opened 2022-02-25
Add org.code.lang.System

molly-moen:
Add `org.code.lang` package and new `org.code.lang.System` class that only includes the System methods we want to expose to end users. We will import this class by default into every user file. End users should not see a change for normal usage of System. The differences they will now see are:

- If they attempt to use an unapproved System method they will get the following error:
 ```
/MyClass.java:11: error: cannot find symbol
    System.getEnv();
          ^
  symbol:   method getEnv()
  location: class org.code.lang.System
```
- If they try to directly import `java.lang.System` they will see `Import of java.lang.System is not supported.`

## Links
- [Doc](https://docs.google.com/document/d/1C19rFe6hl7h1V14OStg4N8bk43Bmtys6mqjz6GaviDM/edit#)
- [Jira](https://codedotorg.atlassian.net/browse/JAVA-452)

### CODE REVIEW

File: org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaProjectFile.java [hunk 1]
```diff
@@ -43,10 +43,14 @@ public void setFileName(String fileName) {
...
-    this.fileContents = fileContents;
+    this.fileContents = this.importSystem(fileContents);
   }
 
   public void setClassName(String className) {
     this.className = className;
   }
+
+  private String importSystem(String fileContents) {
+    return "import org.code.lang.System;\n" + fileContents;
```
sanchitmalhotra126, line 54: Should we check if this import already exists before adding it (especially in the case of `setFileContents`?)

sanchitmalhotra126 (commented):
Looks good! Just a couple questions.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaProjectFile.java
molly-moen: good idea

org-code-javabuilder/lang/src/main/java/org/code/lang/System.java
```diff
@@ -0,0 +1,29 @@
...
+import java.io.InputStream;
+import java.io.PrintStream;
+
+/**
+ * User-facing System class. We only support a few methods and fields from java.lang.System, so we
+ * wrap java.lang.System in org.code.lang.System.
+ */
+public class System {
+    public static InputStream in = java.lang.System.in;
+    public static PrintStream out = java.lang.System.out;
```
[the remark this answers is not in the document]
molly-moen: done, good catch

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaProjectFile.java
molly-moen: I added a check if the file contents start with the import we won't add it. I just learned you actually can have duplicate imports of the same package and Java doesn't care, so even if we end up double-importing student's won't see an issue. I wanted to only check startsWith in the case that for some reason the import is buried in a string somewhere.

sanchitmalhotra126 (approved):
LGTM!

molly-moen:
Note: waiting to merge this until an email goes out to teachers

org-code-javabuilder/lib/src/test/java/org/code/javabuilder/UserProjectFileParserTest.java
```diff
@@ -25,7 +25,6 @@ public void canParseValidFileJson() throws InternalServerError, UserInitiatedExc
     assertEquals(projectFiles.getJavaFiles().size(), 1);
     ProjectFile firstFile = projectFiles.getJavaFiles().get(0);
     assertEquals(firstFile.getFileName(), "HelloWorld.java");
-    assertEquals(firstFile.getFileContents(), expectedCode);
```
jmkulwik, line 28: Should this check still be here, but with the string prepended?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaProjectFile.java
```diff
@@ -43,10 +45,23 @@ public void setFileName(String fileName) {
...
   public void setClassName(String className) {
     this.className = className;
   }
+
+  private String importSystem(String fileContents) {
+    // Check if file contents start with system import only because system import should always
+    // be done by default by us, and therefore be the first thing in the file.
+    // We still want to import System in the rare case where SYSTEM_IMPORT is in
+    // a String somewhere in fileContents. In addition, a duplicate import will
+    // not cause any error to the student, it just makes the file longer.
```
jmkulwik, line 60: Why not just always prepend the import? i.e. even if it's already the first line?

org-code-javabuilder/lang/README.md
```diff
@@ -0,0 +1,13 @@
+# Lang
```
jmkulwik, line 1: 🎉 Thanks for adding a readme!!! Mind adding a note in the org-code-javabuilder readme too?  (can be done as a follow-up) https://github.com/code-dot-org/javabuilder/blob/main/org-code-javabuilder/README.md

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaProjectFile.java
molly-moen: There isn't much risk to always prepending, the thought here was not to do it if it's not strictly necessary in case we end up somehow calling `setFileContents` a bunch of times. But I can revert it back to just always adding it at the beginning.

org-code-javabuilder/lib/src/test/java/org/code/javabuilder/UserProjectFileParserTest.java
molly-moen: I removed it here because the import logic isn't related to `UserProjectFileParser` specifically. I added a test for the import statement in `JavaProjectFile` instead. I could instead check that the fileContents contain the expectedCode?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaProjectFile.java
molly-moen: thinking more on this, I'll just always add the import line, it's pretty low risk

org-code-javabuilder/lang/README.md
molly-moen: done, thanks for the reminder!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaProjectFile.java
molly-moen: ended up switching this back because it felt overly complicated for limited benefit.

jmkulwik: Will this cause compiler errors to always be off-by-one?

org-code-javabuilder/lang/src/main/java/org/code/lang/System.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * User-facing System class. We only support a few methods and fields from java.lang.System, so we
+ * wrap java.lang.System in org.code.lang.System.
+ */
+public class System {
+    public static final InputStream in = java.lang.System.in;
+    public static final PrintStream out = java.lang.System.out;
+
+    public static void arraycopy(Object src, int srcPos, Object dest, int destPos, int length) {
```
bencodeorg, line 14: Nit: `private System() {}` constructor? That's what it looks like is in real System. Not that there's anything you could do with a system object, but feels like something a student who is learning might try / might be good to be consistent with java.

molly-moen: good catch, added

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaProjectFile.java
molly-moen: added a fix for this!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserCodeCompiler.java
```diff
@@ -89,4 +92,107 @@ private CompilationTask getCompilationTask(DiagnosticCollector<JavaFileObject> d
...
   }
+
+  /**
+   * For specific compiler errors we want to return custom messages. Check if the given diagnostic
+   * is due to one of those errors and return the more specific message, or null if not.
+   *
+   * @param diagnostic
+   * @return custom compiler error or null
+   */
+  private String getCustomCompilerError(Diagnostic<? extends JavaFileObject> diagnostic) {
```
jmkulwik, line 103: Had a thought: Would it be better for us to implement Diagnostic?

molly-moen: Can you expand on that? The diagnostics are returned by the compiler

jmkulwik: Ah nevermind. I take that back. I was thinking it could be done with some sort of custom DiagnosticListener, but you're right, the compiler controls the diagnostic.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserCodeCompiler.java
```diff
@@ -89,4 +92,107 @@ private CompilationTask getCompilationTask(DiagnosticCollector<JavaFileObject> d
...
+    // Check if the compiler error is due to the student importing java.lang.System
+    // directly and give a more helpful message.
+    if (diagnostic.getCode().equals("compiler.err.already.defined.single.import")
+        && diagnostic.getMessage(Locale.US).contains("org.code.lang.System")) {
+      return "Import of java.lang.System is not supported.";
+    }
+    return null;
+  }
```
sanchitmalhotra126, line 111: nit: could pull this logic into `getCompilerError` so you don't need to call two different methods to get the right error message?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserCodeCompiler.java
```diff
@@ -89,4 +92,107 @@ private CompilationTask getCompilationTask(DiagnosticCollector<JavaFileObject> d
...
+   *     System.out.println(System.currenTimeMillis());
+   *                              ^
+   *   symbol:   method currenTimeMillis()
+   *   location: class org.code.lang.System
+   * </pre>
+   *
+   * @param diagnostic
+   * @return compiler error String
+   */
+  private String getCompilerError(Diagnostic<? extends JavaFileObject> diagnostic) {
```
sanchitmalhotra126, line 139: Might be nice to add a test case for this to make sure the format outputs correctly? But also looks like we don't have any unit tests for `UserCodeCompiler`, so that can also be a future tech debt item

sanchitmalhotra126 (approved):
Looks good!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserCodeCompiler.java
molly-moen: The reason we have two separate methods is if we get a result from this method we don't show any more compiler errors, but for the generic errors we show all of them.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserCodeCompiler.java
molly-moen: I'll add a tech debt item, I have a feeling it's a fairly big undertaking because we have to mock quite a few things.

molly-moen: https://codedotorg.atlassian.net/browse/JAVA-467

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserCodeCompiler.java
sanchitmalhotra126: Oh yep, I was just thinking we could early return the custom error message if the diagnostic matches the import error, and if not, return the full message? Maybe I'm misunderstanding though

molly-moen: They have slightly different behavior after we get the string. We get a list of Diagnostics from the compiler and go through them one by one in a for loop on line 41. If we see this specific message we don't want to show any more diagnostics in that list, because they will be confusing to students and all we want to show them is this no import message. But for the general case we do want to show every compiler error (for example, they referenced multiple methods incorrectly).

sanchitmalhotra126: Ahhh right sorry I missed the break statement :) thanks for clarifying!

PR #221, opened 2022-02-28
Rename S3 bucket

sanchitmalhotra126:
This renames the Javabuilder S3 bucket from cdo-javabuilder-output to cdo-javabuilder-content to more accurately reflect its use. Note: this PR should NOT be merged until this [policy change PR](https://github.com/code-dot-org/javabuilder/pull/220) has been merged AND deployed in the production account (see this [Slack thread](https://codedotorg.slack.com/archives/C03CK49G9/p1645815922432079) for details).

Testing pending updating the policy in the dev account and testing the name change on a deployed instance (will update this once testing is complete).

PR #220, opened 2022-02-25
Add new bucket name to IAM template

sanchitmalhotra126:
First step of renaming the Javabuilder S3 bucket from cdo-javabuilder-output to cdo-javabuilder-content. This updates the IAM permissions to grant access to the new bucket name.

JIRA: https://codedotorg.atlassian.net/browse/JAVA-440

PR #223, opened 2022-03-02
Remove old bucket permissions from IAM template

sanchitmalhotra126:
Cleanup step following https://github.com/code-dot-org/javabuilder/pull/220 and https://github.com/code-dot-org/javabuilder/pull/221. This removes permissions on the old bucket name (*output*) now that Javabuilder has been updated to reference the new bucket name.

https://codedotorg.atlassian.net/browse/JAVA-440

Tested in dev account with dev Javabuilder instance.

### DIFF
--- a/iam.yml
+++ b/iam.yml
@@ -39,12 +39,6 @@ Resources:
                   - "cloudformation:DescribeStacks"
                 Resource: '*'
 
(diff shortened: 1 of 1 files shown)


bencodeorg (approved):
LGTM

cat5inthecradle (approved):

PR #224, opened 2022-03-03
Add an API Gateway that supports PUT requests to the content bucket

jmkulwik:
This adds the new API Gateway HTTP API that supports PUT requests to the content bucket in order to pre-seed student projects and metadata prior to starting a Javabuilder execution.

https://codedotorg.atlassian.net/browse/JAVA-447 

As part of this, I added a new authorizer for the HTTP API that is similar to the existing authorizer for the WebSocket API. There were a small number of differences needed for the HTTP API, and the two will need to be split up as part of our upcoming throttling work. This seemed like a good point to split the two.

There were a couple concessions made in order to be able to update an existing Javabuilder stack. These are mostly in the form of odd names for a few of the services. HttpAPI, rather than HttpApi is one such example. In order to stay consistent with the existing code in the template, Api anywhere defined by us was maintained as API. This is because AWS is not case sensitive but is case aware. When updating the services, it detects that WebSocketAPI != WebSocketApi in some places but not others. When we deploy a fresh Javabuilder to remove the "beta" from the domain name, we should update all these casing differences.

Additionally, the WebSocket DomainName object is just called "DomainName" rather than being updated to "WebSocketDomainName" to match "HttpDomainName." This is because a domain name service cannot be updated via a CloudFormation template without updating the text of the domain name itself. It will always hit an error saying that "the domain name already exists."

For folks who'd like to test it out, you can use the following endpoints:
Fresh deploy of Javabuilder:
wss://javabuilder-jess.dev-code.org
https://javabuilder-jess-http.dev-code.org

Updated Javabuilder (starting from the current origin/main):
wss://javabuilder-jess5.dev-code.org
https://javabuilder-jess5-http.dev-code.org


javabuilder-authorizer/http_authorizer_function.rb
```diff
@@ -0,0 +1,109 @@
...
+end
+
+# Load the JWT public key from the appropriate environment variable
+# based on the environment the request was sent from.
+#
+# method_arn is in format
+# arn:aws:execute-api:region:account-id:api-id/stage-name/$connect
+# We only care about stage--that tells us the environment (development, staging, etc.)
+# def get_public_key(method_arn)
+def get_public_key(origin)
```
sanchitmalhotra126, line 78: Do you think it would make sense to move some of the shared code between authorizers (like the key selection logic) to a shared file? Or do you think these pieces are likely to diverge as well?

template.yml.erb
```diff
@@ -41,6 +41,146 @@ Globals:
 Conditions:
   IsDevCondition: !Equals [!Ref BaseDomainName, "dev-code.org"]
 Resources:
+  HttpDomain:
+    Type: AWS::Route53::RecordSet
+    Properties:
+      HostedZoneName: !Sub "${BaseDomainName}."
+      Name: !Sub "${SubDomainName}-http.${BaseDomainName}"
```
sanchitmalhotra126, line 48: Will the prod route look something like `https://javabuilderbeta-http.code.org`?

template.yml.erb
```diff
@@ -41,6 +41,146 @@ Globals:
...
+    Type: AWS::ApiGatewayV2::Api
+    Properties:
+      Name: !Sub "${SubDomainName}-http.${BaseDomainName}"
+      ProtocolType: HTTP
+
+  PutRoute:
+    Type: AWS::ApiGatewayV2::Route
+    Properties:
+      ApiId: !Ref HttpAPI
+      RouteKey: PUT /seedsources/sources.json
```
sanchitmalhotra126, line 91: will the final POST/PUT url be `https://javabuilderbeta-http.code.org/seedsources/sources.json`? We may not need the `sources.json` part since we already assign the file name directly in the lambda, right?

template.yml.erb
```diff
@@ -225,12 +368,21 @@ Resources:
       FunctionName: !Ref StartSessionAndRelayMessagesFunction
       Principal: apigateway.amazonaws.com
 
-  WebsocketAuthorizerLambda:
+# AWS does not seem to be case sensitive, but is case aware. Therefore, we can't update from
+# Websocket here to WebSocket as is used elsewhere when updating an existing stack. Updating this is
+# a step we should take when we create our non-beta environment from scratch. At that point, we
```
sanchitmalhotra126, line 373: Should we create a tech debt task for this?

sanchitmalhotra126 (commented):
Nice! Just a couple questions. Also, what is the difference between the two URLs you provided in the description? Is the "fresh deploy" not from main?

api-gateway-routes/api_gateway_put_function.rb
```diff
@@ -0,0 +1,11 @@
...
+
+def lambda_handler(event:, context:)
+  region = context.invoked_function_arn.split(':')[3]
+  client = Aws::S3::Client.new(region: region)
+  client.put_object(
+    body: event["body"],
+    bucket: ENV["CONTENT_BUCKET_NAME"],
+    key: event["requestContext"]["authorizer"]["lambda"]["sid"].to_s + "/sources.json"
+  )
+end
```
molly-moen, line 11: nit: newline?

javabuilder-authorizer/http_authorizer_function.rb
```diff
@@ -0,0 +1,109 @@
...
+  # Temporarily choose the key based on the client origin rather than the
+  # method_arn until we have environment-specific Javabuilders set up.
+  # tmp = method_arn.split(':')
+  # api_gateway_arn = tmp[5].split('/')
+  # stage_name = api_gateway_arn[1]
+
+  stage_name = ""
+  if origin.include? "localhost"
```
molly-moen, line 86: this might be a good time to update this to be the entire url instead of just a portion (at least for localhost, staging, levelbuilder, test and prod)

javabuilder-authorizer/http_authorizer_function.rb
```diff
@@ -0,0 +1,109 @@
...
+# Load the JWT public key from the appropriate environment variable
+# based on the environment the request was sent from.
+#
+# method_arn is in format
+# arn:aws:execute-api:region:account-id:api-id/stage-name/$connect
+# We only care about stage--that tells us the environment (development, staging, etc.)
+# def get_public_key(method_arn)
+def get_public_key(origin)
+  # Temporarily choose the key based on the client origin rather than the
+  # method_arn until we have environment-specific Javabuilders set up.
```
molly-moen, line 80: is this comment still applicable?

template.yml.erb
```diff
@@ -225,12 +368,21 @@ Resources:
...
+# should also update uses of HttpAPI and WebSocketAPI to be HttpApi and WebSocketApi, respectively,
+# to match the AWS standard elsewhere in this template. Example: "AWS::ApiGatewayV2::ApiMapping"
+<%{
+  Http: {LambdaName: "Http", Handler: "http_authorizer_function", Description: "'Authorize PUT by decoding JWT in Authorization querystring parameter.'"},
```
bencodeorg, line 377: Nit: can this be broken up onto multiple lines for readability?

template.yml.erb
```diff
@@ -41,6 +41,146 @@ Globals:
 Conditions:
   IsDevCondition: !Equals [!Ref BaseDomainName, "dev-code.org"]
 Resources:
+  HttpDomain:
```
bencodeorg, line 44: For future reference, what was your workflow for figuring out what resources you needed for this?

jmkulwik: Oof. In this particular case, I just copied and pasted everything from the WebSocket API that was domain name related and then didn't really touch it besides updating all the names to use "http." Writing that out makes me realize this may be a good opportunity to collect these into a grouping similar to the authorizer lambda so it's a little clearer that those two sections are repeated code.

template.yml.erb
jmkulwik: Yes

template.yml.erb
jmkulwik: Yeah, we can make the latter part of the URL whatever we want. I can also update this to be 
`PUT /seedsources/{filename}`
instead, so we have more flexibility on the filename. Thoughts?

template.yml.erb
jmkulwik: 👍

template.yml.erb
jmkulwik: Good call: https://codedotorg.atlassian.net/browse/JAVA-470

template.yml.erb
```diff
@@ -41,42 +41,153 @@ Globals:
 Conditions:
   IsDevCondition: !Equals [!Ref BaseDomainName, "dev-code.org"]
 Resources:
-  Domain:
+# Note: We can't update the name of a DomainName resource once it has been created because the
+# domain name itself has already been provisioned. When we change from javabuilderbeta to
+# javabuilder, we should update the WebSocket resources here to use the prefix "WebSocket"
```
jmkulwik, line 46: Task: https://codedotorg.atlassian.net/browse/JAVA-470

javabuilder-authorizer/http_authorizer_function.rb
jmkulwik: Good catch. Although that does bring up a deeper conversation. Do we think we'll ever choose the key based on the version of Javabuilder? I'd guess since Dashboard is the one authenticating an authorized user and we're just verifying it here, we wouldn't want Dashboard to choose its private key based on the version of Javabuilder it was invoking. Instead, it seems to make sense that Javabuilder would have multiple public keys registered and select the one it uses based on the client invoking it.

template.yml.erb
sanchitmalhotra126: I think we'll want to be restrictive on the file name because Javabuilder is currently configured to always look for a file named "sources.json". In which case either `PUT /seedsources` or `PUT /seedsources/sources.json` should be fine.

javabuilder-authorizer/http_authorizer_function.rb
jmkulwik: Good point. In both lambdas, we'll be wanting to decode and verify the token even though the http authorizer will end up doing a bunch of additional things too. I'll see what I can split out.

api-gateway-routes/api_gateway_put_function.rb
jmkulwik: 👍

jmkulwik:
> Nice! Just a couple questions. Also, what is the difference between the two URLs you provided in the description? Is the "fresh deploy" not from main?

Yep, the "fresh deploy" is deploying this template without a pre-existing stack. It's there to prove that the stack can be deployed from scratch. The "updated" version started with a stack deployed from main and then deployed with this template. It's there to prove that existing deploys will continue to work when they apply this template.

template.yml.erb
jmkulwik: Actually, were you asking about figuring out the resources I needed for the new domain name or for the entire set of work here?

javabuilder-authorizer/http_authorizer_function.rb
jmkulwik: I'll plan to do this as a follow-up task so as to unblock other workstreams. https://codedotorg.atlassian.net/browse/JAVA-471

javabuilder-authorizer/http_authorizer_function.rb
jmkulwik: I'll plan to do this as a follow-up task so as to unblock other workstreams. https://codedotorg.atlassian.net/browse/JAVA-471

template.yml.erb
jmkulwik: Regarding all the resources needed for the API Gateway api, there were a few steps

- First, Sanchit had created a working API Gateway api. I was able to use this as a reference to start off creating my version.
- Part way through, I realized the type of API he'd created (REST) wasn't supported in CloudFormation. So then I made a new version of his API that used a different type (HTTP). I started using this as a reference. I started off without the authorizer in order to make testing easier (no need to generate a JWT every time)
- When working on the API, I started with a fresh yml template and hacked the deploy.sh script to deploy mine instead of the main template. This allowed me to iterate much faster. (deploy time of ~30s rather than ~5min)
- I also used a .yml template rather than a .yml.erb template. This just generally made things easier since I didn't have to mix syntaxes
- To figure out the resources I needed, I did a combination of referencing the docs (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigatewayv2-api.html) which are about as easy to read as ancient Greek, referencing the sample I'd created, and referencing the WebSocket API resources that Suresh had created. As I got closer to the right solution, the docs did get more useful, but more so in telling me how to assemble things I already had figured out I needed, rather than telling me what I needed.
- To be completely honest, though, there was a lot of guess-and-check. Looking back at my pomodoro history in Marinara, I spent 17 hours over the past two weeks in deep focus just trying various things out in addition to time I spent bouncing ideas off the infra team and time I didn't track. The pomodoros were helpful because I could use them as a litmus test for whether I was making any progress - two pomodoros without any progress was a pretty clear indicator that I was stuck (or really needed a break)

sanchitmalhotra126 (approved):
LGTM!

template.yml.erb
bencodeorg: Yeah I was looking for a overall summary, just like ^^^, thanks! lol at "ancient Greek"

api-gateway-routes/api_gateway_put_function.rb
```diff
@@ -0,0 +1,11 @@
+require 'aws-sdk-s3'
+
+def lambda_handler(event:, context:)
+  region = context.invoked_function_arn.split(':')[3]
```
sureshc, line 4: Not critical, but there's an [ARN class](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/ARN.html) with handy methods for parsing the ARN and accessing the different logical identifiers within it.

cat5inthecradle (approved):

template.yml.erb
```diff
@@ -41,42 +41,153 @@ Globals:
...
+    Type: AWS::ApiGatewayV2::Api
+    Properties:
+      Name: !Sub "${SubDomainName}-http.${BaseDomainName}"
+      ProtocolType: HTTP
+
+  PutRoute:
+    Type: AWS::ApiGatewayV2::Route
+    Properties:
+      ApiId: !Ref HttpAPI
+      RouteKey: PUT /seedsources/sources.json
```
sureshc, line 99: Probably too late to change this, but I would think the Resource we're trying to PUT is the source code of a project, so the path of this API would be something like `/project/sources.json` and eventually we'll implement the PUT `/project/assets/whatever.jpg` or `/project/assets/assets.zip` routes.

template.yml.erb
jmkulwik: Oh one other item. Only the infra team can update the iam permissions template. I hit a couple snags with this, but ultimately ended up not needing to update it after all. Adding a completely new resource type (such as db tables) will likely need an update to the IAM permissions.

PR #225, opened 2022-03-05
Prevent null pointer exception in compiler error generation

molly-moen:
We are seeing occasional null pointer exceptions due to [diagnostic.getSource()](https://github.com/code-dot-org/javabuilder/blob/5a949622aabac7e608aeed57c87c883a05c57e9d/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserCodeCompiler.java#L154) being null. I have not been able to reproduce this on my own, so I added a check for a null source or -1 line number (which I learned can also be the case based on the [docs](https://docs.oracle.com/en/java/javase/11/docs/api/java.compiler/javax/tools/Diagnostic.html#getLineNumber()). If we hit one of these cases, log a warning and return the default compiler error. We can then hopefully get more insight into why this is happening based on the warning we log.


sanchitmalhotra126 (approved):

jmkulwik (approved):
LGTM

PR #226, opened 2022-03-08
Set up CodeBuild for pull requests

cat5inthecradle:
See "cicd/README.md" for an overview of the setup.

There are three categories of resources we need to deploy

1. Global resources, created once, may require admin permission to deploy, as this includes IAM Roles
2. CICD resources, created once per deployed branch (`main` plus adhocs)
3. App resources, created once per environment (dev, test, prod)

## Immediate Next Steps:

1. Merge this PR into `main`
4. Manually deploy "setup" stack to AWS Prod
5. Manually deploy the "cicd" stack to AWS Prod
6. Remove `JavabuilderCodeBuildRole` from "iam.yml", updating usage to use the role defined in "setup"

## Bigger Next Steps

1. Build a environment-parameterized version of "template.yml" in "cicd/3-app"
2. Create a CodePipeline project in "cicd/2-cicd/pipeline.template.yml" which builds, tests, and packages the app cloudformation template.
3. Deploy the app stack to a Test environment
4. Configure a CodePipeline step to run integration/acceptance tests in the Test environment
7. Plan the deployment of the app stack to the Prod environment (fully converting the root "template.yml")


cat5inthecradle:
The build status indicators in this PR are coming from an AWS Dev instance of this codebuild project. They are failing as expected because the template.yml has linting errors.

cicd/1-setup/setup.template.yml
```diff
@@ -0,0 +1,64 @@
...
+    Value: !Ref ArtifactStore
+    Export: {Name: JavabuilderCodeBuildArtifactBucket}
+  JavabuilderCodeBuildRoleArn:
+    Description: Javabuilder CodeBuild Role ARN
+    Value: !GetAtt JavabuilderCodeBuildRole.Arn
+    Export: {Name: JavabuilderCodeBuildRoleArn}
+  JavabuilderCodeBuildRoleName:
+    Description: Javabuilder CodeBuild Role Name
+    Value: !Ref JavabuilderCodeBuildRole
+    Export: {Name: JavabuilderCodeBuildRoleName}
```
cat5inthecradle, line 64: I haven't found a good way to get both of these without exporting both of them. Unfortunately in the pipeline stack I need the ARN in one case, and the Role Name in another.

cicd/2-cicd/pipeline.template.yml
```diff
@@ -0,0 +1,82 @@
...
+              - codebuild:CreateReport
+              - codebuild:UpdateReport
+              - codebuild:BatchPutTestCases
+              - codebuild:BatchPutCodeCoverage
+            Resource:
+              - !Sub arn:aws:codebuild:us-east-1:165336972514:report-group/${AWS::StackName}-${GitHubBranch}-pr-build
+          - Effect: Allow
+            Action:
+              - route53:ListHostedZonesByName
+            Resource: "*"
```
cat5inthecradle, line 80: I have a note to revisit this '*' permission. It's a very minor violation of least-privilege and could probably be tightened or removed.

cicd/3-app/buildspec.yml
```diff
@@ -0,0 +1,29 @@
...
+      java: corretto11
+      python: 3.8
+    commands:
+      - gem install bundler
+      - pip install cfn-lint
+  build:
+    # This should be moved to a shell script if it gets more complicated.
+    commands:
+      - set -e
+      - BRANCH_NAME=${CODEBUILD_WEBHOOK_HEAD_REF##*/}
```
cat5inthecradle, line 15: This keeps only the last item of a `/` separated string. So `refs/heads/my-branch` becomes `my-branch`. This will not work for branches with a `/` in the name, like `feature/mybranch` or `darin/mybranch`

The only reason I'm trimming it is so I can generate a path in the S3 bucket. Need a better solution than `${HEAD_REF##*/}`

cat5inthecradle: Might just be `${HEAD_REF#refs/heads/}`...

cicd/README.md
```diff
@@ -0,0 +1,55 @@
...
+  * Update the PR with the build/test status
+  * Provide easily accessible logs and artifacts from the build/test job
+* **CodePipeline** For every commit merged to `main`
+  * Run tests and build artifacts necessary for deployment
+  * Deploy the application to a Test environment
+    * As prod-like as possible
+  * Run integration and acceptance tests in the Test environment
+  * Make test results and artifacts easily accessible to developers
+  * If Test environment passes all tests, then deploy to Prod environment
+  * Deploy the application to a Dev environment
```
jmkulwik, line 24: This last line doesn't seem like it fits with the other. Is this a separate dev environment similar to the ones we deploy to test out AWS changes or is this something similar to the staging environment?

cicd/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+### 3 - App
+
+All of the application resources are defined in a CloudFormation template. We could manually update the CloudFormation stack with an updated template every time we have a code change, but it is considerably less effort and less risky to automate this with a CI/CD pipeline. By using parameters, we can use the same code to deploy multiple environments.
+
+**Note: The application template has not been incorporated into the CI/CD implementation yet. That is the next step. The plan is to migrate 'template.yml.erb' from the root, into this folder.**
+
+We also keep some CodeBuild configuration here, as this code tends to be more coupled to current application code, than the CI/CD code in the next section.
+
+* "app.template.yml" - AWS resources for the application
```
jmkulwik, line 38: "app.template.yml" - Is this just template.yml right now? Will this be renamed as part of this work?

jmkulwik (approved):
LGTM! I don't have too much knowledge in this area, but from what I know, this seems good. Thanks for the detailed readme.

cicd/README.md
cat5inthecradle: Since it's one of several cloudformation templates, I think a specific name can help, TBD whether 'app' is the most appropriate.

cicd/README.md
cat5inthecradle: The idea here would be an environment that developers can interact with and not worry about impacting production or the cicd pipeline. Resources might be running at a much lower scale, and access to these resources via the AWS console would be more permissive to support troubleshooting and issue replication.

IMO the Test environment should be as prod-like as possible and protected from out-of-pipeline changes. TBD whether a third environment running `main` is actually useful to the team.

I've usually had this environment get updated first, and a successful deployment to Dev be a prerequisite for deploying to Test - but I could also see this being a parallel deployment.

cicd/README.md
cat5inthecradle: But yes, this is "template.yml" today.

cicd/README.md
jmkulwik: Ahh ok. That makes sense.

cicd/README.md
jmkulwik: 👍

PR #227, opened 2022-03-09
Add configurable hourly and daily limit parameters to Javabuilder

bencodeorg:
Allows setting a limit per hour and limit per day parameter for Javabuilder, and passes that value to the authorizers as an environment variable (HTTP is the only one that actually needs it, but easier to pass to both WebSocket and HTTP). Values can be overridden by parameters passed in via the deploy script. Values are 10 per hour and 40 per day be default.

This PR does not actually implement any limits -- it just allows us to set the limit values. Part of preventing overuse of javabuilder work, details here:
- [Jira item](https://codedotorg.atlassian.net/browse/JAVA-458)

## Testing story

I tested deploying without setting any explicit values in the deploy script call and saw environment variables being set for the HTTP authorizer appropriately. I deployed again withe explicit values as arguments, and saw those values reflected in the HTTP authorizer's environment variables.

### DIFF
--- a/template.yml.erb
+++ b/template.yml.erb
@@ -355,6 +355,9 @@ Resources:
 # a step we should take when we create our non-beta environment from scratch. At that point, we
 # should also update uses of HttpAPI and WebSocketAPI to be HttpApi and WebSocketApi, respectively,
 # to match the AWS standard elsewhere in this template. Example: "AWS::ApiGatewayV2::ApiMapping"
+#
+# Note: hourly and daily limit values provided to both authorizers here as environment variables,
+# but are only needed in the HTTP authorizer.
 <%{
   Http: {
     LambdaName: "Http",
--- a/template.yml.erb
+++ b/template.yml.erb
@@ -412,6 +412,8 @@ Resources:
             ZUbJMN3hMFIFP4smAtjkKqguLJPFdcMij7WFDjL5kPQpxuc4RSq97iXRODIBjW1B\n zdN9ZpGRA4MDOu3BtLC69PwoFrSULAzlUmC5fJIwpaXoRGZEYRMQThlJ9VsRGwTG\n
             JTJJ4dWJ2cw61prbeU/HVCWD8OaNumz9tQIDAQAB\n
             -----END RSA PUBLIC KEY-----'
+          limit_per_hour: !Ref LimitPerHour
+          limit_per_day: !Ref LimitPerDay
 
   <%=name%>AuthorizerPermission:
     Type: AWS::Lambda::Permission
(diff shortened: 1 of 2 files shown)


template.yml.erb
```diff
@@ -401,6 +412,8 @@ Resources:
             ZUbJMN3hMFIFP4smAtjkKqguLJPFdcMij7WFDjL5kPQpxuc4RSq97iXRODIBjW1B\n zdN9ZpGRA4MDOu3BtLC69PwoFrSULAzlUmC5fJIwpaXoRGZEYRMQThlJ9VsRGwTG\n
             JTJJ4dWJ2cw61prbeU/HVCWD8OaNumz9tQIDAQAB\n
             -----END RSA PUBLIC KEY-----'
+          limit_per_hour: !Ref LimitPerHour
+          limit_per_day: !Ref LimitPerDay
```
jmkulwik, line 416: NIT: Intentional to add these to both lambdas? (Not sure that there's a straightforward way to omit this from just the WebSocket Lambda, so there may not be anything to do, but it may be worth adding a comment in case these get split out in the future)

jmkulwik:
Mind linking the task and dev doc in the PR description?

jmkulwik (approved):
LGTM! Once this is merged, mind updating the "Deploying Javabuilder" doc to include the environment variables for the production deploy so the limits are set to their higher values in production?

template.yml.erb
bencodeorg: Yeah there might? be a way to add just to http, but think it's probably more confusing than it'd be worth. I'll add a comment.

bencodeorg:
> LGTM! Once this is merged, mind updating the "Deploying Javabuilder" doc to include the environment variables for the production deploy so the limits are set to their higher values in production?

@jmkulwik just confirming, the daily limit we want to set is 300? See it mentioned in the dev doc but doesn't seem set in stone?

sanchitmalhotra126 (commented):
For my understanding - does this change actually enforce the limits, or is this just a way for us to pass these parameters to the lambdas?

bencodeorg:
> For my understanding - does this change actually enforce the limits, or is this just a way for us to pass these parameters to the lambdas?

@sanchitmalhotra126 it just allows us to pass the values, I've updated the PR description to clarify.

sanchitmalhotra126:
> > For my understanding - does this change actually enforce the limits, or is this just a way for us to pass these parameters to the lambdas?
> 
> @sanchitmalhotra126 it just allows us to pass the values, I've updated the PR description to clarify.

Thanks for clarifying!

PR #228, opened 2022-03-10
IAM updates to create DynamoDB tables

bencodeorg:
Updates to IAM template to allow creating DynamoDB tables for preventing overuse of Javabuilder work. Needs to be merged before merging [updates to Javabuilder template that create new DynamoDB tables](https://github.com/code-dot-org/javabuilder/pull/229).

Part of preventing overuse of Javabuilder work:
- [JIRA item](https://codedotorg.atlassian.net/browse/JAVA-458)

## Testing story

This hasn't really been tested, since my account in our dev AWS account is in the "Developer" group. which has superuser permissions. I've [reached out to infrastructure](https://codedotorg.slack.com/archives/C03CK49G9/p1646941303619749) to see how I should test this change.

### DIFF
--- a/iam.yml
+++ b/iam.yml
@@ -161,10 +161,7 @@ Resources:
                 Action:
                   - 'dynamodb:CreateTable'
                 Resource:
-                  - !Sub "arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/*_blocked_users"
-                  - !Sub "arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/*_tokens"
-                  - !Sub "arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/*_user_requests"
-                  - !Sub "arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/*_teacher_associated_requests"
+                  - !Sub "arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/*"
   JavabuilderCodeBuildRole:
     Type: AWS::IAM::Role
     Properties:
--- a/iam.yml
+++ b/iam.yml
@@ -157,6 +157,14 @@ Resources:
                   - !Sub "arn:aws:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/apigateway/*"
                   - !Sub "arn:aws:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/apigateway/"
                   - !Sub "arn:aws:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/apigateway/*:log-stream:*"
+              - Effect: Allow
+                Action:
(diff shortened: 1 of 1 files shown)


jmkulwik (commented):
This seems good to me, but I'm going to defer to Infra here.

iam.yml
```diff
@@ -157,6 +157,14 @@ Resources:
...
                   - !Sub "arn:aws:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/apigateway/"
                   - !Sub "arn:aws:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/apigateway/*:log-stream:*"
+              - Effect: Allow
+                Action:
+                  - 'dynamodb:CreateTable'
+                Resource:
+                  - !Sub "arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/*_blocked_users"
+                  - !Sub "arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/*_tokens"
+                  - !Sub "arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/*_user_requests"
+                  - !Sub "arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/*_teacher_associated_requests"
```
cat5inthecradle, line 167: I'm curious about this specific filter pattern. Would something like `javabuilder_*` or `javabuilder_*_blocked_users" be more appropriate (with the `*` in the second case likely representing an environment name).

bencodeorg: Yeah that would be preferable -- the reason I left it more broad is that the way I'm setting up table names is by concatenating the subdomain name and `_blocked_users`, etc. We don't enforce that subdomain names need to start with javabuilder (although by convention, they have -- eg, javabuilderbeta, javabuilder-ben).

https://github.com/code-dot-org/javabuilder/pull/229

I didn't think that dynamoDB table names was the right place to enforce that javabuilder* prefix for subdomain names, but LMK if you disagree.

cat5inthecradle: That makes sense. Honestly a `*` wouldn't be inappropriate, as I think you could make a similar argument that this isn't the place to define what tables the application needs. This doesn't stop you from creating more than four tables, since you could easily create `pokemon_name_but_lets_call_it_user_requests`.

The cloudformation role needs the ability to create dynamo db tables, and unless we know the exact name, we're not so much restricting as we are enforcing a naming convention.

bencodeorg: Yeah good point -- would you prefer that I just wildcard it entirely so that if we ever add dynamodb tables in the future, we don't need to loop in infra?

cat5inthecradle: I think that would be fine. I will be revisiting this stack as a whole in the near future, and some of my thoughts for restructuring what gets deployed with which stack might open opportunities to more appropriate limit the access.

cat5inthecradle (approved):

cat5inthecradle (approved):

PR #229, opened 2022-03-10
Create DynamoDB tables for overuse protection

bencodeorg:
Creates new tables needed for Javabuilder overuse protection. See dev doc for more detail about the schema for each table and how it will be used.

Part of preventing overuse of Javabuilder work:
- [JIRA item](https://codedotorg.atlassian.net/browse/JAVA-458)

Do not merge until [this PR to update IAM roles ](https://github.com/code-dot-org/javabuilder/pull/228)has been merged and deployed.

Some notes:
- The names of the tables are currently implemented as `${SubDomainName}_blocked_users` for example, which should work well for our future production javabuilder (ie, tables will be named like `javabuilder_blocked_users`). It is less clean for a) situations where we have dashes in subdomain names (eg, `javabuilder-ben`), where we'd end up with table names like `javabuilder-ben_blocked_users`. It also could result in not all Javabuilder DynamoDB tables having the javabuilder prefix if someone decides to use a subdomain name without the javabuilder prefix. We could rename the tables to be kebab case to be consistent, but not sure if that's worthwhile.
- I configured backups for our `blocked_users` table. Other tables have data retention policies of a day or less, so didn't feel worthwhile.
- I set per second limits of reads/writes of these tables to 1000 (somewhat arbitrarily). May want to configure this to be different between dev and production deploys.

## Testing story

I tested in our dev account that I was able to create each of these 4 tables, and that the configured settings were as expected.


template.yml.erb
```diff
@@ -693,6 +696,91 @@ Resources:
                 Period: 300
                 Stat: Sum
 <%end -%>
+# We use shortened versions of names for partition keys (eg, user_id),
+# but values will be a concatenation of the domain name and appropriate ID.
+# Values will look something like:
+# studio.code.org#123456 (user_requests table)
+# studio.code.org#UserId#123456 (blocked_users table)
```
sanchitmalhotra126, line 703: Just curious, why are we doing this concatenation instead of storing the domain separately? Is it so storage/lookup is quicker?

bencodeorg: Yeah good question, I was doing some reading on this as well. My understanding is DynamoDB is very fast when you are looking things up by the partition key. Our access pattern will always be to have a domain and user ID we want to look up, so we can get speedy results if we concatenate them and use the concatenated string to look up the record of interest.

Jessie had a couple of resources in her dev doc in case you want to learn more:
https://www.profit4cloud.nl/blog/designing-for-amazon-dynamodb/
https://aws.amazon.com/blogs/database/choosing-the-right-dynamodb-partition-key/

sanchitmalhotra126 (commented):
LGTM as far as I can tell! I'll defer to Jessie though

template.yml.erb
```diff
@@ -34,7 +34,6 @@ Parameters:
     Description: The number of Javabuilder invocations allowed per user per day.
     MinValue: 1
     Default: 40
-
```
bencodeorg, line 37: I added this whitespace accidentally in a previous PR, removing it here.

template.yml.erb
```diff
@@ -693,6 +696,91 @@ Resources:
...
+          KeyType: HASH
+      ProvisionedThroughput:
+        ReadCapacityUnits: <%=READ_CAPACITY_PER_SECOND%>
+        WriteCapacityUnits: <%=WRITE_CAPACITY_PER_SECOND%>
+      AttributeDefinitions:
+        - AttributeName: <%=DOMAIN_AND_USER_ID_COMPOSITE_ATTRIBUTE_NAME%>
+          AttributeType: S
+      PointInTimeRecoverySpecification:
+        PointInTimeRecoveryEnabled: true
+  TokenStatusTable:
```
jmkulwik, line 728: NIT: Add a whitespace between resources. We've generally been putting a whitespace between each AWS resource to make it a bit easier to tell where one resource starts and the other ends.

template.yml.erb
```diff
@@ -693,6 +696,91 @@ Resources:
...
+          AttributeType: S
+      TimeToLiveSpecification:
+        AttributeName: <%=TIME_TO_LIVE_ATTRIBUTE_NAME%>
+        Enabled: true
+  UserRequestsTable:
+    Type: AWS::DynamoDB::Table
+    Properties:
+      TableName: !Sub "${SubDomainName}_user_requests"
+      KeySchema:
+        - AttributeName: <%=DOMAIN_AND_TOKEN_ID_COMPOSITE_ATTRIBUTE_NAME%>
```
jmkulwik, line 749: Was this supposed to be `DOMAIN_AND_USER_ID_COMPOSITE_ATTRIBUTE_NAME`

bencodeorg: Yes it is 🤦 , thanks for looking closely. Updating now.

template.yml.erb
bencodeorg: I can do that. Tangentially, is it possible for us to break up this file? It's getting pretty hairy.

jmkulwik: It should be possible. The code-dot-org repo has a few different template files that all get stitched together, although I don't know exactly how that works.

template.yml.erb
bencodeorg: Done.

jmkulwik (approved):
LGTM!

PR #230, opened 2022-03-15
Create Scene class and action models

sanchitmalhotra126:
First part of the new Theater API. This implements the student-facing Scene class which essentially just creates action objects for each operation. These action objects are just small model classes that encapsulate the parameters needed to perform the action. This also allows us to do some small optimizations, such as converting file names to Image objects/sound sample arrays before storing the action data.

This also includes a small change to add a FILE_NOT_FOUND exception key for TheaterRuntimeException so we can throw custom runtime exceptions for FileNotFoundExceptions, because a requirement of the new Theater API is to not have students explicitly catch FileNotFoundExceptions.

JIRA: https://codedotorg.atlassian.net/browse/JAVA-487


bencodeorg:
I know a lot of this is boilerplate from the API spec, but is there any way to break this PR up?

org-code-javabuilder/theater/src/main/java/org/code/theater/Scene.java
```diff
@@ -0,0 +1,436 @@
...
+  private int textHeight;
+  private Color textColor;
+  private Color strokeColor;
+  private Color fillColor;
+  private double strokeWidth;
+
+  public Scene() {
+    this.actions = new ArrayList<>();
+
+    this.font = DEFAULT_FONT;
```
molly-moen, line 35: style nit: you user `Scene.WIDTH` and `Scene.HEIGHT` below, but just `DEFAULT_FONT` here. Can we pick one pattern for static variables?

molly-moen (approved):
LGTM! I think it's fine to leave this as 1 PR since it's mostly getters and setters

org-code-javabuilder/theater/src/test/java/org/code/theater/SceneTest.java
```diff
@@ -0,0 +1,196 @@
...
+
+  private Scene unitUnderTest;
+
+  @BeforeEach
+  public void setUp() {
+    unitUnderTest = new Scene();
+  }
+
+  @Test
+  public void testAllActions() {
```
jmkulwik, line 26: This unit test would fail if any action was updated. Break this up into individual units so it's clearer which piece is failing when a code change happens.

org-code-javabuilder/theater/src/main/java/org/code/theater/Scene.java
```diff
@@ -0,0 +1,436 @@
...
+  }
+
+  /**
+   * Plays a note with the selected instrument.
+   *
+   * @param instrument the instrument to play.
+   * @param note the note to play. 60 represents middle C on a piano.
+   * @param seconds length of the note.
+   */
+  public final void playNote(Instrument instrument, int note, double seconds) {
```
jmkulwik, line 101: NIT: Should we also do the `playNote` and `playNoteAndPause` without an instrument? Thinking about the discussion here: https://codedotorg.slack.com/archives/C01EF4GJ9GE/p1647005749541949?thread_ts=1646959030.811979&cid=C01EF4GJ9GE

org-code-javabuilder/theater/src/main/java/org/code/theater/Scene.java
```diff
@@ -0,0 +1,436 @@
...
+   *
+   * @param color the color value to fill any shape.
+   */
+  public final void setFillColor(Color color) {
+    this.fillColor = color;
+  }
+
+  /**
+   * Sets the color of lines drawn. Note that this only applies to drawing operations performed by
+   * this actor.
```
jmkulwik, line 387: NIT: `actor` is still shown here. Perhaps "displayed by this scene" would be better? (same for the next several methods)

org-code-javabuilder/theater/src/main/java/org/code/theater/Scene.java
```diff
@@ -0,0 +1,436 @@
...
+
+  /**
+   * Removes the fill color so any shapes have no fill. Note that this only applies to drawing
+   * operations performed by this actor.
+   */
+  public final void removeFillColor() {
+    this.fillColor = null;
+  }
+
+  final List<SceneAction> getActions() {
```
jmkulwik, line 433: TIL: the default access modifier acts differently from public, private, and protected: https://www.baeldung.com/java-access-modifiers#:~:text=2.-,Default,package%20com.

jmkulwik: Mind blown. I was definitely about to ask for an explicit keyword here, but this is, in fact, the exact modifier needed. :)

org-code-javabuilder/theater/src/main/java/org/code/theater/Scene.java
```diff
@@ -0,0 +1,436 @@
...
+public class Scene {
+  private static final int WIDTH = 400;
+  private static final int HEIGHT = 400;
+
+  // Visible for testing
+  static final Font DEFAULT_FONT = Font.MONO;
+  static final FontStyle DEFAULT_FONT_STYLE = FontStyle.NORMAL;
+  static final int DEFAULT_TEXT_HEIGHT = 20;
+  static final Color DEFAULT_COLOR = Color.BLACK;
+  static final double DEFAULT_STROKE_WIDTH = 1.0;
```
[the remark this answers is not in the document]
jmkulwik: Just got a note from Mike (afk): The default font should be sans instead of mono, but otherwise, these are good.

org-code-javabuilder/theater/src/main/java/org/code/theater/Scene.java
sanchitmalhotra126: Oh good call, I'll add these

org-code-javabuilder/theater/src/main/java/org/code/theater/Scene.java
sanchitmalhotra126: Oh interesting! I actually didn't know this either and just thought it defaulted to `protected`. Good to know!

PR #231, opened 2022-03-16
Move internal Theater classes to support package

sanchitmalhotra126:
Quick prefactor for the next set of Theater v2 work. This moves all internal Theater classes (everything except Scene, Theater, Prompter, Instrument, and Stage) into the `support` package. The only code changes here are making a few constructors/methods public so that non-support classes can access them.

Tested that Theater still runs correctly.


molly-moen (approved):
Nice!

PR #232, opened 2022-03-17
Create classes for playing scenes

sanchitmalhotra126:
Second part of the Theater API v2 update. This adds the classes necessary for playing scenes by collecting actions and generating output. This is mostly just copying over the existing code from `Stage`, but broken into a couple different classes to help modularize things a bit.

- `ConcertCreator` is the main class that reads the list of `SceneAction`s, generates the output, writes the files, and sends messages via the output adapter. For audio commands, it invokes the AudioWriter/InstrumentLoader APIs directly and for drawing commands it delegates to `GraphicsHelper`. Also, because this just an instantiable class (as opposed to a singleton like Stage), we can have it implement `AutoCloseable` and use it in a try-with-resources block to make sure we always release resources if any exception is thrown.
- `GraphicsHelper` is where all the drawing logic happens, again just copied over from stage. This class deals with drawing shapes, lines, text, images, etc.
- `TheaterPlayer` is a singleton class that creates a new `ConcertCreator` and publishes the output. This is kind of a middle layer between the static `Theater` and the non-static, instantiable `ConcertCreator` - the main reason this class exists is because we need a way to track the global state of if `play` has been called or not, and then reset that state when execution ends. It didn't feel as clean to just store this state on a static variable in `Theater` that we keep setting and unsetting, so I opted to create this small class instead, which lets us hook into the LifecycleNotifier, and allows both ConcertCreator and Theater to not have to be concerned with if play has been called already or not.
- - This class also has a temporary `onStagePlay()` method that we call from Stage to prevent both Stage.play() and Theater.play(scenes) from being called in a project. Once stage has been deleted, we can get rid of this.
- `Theater` now has a `play(Scene ...scenes)` method so it can be called with one or multiple `Scenes`. However, if we do want explicit `playScene(Scene scene)` and `playScenes(Scene[] scenes)` methods, that's an easy switch.

JIRA: https://codedotorg.atlassian.net/browse/JAVA-487

Testing: tested locally with both Theater v1 and v2 commands. Confirmed that all drawing and sound APIs work as intended, and that using both v1 and v2 in conjunction results in an error.

### DIFF
--- a/org-code-javabuilder/protocol/src/main/java/org/code/protocol/ClientMessageDetailKeys.java
+++ b/org-code-javabuilder/protocol/src/main/java/org/code/protocol/ClientMessageDetailKeys.java
@@ -24,6 +24,7 @@ public class ClientMessageDetailKeys {
   public static final String UPLOAD_URL = "uploadUrl";
   public static final String PROGRESS_TIME = "progressTime";
   public static final String TOTAL_TIME = "totalTime";
+  public static final String URL = "url";
 
   // Exception specific
   public static final String CONNECTION_ID = "connectionId";
--- a/org-code-javabuilder/theater/src/main/java/org/code/theater/Scene.java
+++ b/org-code-javabuilder/theater/src/main/java/org/code/theater/Scene.java
@@ -1,5 +1,7 @@
 package org.code.theater;
 
+import static org.code.theater.support.Constants.THEATER_HEIGHT;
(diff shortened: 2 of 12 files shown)


molly-moen (approved):
Nice! The structure of these files makes things much easier to parse than Stage.

sanchitmalhotra126:
CodeBuild is failing due to linting (which is now fixed). Merging this change.

PR #234, opened 2022-03-17
Improve cold boot time of the lambdas

jmkulwik:
This is an update to move some expensive one-time steps to the initialization phase of the lambda. This allows us to take advantage of our [provisioned concurrency](https://docs.aws.amazon.com/lambda/latest/dg/provisioned-concurrency.html) in order to decrease our cold boot times.

On the Neighborhood and Console lambdas, cold boot times decreased from ~18s to ~9s
On the Theater lambda, cold boot times decreased from ~5s to ~3s

We may not be able to optimize our cold boot times much further than this because it appears that AWS Lambda only allows for a [10s initialization phase](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-context.html). This is not widely publicized and may change. I haven't verified this myself, so it's worth checking if we can use more initialization time. If we can, we definitely should as we have another ~8s of time we can improve on the smaller lambdas.

Note: The theater lambda has more than 3x the RAM of the other two lambdas, and therefore is expected to execute faster.

Note: This only improves the "cold boot" time for the lambda. This is the time it takes to set up a new lambda container the first time it is used. Both before and after this change, the non-cold boot run time for the Theater lambda running a "Hello World" program is ~0.5s and for the other two is ~0.75s. 

As part of this change, I also went ahead and removed the disk space logs, which were very verbose. I kept the "disk space report" log which we use to track disk space issues in Cloud Watch.

As a follow-up change, I'd like to add some sort of time report, similar to the disk space report, which records the amount of time we spend doing various pieces of work so we can track performance metrics separately from the time the user's project runs. [JAVA-499](https://codedotorg.atlassian.net/browse/JAVA-499)

As an additional follow-up, we should look into using the Java "Ahead of Time" compiler to [optimize specifically for AWS Lambda](https://www.baeldung.com/ahead-of-time-compilation). [JAVA-497](https://codedotorg.atlassian.net/browse/JAVA-497)

We may also be able further improve performance by assigning provisioned concurrency to the auxiliary lambdas (such as the authorizer lambdas) and move some steps to their init phase as well. The cold boot time for all of these combined appears to be around 5s (from watching my wrist watch) whereas the later runs are closer to 1s. We may see less improvement from adding provisioned concurrency here since these lambdas are used by all project types and therefore are used more often and cold boot less often. [JAVA-500](https://codedotorg.atlassian.net/browse/JAVA-500)

Finally, here's a potentially helpful [slack conversation](https://codedotorg.slack.com/archives/C03CK49G9/p1647459220180329) that contributed to this change.

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
+++ b/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
@@ -17,7 +17,6 @@ import com.amazonaws.services.sqs.AmazonSQSClientBuilder;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Paths;
-import java.time.LocalDateTime;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
@@ -44,7 +43,7 @@ public class LambdaRequestHandler implements RequestHandler<Map<String, String>,
   private static final String CONTENT_BUCKET_NAME = System.getenv("CONTENT_BUCKET_NAME");
   private static final String CONTENT_BUCKET_URL = System.getenv("CONTENT_BUCKET_URL");
   private static final String API_ENDPOINT = System.getenv("API_ENDPOINT");
-  private final AmazonApiGatewayManagementApi API;
+  private final AmazonApiGatewayManagementApi API_CLIENT;
   private final AmazonSQS SQS_CLIENT;
   private final AmazonS3 S3_CLIENT;
 
@@ -53,7 +52,10 @@ public class LambdaRequestHandler implements RequestHandler<Map<String, String>,
     // This will only be called once in the initial creation of the lambda instance.
     // Documentation: https://docs.aws.amazon.com/lambda/latest/dg/java-handler.html
     CachedResources.create();
-    API =
+
+    // Creating these clients here rather than in the request handler method allows us to use
+    // provisioned concurrency to decrease cold boot time by 3-10 seconds, depending on the lambda
+    API_CLIENT =
         AmazonApiGatewayManagementApiClientBuilder.standard()
             .withEndpointConfiguration(
                 new AwsClientBuilder.EndpointConfiguration(API_ENDPOINT, "us-east-1"))
@@ -74,7 +76,6 @@ public class LambdaRequestHandler implements RequestHandler<Map<String, String>,
    */
   @Override
   public String handleRequest(Map<String, String> lambdaInput, Context context) {
-    context.getLogger().log("1: " + LocalDateTime.now().toString());
     // The lambda handler should have minimal application logic:
     // https://docs.aws.amazon.com/lambda/latest/dg/best-practices.html
     final String connectionId = lambdaInput.get("connectionId");
@@ -93,10 +94,6 @@ public class LambdaRequestHandler implements RequestHandler<Map<String, String>,
         JSONUtils.booleanFromJSONObjectMember(options, "useNeighborhood");
     final List<String> compileList = JSONUtils.listFromJSONObjectMember(options, "compileList");
 
-    // The biggest chunk of time happens between points 2 and 3. Followed by points 3-4. Can we
-    // reuse parts of this between invocations.
-    // Also, wtf is going on with Neighborhood not updating??
-    context.getLogger().log("2: " + LocalDateTime.now().toString());
     Logger logger = Logger.getLogger(MAIN_LOGGER);
     logger.addHandler(
         new LambdaLogHandler(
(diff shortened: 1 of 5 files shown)


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
```diff
@@ -42,30 +40,45 @@
...
+  private final AmazonS3 S3_CLIENT;
 
   public LambdaRequestHandler() {
     // create CachedResources once for the entire container.
     // This will only be called once in the initial creation of the lambda instance.
     // Documentation: https://docs.aws.amazon.com/lambda/latest/dg/java-handler.html
     CachedResources.create();
+
+    // Creating these clients here rather than in the request handler method allows us to use
+    // provisioned concurrency to decrease cold boot time by 3-10 seconds, depending on the lambda
```
molly-moen, line 57: clarifying question: my gut feeling on "cold boot" is it would be running the init stage, so we would want less logic in the constructor. Clearly that gut feeling is wrong, can you explain why putting more in the constructor makes cold boot faster? Is there some prior stage to cold boot that I'm missing?

jmkulwik: There's an `Init` phase which "downloads the code for the function and all layers, initializes any extensions, initializes the runtime, and then runs the function’s initialization code (the code outside the main handler)." For lambdas that have provisioned concurrency set, this all happens before the first function invocation, so long as it takes less than 10s. 

Details there came from this doc: [Lambda execution lifecycle](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-context.html). 

There was also a note on the [java-specific docs](https://docs.aws.amazon.com/lambda/latest/dg/java-handler.html) that seems to indicate that the class constructor also runs during the init phase. "You can add initialization code outside of your handler method to reuse resources across multiple invocations. When the runtime loads your handler, it runs static code and the class constructor. Resources that are created during initialization stay in memory between invocations, and can be reused by the handler thousands of times."

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
```diff
@@ -42,30 +40,45 @@
   // This is an error code we made up to signal that available disk space is less than 50%.
   // It may be used in tracking on Cloud Watch.
   private static final int LOW_DISK_SPACE_ERROR_CODE = 50;
+  private static final String CONTENT_BUCKET_NAME = System.getenv("CONTENT_BUCKET_NAME");
+  private static final String CONTENT_BUCKET_URL = System.getenv("CONTENT_BUCKET_URL");
+  private static final String API_ENDPOINT = System.getenv("API_ENDPOINT");
+  private final AmazonApiGatewayManagementApi API_CLIENT;
+  private final AmazonSQS SQS_CLIENT;
+  private final AmazonS3 S3_CLIENT;
```
sanchitmalhotra126, line 48: nit: lowercase names? (`apiClient`,  etc)

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
```diff
@@ -155,9 +157,7 @@ public String handleRequest(Map<String, String> lambdaInput, Context context) {
       // Log disk space before clearing the directory
       LoggerUtils.sendDiskSpaceReport();
 
-      LoggerUtils.sendDiskSpaceUpdate(SessionTime.START_SESSION, ClearStatus.BEFORE_CLEAR);
       fileManager.cleanUpTempDirectory(null);
-      LoggerUtils.sendDiskSpaceUpdate(SessionTime.START_SESSION, ClearStatus.AFTER_CLEAR);
```
sanchitmalhotra126, line 160: Why were these logs removed?

sanchitmalhotra126 (commented):
Interesting!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
jmkulwik: Oh good call. Actually, I'm realizing these should probably be static since they are configured based on environment variables and won't change unless they're on a different environment.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
molly-moen: ah makes sense. My confusion was I thought our issue was being over that 10 second limit already, when really we weren't utilizing that limit to its fullest.

jmkulwik: Oh yeah, good callout. I forgot to clarify that once I discovered that was the case.

molly-moen (approved):

PR #239, opened 2022-03-21
Remove loading code sources from dashboard

sanchitmalhotra126:
Next cleanup PR - this removes code related to loading code sources from dashboard. Note - the `dashboardHostname` param is now no longer used but I did keep it around in the handler code because I think we may need it to be able to determine if a request is coming from localhost for when we stub asset files. If that doesn't end up being the case, I figured we can just remove it then.

JIRA: https://codedotorg.atlassian.net/browse/JAVA-451

Testing: tested locally, ensured that projects still load


molly-moen (approved):

jmkulwik (approved):
Nice cleanup!

PR #240, opened 2022-03-22
Updated the authorizer to match exact origins

jmkulwik:
With this, we'll use the exact client origins to choose the JWT decoder. If the request comes from a different origin, we won't decode the token. When dashboard invokes the HTTP authorizer, it passes the hostname (i.e. studio.code.org) whereas when Java Lab invokes the WebSocket authorizer, it passes the full origin (i.e. https://studio.code.org). Therefore a little bit of standardizing was needed to clean these up.

I tested these in postman against each environment the strings logged on Cloud Watch from the `puts` statement that is removed as part of this. I wasn't able to test directly against an adhoc, but did use an adhoc's hostname to verify the logic would work correctly.

Task: [JAVA-471](https://codedotorg.atlassian.net/browse/JAVA-471)

### DIFF
--- a/javabuilder-authorizer/jwt_helper.rb
+++ b/javabuilder-authorizer/jwt_helper.rb
@@ -66,25 +66,20 @@ module JwtHelper
     # tmp = route_arn.split(':')
     # api_gateway_arn = tmp[5].split('/')
     # stage_name = api_gateway_arn[1]
-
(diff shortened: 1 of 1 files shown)


sanchitmalhotra126 (approved):
Nice fix!

molly-moen (approved):
LGTM!

PR #241, opened 2022-03-22
Remove unused methods in JavabuilderFileManager interface

sanchitmalhotra126:
More cleanup. This removes the now unused methods from the `JavabuilderFileManager` interface, leaving it with only `cleanUpTempDirectory`. I also went ahead and renamed the interface and its implementations to better reflect its updated use.

Testing: tested locally


molly-moen (approved):
🎉

PR #242, opened 2022-03-22
Update Theater play methods

sanchitmalhotra126:
Updates to the `Theater.playScenes/play` method signatures (see https://codedotorg.slack.com/archives/C01EF4GJ9GE/p1647897243541699 for details)

Tested all methods locally.


jmkulwik (approved):
LGTM! Thanks for the follow-up with curriculum here!

PR #244, opened 2022-03-22
Fetch project data on initialization

sanchitmalhotra126:
Final cleanup/refactor 🎉 (at least for now) - this moves project data initialization to the ContentManager constructors now that we're no longer supporting dashboard sources.

Tested with local dashboard against local Javabuilder and deployed Javabuilder.

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalMain.java
+++ b/org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalMain.java
@@ -21,14 +21,22 @@ public class LocalMain {
     // turn off the default console logger
     logger.setUseParentHandlers(false);
 
+    final LocalContentManager localContentManager;
+    try {
+      localContentManager = new LocalContentManager();
+    } catch (InternalServerError e) {
+      System.out.println("Error loading data");
+      return;
+    }
+
     GlobalProtocol.create(
-        outputAdapter, inputAdapter, new LifecycleNotifier(), new LocalContentManager());
+        outputAdapter, inputAdapter, new LifecycleNotifier(), localContentManager);
     CachedResources.create();
 
     // Create and invoke the code execution environment
     CodeExecutionManager codeExecutionManager =
         new CodeExecutionManager(
-            new LocalContentManager(),
+            localContentManager,
             GlobalProtocol.getInstance().getInputHandler(),
             outputAdapter,
             ExecutionType.RUN,
--- a/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSContentManager.java
+++ b/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSContentManager.java
@@ -17,7 +17,7 @@ import java.util.Date;
 import org.code.protocol.ContentManager;
 import org.code.protocol.InternalErrorKey;
 import org.code.protocol.JavabuilderException;
-import org.code.protocol.LoggerUtils;
+import org.json.JSONException;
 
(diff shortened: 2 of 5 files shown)


org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalMain.java
```diff
@@ -21,14 +21,22 @@ public static void main(String[] args) throws InterruptedException {
     // turn off the default console logger
     logger.setUseParentHandlers(false);
 
+    final LocalContentManager localContentManager;
```
bencodeorg, line 24: Side note -- do you use `LocalMain`? I don't know that I have, but the description of how to use it mentions `LocalProjectFileManager`, which no longer exists?

Since this is maybe the last cleanup PR for a while, is there a quick update you could add of how to use `LocalMain` to those docs if we are still using it?

https://github.com/code-dot-org/javabuilder/blob/main/org-code-javabuilder/README.md

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSContentManager.java
```diff
@@ -43,8 +43,16 @@ public AWSContentManager(
...
-      Context context) {
-    this(s3Client, bucketName, javabuilderSessionId, contentBucketUrl, context, null);
+      Context context)
+      throws InternalServerError {
+    this.bucketName = bucketName;
+    this.s3Client = s3Client;
+    this.javabuilderSessionId = javabuilderSessionId;
+    this.contentBucketUrl = contentBucketUrl;
+    this.context = context;
+    this.projectData = this.loadProjectData();
```
bencodeorg, line 53: Is the other constructor just for tests?

bencodeorg (approved):
Couple of comments but LGTM!

org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalMain.java
molly-moen: `LocalMain` was originally added so we could run Javabuilder without any connection to code-dot-org, which was useful in the very beginning of Javabuilder's life but I don't think has had usage in 6+ months. I think we might want to revisit this tech debt item :) https://codedotorg.atlassian.net/browse/JAVA-227 @jmkulwik do you have any thoughts here?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSContentManager.java
```diff
@@ -66,23 +74,11 @@ public AWSContentManager(
 
   @Override
   public UserProjectFiles loadFiles() throws InternalServerError, UserInitiatedException {
```
molly-moen, line 76: should we rename this to `getFiles`?

org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalMain.java
sanchitmalhotra126: Yeah, the above was my understanding as well. If we do end up keeping it around though, I have a documentation task I'm planning on picking up and I can add some detail there.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSContentManager.java
sanchitmalhotra126: Yep! I couldn't call the other constructor with `this(...)` since I needed to initialize project data with `this.loadProjectData()`

PR #245, opened 2022-03-23
Only allow one instance of GlobalProtocol to be created per session

sanchitmalhotra126:
This change prevents multiple GlobalProtocol instances from being created in a single session. This wasn't really affecting us in production before, but in theory, the GlobalProtocol instance could be created multiple times which can lead to weird side effects since it's a very commonly used singleton across the service. Thanks to our class loader checks, this technically wouldn't be possible to do in student code, but there was still the possibility we could accidentally reset the instance internally, so this allows us to guard against that.

To accomplish this, GlobalProtocol.create() now happens in the onPreExecute() method in CodeExecutionManager, and the instance is destroyed (via GlobalProtocol.destroy()) in the onPostExecute() method.

I also did a small refactor to have the ProjectData class implement the ProjectFileLoader interface rather than the various ContentManagers. This is just because, without this refactor, we'd have to pass ContentManager into CodeExecutionManager twice (once as a ContentManager and once as a ProjectFileLoader) which just felt a little wonky - plus, the ContentManagers just call ProjectData directly to load source files so it wasn't a huge change to have ProjectData implement the interface instead. I think there is still some more cleanup we can do here and possibly eliminate the ProjectFileLoader interface entirely but there are some caveats with that which make it less straightforward than anticipated.

JIRA: https://codedotorg.atlassian.net/browse/JAVA-220

Testing: local + deployed instance

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalContentManager.java
+++ b/org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalContentManager.java
@@ -14,7 +14,7 @@ import org.code.protocol.InternalErrorKey;
 import org.code.protocol.JavabuilderException;
 import org.json.JSONException;
 
-public class LocalContentManager implements ContentManager, ProjectFileLoader {
+public class LocalContentManager implements ContentManager {
   private static final String SERVER_URL_FORMAT = "http://localhost:8080/%s/%s";
 
   private final ProjectData projectData;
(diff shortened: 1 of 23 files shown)


org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalContentManager.java
```diff
@@ -28,9 +28,8 @@ public LocalContentManager() throws InternalServerError {
     this.projectData = projectData;
   }
 
-  @Override
-  public UserProjectFiles loadFiles() throws InternalServerError, UserInitiatedException {
```
molly-moen, line 32: I think this change makes sense, but we'll need to pay attention to merge order because [this PR](https://github.com/code-dot-org/javabuilder/pull/246) conflicts with it.

molly-moen (approved):
LGTM!

org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalContentManager.java
sanchitmalhotra126: Sounds good, I'll go ahead and merge this now then

PR #246, opened 2022-03-23
Run validation code on a test run

molly-moen:
Update to run validation code on test runs of Javabuilder. Previously all code in the validation section of the json blob was ignored, but we would set up every test run as if we were running validation code so any validation code was run in start mode. Changes in this PR:
- create a new `BaseTestRunner` class which the existing `TestRunner` and the new `ValidationRunner` extend from. This contains almost all of the logic previously in `TestRunner`.
- add `ValidationRunner` which does the necessary setup to run validation code and runs it.
- Update the project loader to get validation code.

Now if a project has validation and user test code, the user will see `Running level tests...` followed by validation results, then `Running your project's tests...` followed by project test results. 

Note: [this PR](https://github.com/code-dot-org/code-dot-org/pull/45443) must be deployed before this is merged.

## Testing
Tested against local and deployed Javabuilder

## Next Steps
I will follow up with [this task](https://codedotorg.atlassian.net/browse/JAVA-468) to send validation code in start mode the same way we send it for other modes. With this change, validation code will generally run in start mode unless it uses some of our special validation apis.

## Links
- [JIRA](https://codedotorg.atlassian.net/browse/JAVA-404)
- [Spec](https://docs.google.com/document/d/14_ka0o1YFhAoXFl-WiyukrfIvT4C0IYIVVYcFaEmV7w/edit#heading=h.w2a9fn62fmlc)


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/BaseTestRunner.java
```diff
@@ -0,0 +1,77 @@
+package org.code.javabuilder;
```
molly-moen, line 1: note: this file is basically a copy/paste from the previous state of `TestRunner`, with an additional parameter for `StatusMessageKey`

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/TestRunner.java
```diff
@@ -1,86 +1,20 @@
...
-import org.junit.platform.launcher.TestPlan;
-import org.junit.platform.launcher.core.LauncherDiscoveryRequestBuilder;
-import org.junit.platform.launcher.core.LauncherFactory;
 
 /** Runs all tests for a given set of Java files */
-public class TestRunner implements CodeRunner {
-  private final List<JavaProjectFile> javaFiles;
-  private final JavabuilderTestExecutionListener listener;
-  private final OutputAdapter outputAdapter;
+public class TestRunner extends BaseTestRunner {
```
jmkulwik, line 8: I'm a touch concerned that "TestRunner" will sound too much like an internal util. Maybe something like "UserTestRunner" would help differentiate from our tests?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeBuilder.java
```diff
@@ -53,6 +60,12 @@ public void buildUserCode(List<String> compileList)
     this.compileCode(javaProjectFiles);
   }
 
+  public void buildUserAndValidationFiles() throws InternalServerError, UserInitiatedException {
```
sanchitmalhotra126, line 63: nit: mind adding a javadoc to differentiate this from `buildAllUserCode` and `buildUserCode`?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaRunner.java
```diff
@@ -2,37 +2,56 @@
...
 
 /** The class that executes the student's code */
 public class JavaRunner {
   private final URL executableLocation;
   private final MainRunner mainRunner;
   private final TestRunner testRunner;
+  private final ValidationRunner validationRunner;
   private final List<String> javaClassNames;
+  private final List<String> validationAndJavaClassNames;
+  private OutputAdapter outputAdapter;
```
sanchitmalhotra126, line 22: nit: final?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/TestRunner.java
molly-moen: sure I can update it

sanchitmalhotra126 (approved):
LGTM! For my understanding, with this change, we will always run validation code if we have it when running tests, right? We wouldn't be in a situation where we have validation code but only want to run student tests?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/ProjectData.java
```diff
@@ -83,4 +82,14 @@ public void addNewAssetUrl(String filename, String url) {
...
   }
+
+  public UserProjectFiles getValidation() throws UserInitiatedException, InternalServerError {
+    if (!this.jsonData.has(VALIDATION_KEY)) {
+      // return empty file list if there is no validation, as no validation is expected behavior
+      return new UserProjectFiles();
+    }
+    final UserProjectFiles validationFiles =
+        this.projectFileParser.parseFileJson(this.jsonData.getString(VALIDATION_KEY));
+    return validationFiles;
```
jmkulwik, line 93: NIT: would it make sense to just return `this.projectFileParser.parseFileJson(this.jsonData.getString(VALIDATION_KEY));`?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/MainRunner.java
```diff
@@ -24,7 +24,7 @@ public MainRunner(List<JavaProjectFile> javaFiles, OutputAdapter outputAdapter)
    * @param urlClassLoader class loader to load compiled classes
    * @throws JavabuilderException when the user's code hits an error
    */
-  public void run(URLClassLoader urlClassLoader) throws JavabuilderException {
+  public boolean run(URLClassLoader urlClassLoader) throws JavabuilderException {
```
jmkulwik, line 27: NIT: Add a `@returns` to the javadoc

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeRunner.java
```diff
@@ -10,6 +10,7 @@
    *
    * @param urlClassLoader class loader to load compiled code
    * @throws JavabuilderException if there is an error running code
+   * @return true if there was code to run, false if there was not
    */
-  void run(URLClassLoader urlClassLoader) throws JavabuilderException;
+  boolean run(URLClassLoader urlClassLoader) throws JavabuilderException;
```
jmkulwik, line 15: The return value here feels a little like code smell from a side effect. Given that the mainRunner ignores the return value, would it make more sense to have something like testRunner.hasTests() followed by testRunner.run()?

molly-moen: I think that would overly complicate things. We don't know whether we have tests until we have done all the setup work to run them, so we would either do that work twice or enforce that the caller has to call the methods in a specific order, which I don't like. I think a run method with a boolean return value of did or did not run code is logical. For main we end up throwing exceptions in every case we fail.

molly-moen:
> LGTM! For my understanding, with this change, we will always run validation code if we have it when running tests, right? We wouldn't be in a situation where we have validation code but only want to run student tests?

@sanchitmalhotra126 yes we will always run validation on a test run if it is present.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeRunner.java
jmkulwik: 👍

jmkulwik (approved):
LGTM!

PR #247, opened 2022-03-23
Add s3-bucket parameter to deploy command

molly-moen:
We have started getting occasional errors on deploy that our template file is too large. This adds an s3-bucket to the deploy command (the same one we use for the `package` command) which allows us to use a larger template file.

I tested this change by deploying to my dev instance of Javabuilder.


cat5inthecradle (approved):

sanchitmalhotra126 (approved):
Thanks for getting this!

PR #248, opened 2022-03-24
Stub asset file requests for localhost Dashboard and integration tests

sanchitmalhotra126:
This adds functionality to stub asset file requests on deployed/prod Javabuilder when running against localhost Dashboard or running integration tests, when those are set up. This is needed because asset URLs sent from Javalab point to Dashboard and if Dashboard is running on localhost on a developer's machine, deployed/prod Javabuilder cannot access these URLs. In the case of integration tests, we also want a way to run tests against a deployed instance of Javabuilder without needing a dashboard server running. I went ahead and used sampleImageBeach.jpg and beatbox.wav as stub files since these were already checked into the repo.

JIRA: https://codedotorg.atlassian.net/browse/JAVA-454

Tested with localhost dashboard + deployed Javabuilder.

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketServer.java
+++ b/org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketServer.java
@@ -61,8 +61,6 @@ public class WebSocketServer {
     final String miniAppType = queryInput.getString("mini_app_type");
     final ExecutionType executionType =
         ExecutionType.valueOf(queryInput.getString("execution_type"));
-    // TODO: dashboardHostname is currently unused but may be needed for stubbing asset files
-    final String dashboardHostname = "http://" + queryInput.get("iss") + ":3000";
     final JSONObject options = new JSONObject(queryInput.getString("options"));
     final List<String> compileList = JSONUtils.listFromJSONObjectMember(options, "compileList");
 
@@ -73,7 +71,6 @@ public class WebSocketServer {
     this.logger.setUseParentHandlers(false);
 
     Properties.setConnectionId(connectionId);
-    Properties.setIsDashboardLocalhost(true);
 
     websocketOutputAdapter = new WebSocketOutputAdapter(session);
     inputAdapter = new WebSocketInputAdapter();
--- a/org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketServer.java
+++ b/org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketServer.java
@@ -73,6 +73,7 @@ public class WebSocketServer {
     this.logger.setUseParentHandlers(false);
 
     Properties.setConnectionId(connectionId);
+    Properties.setIsDashboardLocalhost(true);
 
     websocketOutputAdapter = new WebSocketOutputAdapter(session);
     inputAdapter = new WebSocketInputAdapter();
--- a/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/DashboardConstants.java
+++ b/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/DashboardConstants.java
@@ -1,0 +1,10 @@
+package org.code.javabuilder;
+
+public final class DashboardConstants {
+  private DashboardConstants() {
+    throw new UnsupportedOperationException("Instantiation of constants class is not allowed.");
+  }
+
+  public static final String DASHBOARD_LOCALHOST_URL = "https://localhost-studio.code.org";
+  public static final String DASHBOARD_DOMAIN_SUFFIX = "studio.code.org";
+}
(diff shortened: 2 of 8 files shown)


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/DashboardConstants.java
```diff
@@ -0,0 +1,10 @@
+package org.code.javabuilder;
+
+public final class DashboardConstants {
+  private DashboardConstants() {
+    throw new UnsupportedOperationException("Instantiation of constants class is not allowed.");
+  }
+
+  public static final String DASHBOARD_LOCALHOST_URL = "https://localhost-studio.code.org";
```
jmkulwik, line 8: I think my localhost uses `http` not `https` - can you do this in a way that supports either?

jmkulwik: Oh, actually, I'm realizing that we prepend `https` in `LambdaRequestHandler` - perhaps just removing that will solve things?

sanchitmalhotra126: Oh good catch, so does mine - I think this is here because we're manually adding an "https://" in LambdaRequestHandler, because I think previously it was a safe assumption that all incoming connections would be https, but not anymore. I'll just get rid of the protocol, since it's not relevant for this use case.

org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketServer.java
```diff
@@ -73,6 +73,7 @@ public void onOpen(Session session) {
     this.logger.setUseParentHandlers(false);
 
     Properties.setConnectionId(connectionId);
+    Properties.setIsDashboardLocalhost(true);
```
jmkulwik, line 76: Do we want to set this here? The localhost websocket server should be able to fetch image assets from localhost dashboard

sanchitmalhotra126: So this won't actually affect the localhost websocket server since it uses `LocalContentManager` which will always try to fetch the asset URL - the stubbing logic is only in `AWSContentManager`

jmkulwik: Thinking out loud a bit here. Once we get outside the WebSocketServer class or LambdaRequestHandler class, it's a bit unclear what this property gets used for. The property is set globally, so in the future, I could see someone using it in some of the shared code, expecting it to consistently indicate that dashboard is inaccessible. But that wouldn't be the case for requests originating from the local WebSocket server. Perhaps naming this something like `InaccessibleDashboard` or `DashboardReachable` and then setting it accordingly would clarify intent.

molly-moen (approved):
LGTM!

org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketServer.java
sanchitmalhotra126: Hmm (also thinking out loud here) I think my worry with this is that "inaccessible" is a state derived from certain combinations of dashboard and Javabuilder, whereas whether the request originated from localhost dashboard is only dependent on dashboard. We have:
localhost Dashboard + websocket server = accessible
localhost Dashboard + deployed Javabuilder = inaccessible
adhoc/staging/prod dashboard + deployed Javabuilder = accessible
adhoc/staging/prod dashboard + websocket server (would we ever have this scenario?) = accessible

I guess the question is, does that logical reduction belong in the handler/Properties? Currently that logic is encapsulated in AWSContentManager. I could go either way on this - do we think it's more useful to store this accessible/inaccessible state in Properties, or will knowing that the request originated from localhost dashboard be useful by itself for other situations with different constraints?

jmkulwik: Hmm maybe this is more of a special one-time-use case. I can't off the top of my head think of another situation where we'd want to know whether or not the request came from localhost once we're on the buildAndRun Lambda/WebSocket server. As a side note, I don't _think_ we'd ever set up that 4th scenario you listed. It'd be too much work without much gain.

I'm ok with whichever option you end up going with. To throw one more suggestion in the mix: Perhaps just leaving the property unset from the WebSocket server would solve for both our concerns. Then, if we get a future scenario where having this detail on the WebSocket server would be useful, we can just add it, but we also are less likely to add it without a more full understanding of what it actually does.

sanchitmalhotra126: Gotcha, that makes sense. I think in that case I'll err on the side of being descriptive here and call this something like `canAccessDashboardAssets` so its intended use is clear and that way we can also leave it unset in the WebSocketServer.

PR #250, opened 2022-03-29
Allow lambdas to create, read, update javabuilder dynamodb tables

bencodeorg:
Allow the lambda execution role to create, read, and update DynamoDB records for tables named with javabuilder-specific suffixes. Must be merged and deployed [before this PR is deployed](https://github.com/code-dot-org/javabuilder/pull/251).

## Testing story

I've tested the `_tokens` table permission, but added in the other tables as well so that we don't need infra to deploy the IAM stack repeatedly as we work on throttling.


jmkulwik (approved):
LGTM

PR #251, opened 2022-03-30
Log tokens when used and verify 

bencodeorg:
1. **HTTP authorizer**: Logs to dynamodb each time a token is used by our HTTP authorizer. Sets the record to be deleted automatically two minutes after it is created. Checks that this token has not been used before. Temporarily always sets `vetted` to true on this record -- follow-up work to actually implement vetting.
2. **Websocket authorizer**: Gets the token's dynamodb record, and checks that it has a) passed through the HTTP authorizer, b) not been used before, and c) been vetted by the HTTP authorizer as not being to be throttled (to implement). Logs to CloudWatch if any of these are observed to be false. Sets `used` to true on the token's record in dynamodb regardless of whether all of the above is true.

## Testing story

I tested this manually -- at each of the logging steps, I set the token in a way that would produce the appropriate logging message and observed the appropriate logging message in CloudWatch. With a normal javabuilder run against a dev deploy, I observed normal behavior.

## Follow-up work

The next step here is to add user- and teacher-level throttling logic (ie, actually set the `vetted` bool appropriately). There may be a bit of refactor work to pull out some of the shared dynamodb logic, but wanted to get this out before doing that.


javabuilder-authorizer/http_authorizer_function.rb
```diff
@@ -15,5 +18,45 @@ def lambda_handler(event:, context:)
...
+    table_name: ENV['token_status_table'],
+    key: {token_id: sid},
+    update_expression: 'SET vetted = :v',
+    expression_attribute_values: {':v': true}
+  )
+
+  TokenStatus::VALID_HTTP
+end
+
+# ARN is of the format arn:aws:lambda:{region}:{account_id}:function:{lambda_name}
```
bencodeorg, line 59: I think I'm going to move some stuff into a `ThrottlingHelper` or something like that, this function is repeated and might be one of those things.

template.yml.erb
```diff
@@ -731,13 +731,13 @@ Resources:
     Properties:
       TableName: !Sub "${SubDomainName}_tokens"
       KeySchema:
-        - AttributeName: <%=DOMAIN_AND_TOKEN_ID_COMPOSITE_ATTRIBUTE_NAME%>
+        - AttributeName: <%=TOKEN_ID_ATTRIBUTE_NAME%>
```
bencodeorg, line 734: Realized we don't need domains in token IDs, so updating variable name here.

javabuilder-authorizer/http_authorizer_function.rb
jmkulwik: Suresh left a note for me about this on one of my previous PRs, but I didn't get a chance to implement it. It sounds like there's an API that may help here. His notes: https://github.com/code-dot-org/javabuilder/pull/224#discussion_r821973035

javabuilder-authorizer/token_status.rb
```diff
@@ -0,0 +1,11 @@
+module TokenStatus
+  # Token statuses used in HTTP authorizer
+  VALID_HTTP = 'valid_http'.freeze
+  ALREADY_EXISTS = 'already_exists'.freeze
+
+  # Token statuses used in websocket authorizer
+  VALID_WEBSOCKET = 'valid_websocket'.freeze
+  UNKNOWN_ID = 'unknown_id'.freeze
+  NOT_VETTED = 'not_vetted'.freeze
+  USED = 'used'.freeze
```
jmkulwik, line 10: NIT: Mind adding a quick comment to each of these statuses explaining what they mean? It's clear now, but I'm guessing a little more explanation will help us out in the future.

javabuilder-authorizer/http_authorizer_function.rb
```diff
@@ -15,5 +18,45 @@ def lambda_handler(event:, context:)
...
+  return JwtHelper.generate_deny(route_arn) unless token_status == TokenStatus::VALID_HTTP
+
+  JwtHelper.generate_allow(route_arn, token_payload)
+end
+
+def get_token_status(context, sid)
+  client = Aws::DynamoDB::Client.new(region: get_region(context))
+
+  begin
+    ttl = Time.now.to_i + 120
```
molly-moen, line 33: nit: add a constant for 120 with a name that includes the units of that number (seconds I'm guessing?)

javabuilder-authorizer/http_authorizer_function.rb
```diff
@@ -15,5 +18,45 @@ def lambda_handler(event:, context:)
...
+      table_name: ENV['token_status_table'],
+      item: {
+        token_id: sid,
+        ttl: ttl
+      },
+      condition_expression: 'attribute_not_exists(token_id)'
+    )
+  rescue Aws::DynamoDB::Errors::ConditionalCheckFailedException
+    puts "TOKEN VALIDATION ERROR: #{TokenStatus::ALREADY_EXISTS} token_id: #{sid}"
+    return TokenStatus::VALID_HTTP
```
molly-moen, line 44: should there be a todo here to not return valid http at some point?

molly-moen (approved):
LGTM, just a couple nits

jmkulwik (approved):

javabuilder-authorizer/token_status.rb
bencodeorg: Can do!

PR #252, opened 2022-03-30
Add Performance Tracking

jmkulwik:
This adds logs that allow us to track the performance of our lambdas without the variability of student code. With this we'll be able to measure the following:

1. Cold boot time
2. Time between initiating the lambda and compiling student code
3. Time between compiling and running student code
4. Time spent cleaning up after running student code

We could also measure the time spent compiling as well as the time spent running student code with these if we wanted to, but it may not tell us too much.

This also allows us to differentiate between time spent doing work during the first time a lambda container is used and during subsequent invocations.

**Note**: I'm actually not too happy with the way this change ended up being structured. It feels a little hacky to pass the performance tracker around to all the various classes that need to track performance. I debated putting it in GlobalProtocol, but that also felt odd since GlobalProtocol is intended for objects that are shared between our build-and-run logic and the student-facing libraries. I also thought about using the global logger, but that didn't really meet our needs - it made the cloud watch metrics very verbose and couldn't be used to generate a single performance report at the end of the run. If folks have ideas here, I'd love to hear them. 

Here's what the logs look like in CloudWatch

And here's an example query we could run with them
```
fields @message
| filter level = "INFO"
| filter message.type = "performanceReport"
| fields message.coldBootEnd - message.coldBootStart as coldBootTime
| fields message.compileStart - message.instanceStart as setupTime
| fields message.userCodeStart - message.compileEnd as transitionTime
| fields message.instanceEnd - message.userCodeEnd as cleanupTime
| fields setupTime + transitionTime as ourTime
#| stats pct(ourTime, 95) as ninetyfive, pct(ourTime, 80) as eighty, pct(ourTime, 70) as seventy, pct(ourTime, 60) as sixty, pct(ourTime, 50) as fifty
| display coldBootTime, setupTime, transitionTime, cleanupTime, ourTime
```

Which results in this chart (which I believe could be turned into a nicer graph, but would need more time playing around with it)


org-code-javabuilder/lib/src/main/java/dev/javabuilder/LocalLogHandler.java
```diff
@@ -10,22 +10,18 @@
...
   private final String channelId;
-  private final String miniAppType;
 
-  public LocalLogHandler(
-      PrintStream logStream, String levelId, String channelId, String miniAppType) {
+  public LocalLogHandler(PrintStream logStream, String levelId, String channelId) {
     this.logStream = logStream;
     this.levelId = levelId;
     this.channelId = channelId;
-    this.miniAppType = miniAppType;
```
jmkulwik, line 20: Unrelated to this change, I noticed that we're still logging the miniAppType, which is no longer relevant since we have separate logs for each of the mini apps.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/PerformanceTracker.java
```diff
@@ -0,0 +1,60 @@
...
+
+import static org.code.protocol.LoggerNames.MAIN_LOGGER;
+
+import java.time.Clock;
+import java.time.Instant;
+import java.util.logging.Logger;
+import org.code.protocol.LoggerConstants;
+import org.json.JSONObject;
+
+public class PerformanceTracker {
```
molly-moen, line 11: instead of passing around an instance of performance tracker, could we instead follow a similar pattern to global protocol and have a `getInstance()` method on `PerformanceTracker` that we call whenever we need to add a performance metric? We would also want some sort of reset or clear function so we don't reuse the same tracker across multiple instances.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/PerformanceTracker.java
```diff
@@ -0,0 +1,60 @@
...
+  public PerformanceTracker() {
+    logs = new JSONObject();
+  }
+
+  public void trackStartup(
+      Instant coldBootStart, Instant coldBootEnd, Instant instanceStart, boolean firstInstance) {
+    logs.put(COLD_BOOT_START, coldBootStart.toEpochMilli());
+    logs.put(COLD_BOOT_END, coldBootEnd.toEpochMilli());
+    logs.put(INSTANCE_START, instanceStart.toEpochMilli());
+    logs.put(FIRST_INSTANCE, firstInstance);
```
sanchitmalhotra126, line 33: If firstInstance = false, does this technically qualify as a cold boot? From what I can well, we're reporting this metric every time the handler is invoked right?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/PerformanceTracker.java
```diff
@@ -0,0 +1,60 @@
...
+    logs.put(USER_CODE_END, Clock.systemUTC().instant().toEpochMilli());
+  }
+
+  public void trackInstanceEnd() {
+    logs.put(INSTANCE_END, Clock.systemUTC().instant().toEpochMilli());
+  }
+
+  public void logPerformance() {
+    logs.put(LoggerConstants.TYPE, TYPE);
+    Logger.getLogger(MAIN_LOGGER).info(logs.toString());
```
sanchitmalhotra126, line 58: Do we want to log durations for each operation as well? I know in CloudWatch we can write some custom queries to view durations, but I'm not 100% sure if we can use those to generate graphs easily, so it may be worth logging those here too?

sanchitmalhotra126 (commented):
Looks good overall, just a couple comments. +1 to Molly's suggestion about making this a singleton. I agree that it doesn't belong on GlobalProtocol or the global Logger, but making it a singleton will prevent us from needing to plumb it through a bunch of classes.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/PerformanceTracker.java
jmkulwik: Done!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/PerformanceTracker.java
jmkulwik: Adjusted so it only reports on the first usage of the container

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/PerformanceTracker.java
jmkulwik: Fixed. Thank you!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
```diff
@@ -173,6 +186,8 @@ public String handleRequest(Map<String, String> lambdaInput, Context context) {
     } finally {
       // Stop timeout listener and clean up
       timeoutNotifierThread.interrupt();
+      PerformanceTracker.getInstance().trackInstanceEnd();
+      PerformanceTracker.getInstance().logPerformance();
```
sanchitmalhotra126, line 190: Do we need to call these methods anytime before we call `System.exit()` as well? I think we'd just need add these in `AWSSystemExitHelper` and here above line 160. However, I can also see us intentionally not report these if we want to take their absence to mean an abnormal shutdown?

sanchitmalhotra126 (approved):
LGTM! A few small comments but nothing blocking

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/PerformanceTracker.java
```diff
@@ -0,0 +1,113 @@
...
+  private static final String TRANSITION_TIME = "transitionTime";
+  private static final String USER_CODE_END = "userCodeEnd";
+  private static final String INSTANCE_END = "instanceEnd";
+  private static final String CLEANUP_TIME = "cleanupTime";
+  private static final String TYPE = "performanceReport";
+  private final JSONObject logs;
+  private static PerformanceTracker instance;
+
+  private PerformanceTracker() {
+    logs = new JSONObject();
```
molly-moen, line 39: nit: `this.logs`

molly-moen (approved):
LGTM!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
jmkulwik: Oh yes, we definitely want this to be considered part of the cleanup logic. I'll go ahead and add this to those two areas. This also is a pretty clear indicator that our error handlining and cleanup logic has fractured pretty significantly. I went ahead and moved the [cleanup task](https://codedotorg.atlassian.net/browse/JAVA-531) up to "Up Next."

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/PerformanceTracker.java
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  public static PerformanceTracker getInstance() {
+    if (PerformanceTracker.instance == null || PerformanceTracker.instance.logs == null) {
```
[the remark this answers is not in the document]
jmkulwik: Good point. I don't think it would be null. I included this here after accidentally setting logs to null in a different location (I must not have had it marked final at that point yet). Are you thinking that bit should be removed?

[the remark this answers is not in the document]
sanchitmalhotra126: Yeah, I think it should be safe to remove since in theory there shouldn't be a situation in which `logs` get set to null, unless we want to guard against future changes?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSSystemExitHelper.java
```diff
@@ -19,6 +19,8 @@ public AWSSystemExitHelper(String connectionId, AmazonApiGatewayManagementApi ap
 
   @Override
   public void exit(int status) {
+    PerformanceTracker.getInstance().trackInstanceEnd();
+    PerformanceTracker.getInstance().logPerformance();
```
jmkulwik, line 23: Alright, adding the performance tracker here ended up ballooning the PR a bit. Explanation for the additional changes:

In order to verify that the perf tracking was set up correctly for a system.exit situation, I tested out with our most recent occasion where we added a new System.exit. Namely, when a student recursively opens a file. However, I was blocked on testing this because we were sending the generic "We hit an error on our side" message to the user in this case. 

I spent some time investigating whether this was caused by my change and realized that instead, it was caused because we were unable to use the error handling classes because the class loader couldn't load them in the tmp directory due to the open file handles in the student's project. I was able to patch this by pre-loading those classes (See MainRunner.run).

However, I continued to hit an error where we were logging `SEVERE` errors to CloudWatch. Given that these cause alerts to show up in CloudWatch and Slack, I took a deeper look into these to see how these were being caused. I realized that since these are triggered by a rather innocuous-looking `logException` method, that method was being used in locations that shouldn't be triggering alerts. I was able to patch this by splitting `logException` into `logSevereException` and `logTrackingException`. (see LoggerUtils.logSevereException).

jmkulwik: @sanchitmalhotra126 @molly-moen - I could use a second glance since this added quite a bit to the PR. (see note)

molly-moen (approved):

sanchitmalhotra126 (commented):
Sorry for the late review (I know this is already merged) but this looks good! Thanks for updating the error logging methods. I think this does further highlight the need for us to revisit our overall exception handling strategy like you mentioned.

One thing I wanted to mention - I noticed we're only preloading the `UserInitiatedException` and `UserInitiatedExceptionKeys` classes. Do we know if loading these classes specifically was causing the issue or was this just the first class we tried to load? I'm just wondering if we'd still hit an issue with loading other classes if we hit the same out of space error.

PR #254, opened 2022-03-30
Ensure AWS resources are cleaned up before System.exit

sanchitmalhotra126:
Another quick fix to ensure that we always cleanup AWS resources before calling System.exit(). This is a bit of a one-off solution right now because the AWS cleanup code is copied between LambdaRequestHandler and the new AWSSystemExitHelper, but I figured this can help us work towards a more robust solution and also expose a place for us to add any necessary cleanup code in the case of an abnormal system exit.

Tested using dev lambda.

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
+++ b/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
@@ -156,7 +156,8 @@ public class LambdaRequestHandler implements RequestHandler<Map<String, String>,
             compileList,
             tempDirectoryManager,
             contentManager,
-            lifecycleNotifier);
+            lifecycleNotifier,
+            new AWSSystemExitHelper(connectionId, API_CLIENT));
 
     final Thread timeoutNotifierThread =
         createTimeoutThread(context, outputAdapter, codeExecutionManager, connectionId, API_CLIENT);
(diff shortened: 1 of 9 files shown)


org-code-javabuilder/protocol/src/main/java/org/code/protocol/SystemExitHelper.java
```diff
@@ -0,0 +1,10 @@
+package org.code.protocol;
```
jmkulwik, line 1: Should this be in the org.code.javabuilder package? It seems unlikely that we'd use it in the student-facing APIs.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
```diff
@@ -156,7 +156,8 @@ public String handleRequest(Map<String, String> lambdaInput, Context context) {
             compileList,
```
jmkulwik, line 156: Worth also using the systemExitHelper on line 147?

org-code-javabuilder/protocol/src/main/java/org/code/protocol/SystemExitHelper.java
sanchitmalhotra126: Oh yep good point, I'll move it

jmkulwik (approved):
LGTM!

PR #255, opened 2022-03-30
Update output adapter for validation runs

molly-moen:
We need to set `isValidation` on `UserTestOutputAdapter` to true on validation runs and to false on user test runs. I missed adding this in the previous work to set up validation runs. This allows us to track neighborhood requests and hide system out messages from the end user when in a validation run, and still send system out messages on user test runs.

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/BaseTestRunner.java
+++ b/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/BaseTestRunner.java
@@ -7,6 +7,7 @@ import org.code.protocol.InternalErrorKey;
 import org.code.protocol.OutputAdapter;
 import org.code.protocol.StatusMessage;
 import org.code.protocol.StatusMessageKey;
+import org.code.validation.support.UserTestOutputAdapter;
 import org.junit.platform.commons.PreconditionViolationException;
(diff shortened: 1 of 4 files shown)


jmkulwik (approved):
LGTM!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/BaseTestRunner.java
```diff
@@ -22,16 +23,19 @@
...
   private final OutputAdapter outputAdapter;
   private final StatusMessageKey statusMessageKey;
+  private final boolean isValidation;
 
   public BaseTestRunner(
       List<JavaProjectFile> files,
       JavabuilderTestExecutionListener listener,
       OutputAdapter outputAdapter,
-      StatusMessageKey statusMessageKey) {
+      StatusMessageKey statusMessageKey,
```
sanchitmalhotra126, line 32: Could we just use the value of `statusMessageKey` to tell if this is validation or not?

molly-moen: good idea. I think I might actually remove `statusMessageKey` here and use `isValidation` to pick the key.

sanchitmalhotra126 (approved):
LGTM!

PR #257, opened 2022-03-31
Create ECR Repository

cat5inthecradle:
Creates an ECR Repository for storing docker images for this pipeline.

## Deployment

Run the "cicd/2-cicd/deploy-cicd.sh" script. Elevated permissions should not be required.


molly-moen (approved):
🎉

PR #258, opened 2022-03-31
Load Testing AWS Setup

molly-moen:
This PR contains a new CloudFormation template for setting up the resources required for load testing (besides the [ECR](https://aws.amazon.com/ecr/) which is created in [this template](https://github.com/code-dot-org/javabuilder/blob/900a9f222ec6bfa31c29d1087ec581808e3490eb/cicd/2-cicd/cicd.template.yml#L23)). The resources in this template are:
- An ECS Cluster
- An ECS Task definition for running the load test docker image
- A CloudWatch logs group to put the logs from the task in
- An IAM Role the task uses for fetching secrets and running the task

I've tested this by using the deploy script included in this PR and am able to deploy the stack and then manually run the task in the AWS console. Since we will be doing load testing on an as-needed basis the plan is to trigger the tasks from the AWS console

This PR also adds the new load testing public keys to the cloudformation template for javabuilder, which we will only use for load testing instances of Javabuilder. The work to actually use the keys will come later.


cicd/3-app/load-test.template.yml
```diff
@@ -0,0 +1,87 @@
...
+  LoadTestImage:
+    Type: String
+    Description: Load testing docker image. This should be the full name of the image.
+Resources:
+  ECSCluster:
+    Type: 'AWS::ECS::Cluster'
+    Properties:
+      CapacityProviders:
+        - FARGATE
+        - FARGATE_SPOT
```
cat5inthecradle, line 13: Are you looking to use `FARGATE SPOT`?

cicd/3-app/load-test.template.yml
```diff
@@ -0,0 +1,87 @@
...
+              awslogs-stream-prefix:  !Sub "${AWS::StackName}-task"
+      Cpu: 256
+      ExecutionRoleArn: !Ref ExecutionRole
+      Memory: 512
+      NetworkMode: awsvpc
+      RuntimePlatform: 
+        OperatingSystemFamily: LINUX
+      Tags: 
+        - Key: "javabuilder-mode"
+          Value: "load-test"
```
cat5inthecradle, line 59: We can add tags to the `cloudformation deploy` step and it will apply the same set of tags to all resources in the stack.

cicd/3-app/deploy-load-test.sh
```diff
@@ -0,0 +1,31 @@
...
+aws cloudformation validate-template \
+  --template-body file://${TEMPLATE_FILE}
+
+echo Updating cloudformation stack...
+aws cloudformation deploy \
+  --stack-name $STACK_NAME \
+  --template-file $TEMPLATE_FILE \
+  --parameter-overrides LoadTestImage=$LOAD_TEST_IMAGE \
+  --capabilities CAPABILITY_IAM \
+  "$@"
```
cat5inthecradle, line 29:
```suggestion
  --capabilities CAPABILITY_IAM \
  --tags javabuilder-mode=load-test \
  "$@"
```

jmkulwik:
Why does the ECR get set up as part of the CICD template? I would have thought it would go in the main template.

cicd/3-app/load-test.template.yml
molly-moen: I think this was a copy/paste error, I will remove it.

molly-moen:
> Why does the ECR get set up as part of the CICD template? I would have thought it would go in the main template.

@jmkulwik we need the ECR to already exist and the docker image for the load test to have been pushed before we create the task definition, so we need two separate steps.

jmkulwik:
> > Why does the ECR get set up as part of the CICD template? I would have thought it would go in the main template.
> 
> @jmkulwik we need the ECR to already exist and the docker image for the load test to have been pushed before we create the task definition, so we need two separate steps.

Ahh ok. I didn't realize that all of this was going in the CICD templates. Maybe a follow-up question then: Since we won't be doing load testing on every PR, it seems odd that we'd be deploying it as part of CICD. Why include it here rather than in its own template?

cicd/3-app/load-test.template.yml
molly-moen: done!

jmkulwik (approved):

cat5inthecradle (approved):

PR #259, opened 2022-03-31
Basic load test script and setup

molly-moen:
This PR creates the basic script needed for running load testing with [k6](https://k6.io/docs/), along with a Dockerfile for creating a docker image using that script. The docker image can be used in conjunction with [these AWS resources](https://github.com/code-dot-org/javabuilder/pull/258) to run the image on ECS, or it can be run locally using the instructions in the readme.

This does not contain the auth we plan to use long-term for load testing, as that was a large enough work item to get split into its own PR.

The load test in this PR runs a simple hello world project. For the first 30 seconds it ramps up to 5 virtual users, then it has 5 virtual users run 3 tests each. It measures the number of exceptions, the number of errors, the length of each request, and reports how many sessions were greater than 5 seconds and how many were greater than 10 seconds. Right now it is set up to run against my dev instance, as we don't yet have the load testing instance up and running.


jmkulwik (approved):
Nice work setting this up

load-test/README.md
```diff
@@ -0,0 +1,14 @@
...
+This folder contains scripts and helpers for load testing.
+
+## One time set up
+- [Install Docker](https://docs.docker.com/get-docker/)
+
+## How to run
+### Local Setup
+Note: this will not work once throttling in enabled.
+- Ensure docker is running. This usually entails opening Docker Desktop.
+- On localhost code studio, increase the token timeout to the length of time you want to run your test.
```
sanchitmalhotra126, line 11: nit: add link to the line in code-dot-org for where to make this change

load-test/scripts/hello-world-load-test.js
```diff
@@ -0,0 +1,70 @@
...
+    }
+  });
+
+  socket.on("close", () => {
+    const time = Date.now() - startTime;
+    connectToCloseTime.add(time);
+    if (time > 5000) {
+      longWebsocketSessions.add(1);
+    }
+    if (time > 10000) {
```
sanchitmalhotra126, line 60: Would it be worth pulling up these values somewhere as constants either in this file, or another if we're planning to share thresholds between multiple tests?

load-test/scripts/configuration.js
```diff
@@ -0,0 +1,52 @@
...
+      iterations: 3,
+      maxDuration: "2m",
+      startTime: '30s',
+    }
+  },
+  thresholds: {
+    exceptions: ["count == 0"],
+    errors: ["count == 0"],
+    websocket_session_duration_without_sleep: ["p(95) < 5000"],
+    long_websocket_sessions: ["count <= 1"],
```
sanchitmalhotra126, line 25: I'm guessing the count <= 1 is to account for cold boot time right? Is there a separate threshold for that we're also tracking, or not part of this test?

sanchitmalhotra126 (approved):
Nice! Couple of minor comments/questions but nothing blocking.

load-test/scripts/configuration.js
molly-moen: yes, there is a threshold for 'extra long websocket sessions'. These are not final thresholds yet, we'll probably need to fine tune to account for the fact that each concurrent instance has its own cold boot time.

load-test/scripts/hello-world-load-test.js
molly-moen: I'll move them into the configuration file

PR #260, opened 2022-04-01
Throttling: add checks for user- and classroom-level limits

bencodeorg:
More updates to our throttling work. Does not actually prevent any access to Javabuilder currently, but adds checks for:
- whether user is blocked
- whether user's classrooms are blocked
- user hourly limit
- user daily limit
- classroom hourly limit

If these all pass (they all do for now), we:
- log the user request
- log the a request for each teacher the user is associated with

~~We do not have a classroom-level limit yet for throttling -- the plan here is to set the expiration time (ttl) for `teacher_associated_requests` to two weeks, such that we can query that raw data and see what a reasonable hourly limit might be. Once we have a throttling limit we feel good about, we can set that value and the ttl to two hours for this table.~~
Update: ended up using Cloudwatch logs to get a sense of what our hourly historical data looks like per teacher, and set a limit (can be updated) accordingly.

## Testing story

Heavy manual testing -- went through each condition and looked at the CloudWatch logs generated and DynamoDB records created, as well as confirming normal behavior in local javabuilder. See follow-up for more on testing.
## Follow-ups

- We need unit tests for this code 😬 -- I will look into what it would take to get basic unit tests set up for our authorizer code. Longer term, it might be nice to have some integration tests with dynamodb, but that is a bigger piece of work.
- [Refactor websocket authorizer code](https://github.com/code-dot-org/javabuilder/blob/main/javabuilder-authorizer/websocket_authorizer_function.rb) to use the new `TokenValidator` helper as well.
- We'll be querying DynamoDB figure out what a reasonable classroom limit is, and see whether any students are crossing the hourly and daily thresholds we've set currently. Then, we can actually turn on throttling!

### DIFF
--- a/javabuilder-authorizer/http_authorizer_function.rb
+++ b/javabuilder-authorizer/http_authorizer_function.rb
@@ -41,7 +41,7 @@ def get_token_status(context, token_payload, origin)
   puts TokenStatus::USER_OVER_HOURLY_LIMIT if validator.user_over_hourly_limit?
   puts TokenStatus::USER_OVER_DAILY_LIMIT if validator.user_over_daily_limit?
   puts TokenStatus::TEACHERS_OVER_HOURLY_LIMIT if validator.teachers_over_hourly_limit?
-  validator.log_token
+  validator.log_requests
   validator.mark_token_as_vetted
 
   TokenStatus::VALID_HTTP
--- a/javabuilder-authorizer/http_authorizer_function.rb
+++ b/javabuilder-authorizer/http_authorizer_function.rb
@@ -23,13 +23,15 @@ def lambda_handler(event:, context:)
   return JwtHelper.generate_deny(route_arn) unless decoded_token
 
   token_payload = decoded_token[0]
-  token_status = get_token_status(context, token_payload['sid'])
+  puts token_payload
+  token_status = get_token_status(context, token_payload)
   return JwtHelper.generate_deny(route_arn) unless token_status == TokenStatus::VALID_HTTP
 
   JwtHelper.generate_allow(route_arn, token_payload)
 end
 
-def get_token_status(context, sid)
+def get_token_status(context, token_payload)
+  sid = token_payload['sid']
   client = Aws::DynamoDB::Client.new(region: get_region(context))
 
   begin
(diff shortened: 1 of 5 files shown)


javabuilder-authorizer/http_authorizer_function.rb
```diff
@@ -19,45 +20,29 @@ def lambda_handler(event:, context:)
...
+  validator = TokenValidator.new(token_payload, origin, region)
+
+  return validator.error_message unless validator.log_token
+  return validator.error_message if validator.user_blocked?
+  return validator.error_message if validator.teachers_blocked?
+  return validator.error_message if validator.user_over_hourly_limit?
+  return validator.error_message if validator.user_over_daily_limit?
+  return validator.error_message if validator.teachers_over_hourly_limit?
+  validator.log_requests
+  return validator.mark_token_as_vetted
```
sanchitmalhotra126, line 45: nit: should this be moved to `TokenValidator` if this is going to be shared code?

Also, just a thought - I feel a little iffy on the structure of this logic. We have an assumption that the various check methods (user_blocked, etc), will internally set the @status variable that gets returned by error_message. This is internal code so it's not a huge deal, but in theory if one of these methods forgets to set the @status variable we'd be returning the wrong status value here and it may be a little tricky to nail down why. Could we just return statuses directly instead of setting them and reading them back in error_message? Something like this?

```
return TokenStatus::USER_BLOCKED if validator.user_blocked?
return TokenStatus::TEACHERS_BLOCKED if validator.teachers_blocked?
...
validator.mark_token_as_vetted
return TOKEN_STATUS::VALID_HTTP
```

Having this block inside TokenValidator may also help clean it up a bit since we'd be able to call the various checking functions directly (or even make them private)

sanchitmalhotra126 (commented):
One comment but otherwise looks pretty good to me!

javabuilder-authorizer/http_authorizer_function.rb
bencodeorg: haha yeah a couple thoughts:
- I had it set up something like what you're describing originally (return statuses directly depending on result of each function call). I thought it'd be better to explicitly link the status and the function call that determines that status by moving the status internally, but yeah now I see how that could be confusing.
- re: "move to `TokenValidator` if this is going to be shared code", do you mean just moving this whole set of code into `TokenValidator`, or something else? I was thinking I could just add something like `validator.validate` that does all of this and just returns the validation status.

bencodeorg: Do you prefer something like this (no @status variable):

```
  def validate
    return error_message(ALREADY_EXISTS) unless log_token
    return error_message(USER_BLOCKED) if user_blocked?
    return error_message(TEACHERS_BLOCKED) if teachers_blocked?
    return error_message(USER_OVER_HOURLY_LIMIT) if user_over_hourly_limit?
    return error_message(USER_OVER_DAILY_LIMIT) if user_over_daily_limit?
    return error_message(TEACHERS_OVER_HOURLY_LIMIT) if teachers_over_hourly_limit?
    log_requests
    mark_token_as_vetted
    VALID_HTTP
  end
```

sanchitmalhotra126: Yeah this looks good! Yeah my concern was just that if we add another check but forget to set the @status variable in that function, we'd end up with a weird state

bencodeorg: Makes sense, give it another look when you have a chance!

sanchitmalhotra126 (approved):
LGTM!

molly-moen (approved):

PR #261, opened 2022-04-04
Warn instead of throwing exceptions for GlobalProtocol create/destroy errors

sanchitmalhotra126:
Quick fix to mitigate prod issue where Playground projects are running into exceptions for creating GlobalProtocol when it already exists.


molly-moen (approved):

jmkulwik (approved):
Thanks for the quick fix here!

PR #262, opened 2022-04-04
Remove unneeded environment variables

molly-moen:
We don't need these anymore and they made the lambda configuration too large.


jmkulwik (approved):
LGTM

sanchitmalhotra126 (approved):

PR #263, opened 2022-04-04
Add authentication to load tests

molly-moen:
Add authentication in the form of [JWTs](https://jwt.io/introduction) to load tests. These JWTs are generated slightly differently from our regular JWTs, because k6 does not support RS256 (public/private key pair). Instead, we will use HS256 (shared secret). See discussion [here](https://codedotorg.slack.com/archives/C03CK49G9/p1648845157645139) for more details. I have added a flag to Javabuilder as part of this that we can switch on to use a shared secret for load testing instances of Javabuilder.

Because we want the load testing script to be as small as possible we could not use an npm package to generate the JWT (k6 has minimal support for using npm packages, but all the ones for JWTs were too large). Therefore we need to generate it with k6's [crypto](https://k6.io/docs/javascript-api/k6-crypto/) functionality. 

In our initial setup we assumed we would need public/private key pairs. I removed the public keys in [this PR](https://github.com/code-dot-org/javabuilder/pull/262) and have removed the links to the private keys in this PR.

[Jira ticket](https://codedotorg.atlassian.net/browse/JAVA-516)


bencodeorg (approved):
LGTM!

sanchitmalhotra126 (approved):

PR #264, opened 2022-04-05
Switch from DynamoDB Provisioned Throughput to On-Demand

sureshc:
Each instance of Javabuilder, including development Stacks, is consuming about $2200/month in DynamoDB costs because we specify Provisioned read/write Throughput of 1000 units per table and there are 4 tables. Switch to [On-Demand mode](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.ReadWriteCapacityMode.html#HowItWorks.OnDemand).

Once we get a better sense of usage patterns, we could switch back to Provisioned Mode, and use DynamoDB Autoscaling so we don't pay for unused provisioned capacity during low usage time periods, and purchased Reserved Capacity for our production instance for lower costs in a system that has high sustained usage.

### DIFF
--- a/template.yml.erb
+++ b/template.yml.erb
@@ -707,8 +707,6 @@ Resources:
   TOKEN_ID_ATTRIBUTE_NAME = 'token_id'
   ISSUED_AT_TIMESTAMP_ATTRIBUTE_NAME = 'issued_at'
   TIME_TO_LIVE_ATTRIBUTE_NAME = 'ttl'
-  READ_CAPACITY_PER_SECOND = 1000
-  WRITE_CAPACITY_PER_SECOND = 1000
 -%>
   BlockedUsersTable:
     Type: AWS::DynamoDB::Table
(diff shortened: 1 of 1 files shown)


bencodeorg:
Yikes.

molly-moen (approved):

bencodeorg (approved):
Welp, obviously had no idea what this setting meant cost-wise. Thanks for fixing. A couple questions:
- I haven't historically with cost impacts when making cloudformation updates. Any best practices about how to incorporate that into the process?
- Is it possible to do cloudwatch-style alerts for cost increases?

sureshc:
> * I haven't historically with cost impacts when making cloudformation updates. Any best practices about how to incorporate that into the process?
> * Is it possible to do cloudwatch-style alerts for cost increases?

No worries. There are possibly some automated tools/processes we can incorporate to estimate the cost of a change. For example, [there's an API](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/CloudFormation/Client.html#estimate_template_cost-instance_method) to get the estimated monthly cost of a template.

A good manual practice, for now, is to get familiar with the pricing model for each service we're using. For most AWS services it typically simple and easy to keep in our heads as we're implementing a solution and thinking about tradeoffs between different technical designs.

bencodeorg:
Deployed this to my dev instance and ran a Playground project against it successfully. I see the updated setting on one of the DynamoDB tables (Capacity mode = "on demand")

sureshc:
Thank you for testing and deploying!

PR #265, opened 2022-04-06
Neighborhood API Updates

sanchitmalhotra126:
Includes new getX(), getY(), getDirection(), and setPaint() methods on the Painter class.

https://codedotorg.atlassian.net/browse/JAVA-529

### DIFF
--- a/org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/Painter.java
+++ b/org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/Painter.java
@@ -193,6 +193,39 @@ public class Painter {
     return this.direction.isWest();
   }
 
+  /** @return the x coordinate of the painter's current position */
+  public int getX() {
+    return this.xLocation;
+  }
+
+  /** @return the y coordinate of the painter's current position */
+  public int getY() {
+    return this.yLocation;
+  }
+
+  /** @return the current direction the painter is facing */
+  public String getDirection() {
+    return this.direction.getDirectionString();
+  }
+
+  /**
+   * Sets the amount of paint in the painters bucket. Does nothing if paint is negative.
+   *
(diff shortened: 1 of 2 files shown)


org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/Painter.java
```diff
@@ -193,6 +193,39 @@ public boolean facingWest() {
...
+  }
+
+  /**
+   * Sets the amount of paint in the painters bucket. Does nothing if paint is negative.
+   *
+   * @param paint the amount of paint that should be in the painter's bucket.
+   */
+  public void setPaint(int paint) {
+    if (paint < 0) {
+      System.out.println("Paint amount must not be a negative number.");
```
molly-moen, line 218: should this be an output adapter message instead so we can translate it?

sanchitmalhotra126: Hmm good question, I used System.out.println just to be consistent since we're using that elsewhere in Painter (ex. "There is no more paint in the painter's bucket"). It does feel like these should be translatable though given what we do in other mini-apps, so I can log an issue to translate all of these.

molly-moen: 👍 makes sense, we probably added those before we did translations

molly-moen (approved):

org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/Painter.java
sanchitmalhotra126: Sounds good, created https://codedotorg.atlassian.net/browse/JAVA-535 to track

PR #266, opened 2022-04-08
Split up javabuilder lambda iam roles

cat5inthecradle:
Role Requirements

* The Javabuilder BuildAndRun lambdas all need to be able to read from SQS and the Javabuilder S3 bucket and to write to the @connections endpoint on the API Gateway WebSocket API
* The Theater BuildAndRun lambda needs to be able to write to the Javabuilder S3 bucket and get signed URLs to the bucket
* The API Gateway Relay Message Lambda needs to be able to write to SQS=
* The API Gateway Start Session Lambda (which is currently the Relay Message Lambda) needs to be able to create an SQS queue and invoke the BuildAndRun Lambda
* The API Gateway Close Session Lambda (which is currently the Relay Message Lambda) needs to be able to delete an SQS queue
* The Authorizer Lambda (which may be split into two lambdas) needs to be able to authenticate the token, read and write to the Dynamo DB tables associated with Javabuilder, and pass a request on to the StartSession Lambda
* The PutSources Lambda needs to be able to write objects to the S3 bucket

## Deployment Strategy

Upon PR Approval:
- manually deploy iam stack, which will and exporting new roles
- merge this PR to update lambdas to use the new roles

Upon Merge
- create a cleanup PR to delete the old shared role (may require cleanup of existing stacks)

Revert Strategy
- revert this PR to have lambdas use original JavabuilderLamdaExecutionRole


iam.yml
```diff
@@ -10,13 +10,137 @@ Parameters:
...
+            Effect: Allow
+            Principal: {Service: [lambda.amazonaws.com]}
+      Policies:
+        - PolicyName: BuildAndRunLambdaExecutionPolicy
+          PolicyDocument:
+            Statement:
+              # BuildAndRunJavaProject Lambda needs to put objects to the content bucket.
+              - Effect: Allow
+                Action:
+                  - 's3:PutObject'
```
cat5inthecradle, line 35: Do we need additional permission to get signed URL's too?

iam.yml
```diff
@@ -10,13 +10,138 @@ Parameters:
...
+                  - 'dynamodb:GetItem'
+                  - 'dynamodb:PutItem'
+                  - 'dynamodb:UpdateItem'
+                  - 'dynamodb:Query'
+                Resource:
+                  - !Sub "arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/*_tokens"
+                  - !Sub "arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/*_user_requests"
+                  - !Sub "arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/*_teacher_associated_requests"
+                  - !Sub "arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/*_blocked_users"
+              # Allow invoking of the Build & Run Lambda
```
jmkulwik, line 108: They shouldn't need to invoke the build & run lambda. I'm not sure how API Gateway sets this up, but they might need to be able to invoke the StartSession(AndRelayMessage) lambda after authorizing the payload, but I think API Gateway is what actually invokes that lambda.

iam.yml
jmkulwik: Yes, but I don't see where we gave permission to do that in the past. (sorry, not the most helpful response)

iam.yml
cat5inthecradle: I don't see anything in the code for authorizer other than the dynamodb `put` and `update`, so I think I can scale down to just dynamodb permissions.

cat5inthecradle:
I've adjusted the roles again. I think what makes sense here is to deploy this new iam template. It does not delete the existing shared role, nor does it configure anything to use the new roles. It simply creates the roles.

Once they have been created, we can manually update lamdas in one of the developer environments to use the new permissions. @jmkulwik is there a good stack to do this with, or would it be easy to launch my own instance of javabuilder?

cat5inthecradle:
I've added the PutSources lambda role and limited the S3 permissions for the BuildAndRun lambda role.

iam.yml
sanchitmalhotra126: At the moment, I don't think Javabuilder is actually GET-ing any signed URLs? We read and write to the output bucket and generate signed urls for the client, but I don't think we ever read signed URLs for another bucket?

iam.yml
```diff
@@ -10,13 +10,159 @@ Parameters:
...
+    Type: String
+    Description: The API ID of the Javabuilder API Gateway V2 Api
+    Default: "*"
 
 Resources:
-  # TODO: Split this into separate roles:
-  # 1) Permissions for the long-running Lambda that compiles/runs student projects
-  # 2) Permissions for the synchronous Lambda that invokes the long-running Lambda and then relays web messages to it.
-  # 3) Permissions for the Lambda Authorizer
-  # 4) Permissions for CodePipeline Slack approval Lambda.
```
sanchitmalhotra126, line 19: nit: could we something like this around as a comment describing the new separate roles along with their role names?

sanchitmalhotra126 (approved):
LGTM!

iam.yml
cat5inthecradle: I've added comments describing each role to add more context

molly-moen (approved):
LGTM! I tested this out on my dev instance against our various allthethings levels and they all worked!

iam.yml
cat5inthecradle: for some reason I was thinking that you needed an explicit getPresignedUrl permission, but that's not a thing, it just uses your current Get/Put permissions.

PR #268, opened 2022-04-11
Copy image/sound before draw/play

bencodeorg:
New Theater API has a bug where if an image is drawn, then modified, the drawn image will reflect the modifications made after the draw.

This PR updates to make a copy of the image to put in the list of actions each time drawImage is called.

Ended up being a pretty short change, ~~hoping I'm not missing something else obvious that needs to be done.~~ Missed that we needed to make a similar update when using playSound, which I added to this PR.

## Testing story

I updated a unit test covering this case and tested manually that an image updated via setPixel after being drawn no longer reflects those changes on the initial draw.

### DIFF
--- a/org-code-javabuilder/theater/src/main/java/org/code/theater/Scene.java
+++ b/org-code-javabuilder/theater/src/main/java/org/code/theater/Scene.java
@@ -175,7 +175,11 @@ public class Scene {
    * @param rotation the amount to rotate the image in degrees
    */
   public final void drawImage(Image image, int x, int y, int size, double rotation) {
-    this.actions.add(new DrawImageAction(image, x, y, size, UNSPECIFIED, UNSPECIFIED, rotation));
+    // Copy the image to subsequent changes to this image object are not reflected
+    // in the image drawn here.
+    Image imageCopy = new Image(image);
+    this.actions.add(
+        new DrawImageAction(imageCopy, x, y, size, UNSPECIFIED, UNSPECIFIED, rotation));
   }
 
   /**
@@ -190,7 +194,10 @@ public class Scene {
    * @param rotation the amount to rotate the image in degrees
    */
   public final void drawImage(Image image, int x, int y, int width, int height, double rotation) {
-    this.actions.add(new DrawImageAction(image, x, y, UNSPECIFIED, width, height, rotation));
+    // Copy the image to subsequent changes to this image object are not reflected
+    // in the image drawn here.
+    Image imageCopy = new Image(image);
+    this.actions.add(new DrawImageAction(imageCopy, x, y, UNSPECIFIED, width, height, rotation));
   }
 
   /**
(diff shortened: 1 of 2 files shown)


org-code-javabuilder/theater/src/test/java/org/code/theater/SceneTest.java
```diff
@@ -65,7 +64,7 @@ public void testPause() {
 
   @Test
   public void testDrawImage() {
-    final Image image = mock(Image.class);
+    final Image image = new Image(1, 1);
```
bencodeorg, line 67: This was easier than other ways I could think of to confirm that the image being added to the action list is a copy of the existing image (assert that `new Image(Image source)` constructor was called? assert on `pixels` (2D array) field on `Image` object)?

bencodeorg:
Tangentially, could anyone explain the relationship between the `pixels` 2D array and `bufferedImage` fields on our Image objects?

org-code-javabuilder/theater/src/test/java/org/code/theater/SceneTest.java
sanchitmalhotra126: I see you're using `assertNotSame()` below - that should be enough to verify that this is a copy and not the original object right? Or is this to verify that all the properties of the image are still the same?

sanchitmalhotra126 (commented):
Nice! I think we'll hit the same issue for `playSound(double[] samples)` if the `samples` array is modified. Up to you though if you want to include that change in this PR or have that be a separate fix.

org-code-javabuilder/theater/src/test/java/org/code/theater/SceneTest.java
bencodeorg: Yeah I wanted to assert on some aspect of the image being the same 🤷

bencodeorg:
> Nice! I think we'll hit the same issue for `playSound(double[] samples)` if the `samples` array is modified. Up to you though if you want to include that change in this PR or have that be a separate fix.

Ooh I guess I will do that separately, sorry misunderstood at sprint planning and thought audio wasn't an issue.

bencodeorg:
> > Nice! I think we'll hit the same issue for `playSound(double[] samples)` if the `samples` array is modified. Up to you though if you want to include that change in this PR or have that be a separate fix.
> 
> Ooh I guess I will do that separately, sorry misunderstood at sprint planning and thought audio wasn't an issue.

Ooor maybe I'll just do it here, seems pretty quick.

org-code-javabuilder/theater/src/main/java/org/code/theater/Scene.java
```diff
@@ -75,7 +75,8 @@ public final void clear(Color color) {
    * @param sound an array of samples to play.
    */
   public final void playSound(double[] sound) {
-    this.actions.add(new PlaySoundAction(sound));
+    double[] soundCopy = sound.clone();
+    this.actions.add(new PlaySoundAction(soundCopy));
```
sanchitmalhotra126, line 79: nit: clone inline? `(new PlaySoundAction(sound.clone())`

molly-moen (approved):

PR #270, opened 2022-04-13
Add option to silence alerts

molly-moen:
Add an option to javabuilder deploys to silence alerts. This option will only apply to non-dev account instances (dev account instances will always have alerts silenced. To enable this option, add `SILENCE_ALERTS=true` to the deploy script command.
This option will allow us to silence alerts on the load test instance of Javabuilder.

### DIFF
--- a/template.yml.erb
+++ b/template.yml.erb
@@ -38,6 +38,11 @@ Parameters:
     Type: String
     Description: The default stage name in the API Gateway APIs
     Default: Prod
+  SilenceAlerts:
+    Type: String
+    AllowedValues: [true, false]
+    Description: If alerts should be silenced on this instance
+    Default: false
 <%
 JAVALAB_APP_TYPES = %w(
   Theater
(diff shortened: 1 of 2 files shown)


template.yml.erb
```diff
@@ -586,7 +592,7 @@ Resources:
           Then post in #ap-csa-dev.
       ActionsEnabled: true
       AlarmActions:
-          - !If [IsDevCondition, !Sub "arn:aws:sns:${AWS::Region}:${AWS::AccountId}:javabuilder-dev-test", !Sub "arn:aws:sns:${AWS::Region}:${AWS::AccountId}:CDO-Urgent"]
```
sanchitmalhotra126, line 589: Just curious, do you know if `javabuilder-dev-test` was setup to notify us?

sanchitmalhotra126 (approved):

template.yml.erb
molly-moen: it was setup to go to Jessie's email, which we decided was not ideal so we just silenced it. The alerts will still exist and you can look at the ones on a given instance to see if they have triggered, they just would have any action by default

PR #271, opened 2022-04-13
Remove local main

molly-moen:
As Javabuilder has become more complicated, `LocalMain` has lost its usefulness. This PR deletes `LocalMain` and any classes that only it used, as well as any references to LocalMain

Jira Ticket: [JAVA-227](https://codedotorg.atlassian.net/browse/JAVA-227)


org-code-javabuilder/lib/src/main/java/dev/javabuilder/CORSFilter.java
```diff
@@ -1,81 +0,0 @@
-package dev.javabuilder;
```
jmkulwik, line 1: I think CORSFilter gets used by the WebSocketServer version of javabuilder in order to provide prompter functionality and to get the theater files to the client.

molly-moen: oh good call out, will add back in

sanchitmalhotra126 (approved):
🎉

PR #272, opened 2022-04-14
Handle sqs create/delete throttles

molly-moen:
We found in load tests that at high load we can see SQS Queue create and delete operations get throttled. To resolve this issue this PR makes a few changes:
- set our retry mode to [adaptive](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/SQS/Client.html#initialize-instance_method) which is more robust than the default.
- If we see a request throttled error, retry the request up to 3 times. Before the first retry sleep 0.5 seconds, before the second sleep 1 second, and before retry 3 sleep for 2 seconds. After 3 retries throw the throttled exception and expect the client to retry.

With these changes we went from needing the client to retry 36% of the time during high load times (numbers here based on constant requests for 500 virtual users) to just 0.1% of the time!

## Potential Future Improvement
Long term we may want a fixed pool of SQS queues we can use instead of creating and deleting a queue for each request. That is a longer-term item so for now we will just retry the operation.

### DIFF
--- a/api-gateway-routes/api_gateway_proxy_function.rb
+++ b/api-gateway-routes/api_gateway_proxy_function.rb
@@ -4,6 +4,9 @@ require 'aws-sdk-lambda'
 require 'aws-sdk-apigatewaymanagementapi'
 require 'uri'
 
+MAX_SQS_RETRIES = 3
+INITIAL_RETRY_SLEEP_S = 0.5
+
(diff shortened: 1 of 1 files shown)


sanchitmalhotra126 (approved):
Nice! Awesome to see how much the client retry rate went down because of this!

PR #273, opened 2022-04-14
Add retries to load test

molly-moen:
Add up to two retries to connect to the websocket in load tests. With [this change](https://github.com/code-dot-org/javabuilder/pull/272) we should see very few failures, but it is still worth including this logic so we can track requests that would have succeeded on retry. Since we now do these retries, this PR also increases the timeout for a request to 1 minute, as there were occasional successful requests that took up to 40 seconds, and the timeout is intended to stop requests that have failed.

I also set the default sleep to `false` as we get better data with that option, and decreased the default high load time since sleep is now false. In addition I updated the `README` with details about how to customize a load test.

### DIFF
--- a/load-test/scripts/configuration.js
+++ b/load-test/scripts/configuration.js
@@ -23,7 +23,7 @@ export const PRIVATE_KEY = null;
 // reached this time. This is so we do not issue too many requests.
 export const REQUEST_TIME_MS = 20000;
 // Time after which to timeout a request
-export const TIMEOUT_MS = 40000;
+export const TIMEOUT_MS = 60000;
 
 // Mini-app types
 export const MiniAppType = {
@@ -60,8 +60,7 @@ export function getTestOptions(maxUserGoal, highLoadTimeMinutes) {
       exceptions: ["count == 0"],
       errors: ["count == 0"],
       timeouts: ["count == 0"],
-      total_session_time: ["p(95) < 5000"],
-      dropped_iterations: ["count == 0"]
+      total_session_time: ["p(95) < 5000"]
     },
     summaryTrendStats: ["avg", "min", "med", "max", "p(90)", "p(95)", "p(98)", "p(99)"],
   };
--- a/load-test/scripts/loadTest.js
+++ b/load-test/scripts/loadTest.js
@@ -18,14 +18,14 @@ import {
 // Change these options to increase the user goal or time to run the test.
 export const options = getTestOptions(
   /* User goal */ 1000,
-  /* High load time minutes */ 15
+  /* High load time minutes */ 4
 );
 
 // Change this to test different code
 const sourceToTest = helloWorld;
 // Set this to true to space out requests every REQUEST_TIME_MS milliseconds. Set to
 // false to send as many requests as possible.
-const SHOULD_SLEEP = true;
+const SHOULD_SLEEP = false;
 
 const exceptionCounter = new Counter("exceptions");
 const errorCounter = new Counter("errors");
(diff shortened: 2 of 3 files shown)


load-test/scripts/configuration.js
```diff
@@ -60,8 +60,7 @@ export function getTestOptions(maxUserGoal, highLoadTimeMinutes) {
       exceptions: ["count == 0"],
       errors: ["count == 0"],
       timeouts: ["count == 0"],
-      total_session_time: ["p(95) < 5000"],
-      dropped_iterations: ["count == 0"]
+      total_session_time: ["p(95) < 5000"]
```
sanchitmalhotra126, line 63: (for my understanding) why are we removing this?

load-test/scripts/loadTest.js
```diff
@@ -18,14 +18,14 @@ import {
...
-  /* High load time minutes */ 15
+  /* High load time minutes */ 4
 );
 
 // Change this to test different code
 const sourceToTest = helloWorld;
 // Set this to true to space out requests every REQUEST_TIME_MS milliseconds. Set to
 // false to send as many requests as possible.
-const SHOULD_SLEEP = true;
+const SHOULD_SLEEP = false;
```
sanchitmalhotra126, line 28: nit: should `SHOULD_SLEEP` be part of the `options` object just so all the configurable parameters are in one place?

load-test/scripts/loadTest.js
```diff
@@ -53,20 +56,53 @@ export default function () {
...
+      shouldRetry = false;
+    } 
+    tries++;
+  }
+  if (tries === 1) {
+    zeroRetries.add(1);
+  } else if (tries === 2) {
+    oneRetry.add(1);
+  } else if (tries === 3) {
+    twoRetries.add(1);
```
sanchitmalhotra126, line 87: nit: just a thought, but should we make this more generic by storing the counters in an array? e.g. `retryCounters[tries - 1].add(1)`? may be useful if we ever end up needing more than 3 retries - however if you think that's unlikely, then this is probably easier.

sanchitmalhotra126 (approved):
Couple of small things but LGTM otherwise!

load-test/scripts/configuration.js
molly-moen: Oops meant to add a comment explaining this! I had previously changed the test [executor](https://k6.io/docs/using-k6/scenarios/executors/) from per VU iterations to constant VUs, which meant a change from expecting x number of VUs to do y iterations in z minutes, to just having x VUs do as many requests as possible in z minutes. This meant dropped_iterations was no longer relevant, as that was reporting the iterations some VUs didn't get to execute in the time limit. We no longer have an expected number of iterations per VU so that number is always 0.

load-test/scripts/loadTest.js
molly-moen: The [options object](https://k6.io/docs/getting-started/running-k6/#using-options) is used by k6 to determine specific runtime parameters and `SHOULD_SLEEP` is something we are using manually in our execution function (in `loadTest.js` we don't actually reference `options` anywhere, we just have to define it in the script). My thought with putting this here is that test-specific parameters are all configured here (the parameters to options, `sourceToTest` and `SHOULD_SLEEP`). I'll update `sourceToTest` to also be in all caps to imply it's a constant here, looking at this I realize both variables should have the same casing.

load-test/scripts/loadTest.js
molly-moen: good idea, updated!

PR #274, opened 2022-04-15
Limit payload size in exception messages

bencodeorg:
We have observed errors in cloudwatch relating to students generating messages that exceed [API Gateway's payload size limit](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/apigatewaymanagementapi/ApiGatewayManagementApiClient.html#postToConnection-java.util.function.Consumer-).

This doesn't appear to be an issue when generating long system out messages, as the messages are flushed before we reach the payload limit.

We do seem to hit this error when trying to send very long stack traces (eg, in the case of stack overflow errors). This PR limits the size of the stack trace we send to students.

[Slack discussion of CloudWatch errors observed](https://codedotorg.slack.com/archives/C01EF4GJ9GE/p1649695696604249)
[Slack discussion of long system out message succeeding](https://codedotorg.slack.com/archives/C01EF4GJ9GE/p1650041164411609)

Jira: [JAVA-546](https://codedotorg.atlassian.net/browse/JAVA-546)

## Testing story

I was able to reproduce the payload too large error (error seen in cloudwatch, Javalab hung on "Running...") in the case of a recursive program with many iterations that generated a stack overflow message (code taken from a student project that generated one of these errors originally).

After this changes, I was able to see a (truncated, but descriptive) stack overflow error message when running the same program:

image

I wasn't sure if we were hitting the "WebSocket frame size" (32 KB) or "Message payload size" (128 KB) limit described here:
https://docs.aws.amazon.com/apigateway/latest/developerguide/limits.html

I observed a successful message of over 100,000 characters (100+ KB), and a failed one of closer to 200 KB, so I believe it's the 128 KB limit.

### DIFF
--- a/org-code-javabuilder/protocol/src/main/java/org/code/protocol/JavabuilderThrowableMessageUtils.java
+++ b/org-code-javabuilder/protocol/src/main/java/org/code/protocol/JavabuilderThrowableMessageUtils.java
@@ -7,6 +7,9 @@ import java.util.HashMap;
 
 /* Helper for formatting exceptions that will be displayed to the user. */
 public final class JavabuilderThrowableMessageUtils {
+  private static final int MAX_MESSAGE_SIZE_CHARS = 14000;
+  private static final int MAX_CAUSE_MESSAGE_SIZE_CHARS = 1000;
+
   private JavabuilderThrowableMessageUtils() {
     throw new UnsupportedOperationException("Instantiation of utility class is not allowed.");
   }
@@ -31,15 +34,21 @@ public final class JavabuilderThrowableMessageUtils {
 
     String preferredFallbackMessage = null;
     if (cause != null) {
-      // To do: remove CAUSE key once we're using STACK_TRACE and EXCEPTION_TYPE
-      // in (partially) translated exception messages in Javalab.
-      // https://codedotorg.atlassian.net/browse/JAVA-439
       preferredFallbackMessage = getDefaultFallbackMessageString(cause);
       detail.put(ClientMessageDetailKeys.CAUSE, preferredFallbackMessage);
-      detail.put(ClientMessageDetailKeys.STACK_TRACE, getUserFacingStackTraceString(cause));
-      detail.put(ClientMessageDetailKeys.EXCEPTION_MESSAGE, getExceptionMessageString(cause));
-      if (cause.getMessage() != null) {
-        detail.put(ClientMessageDetailKeys.CAUSE_MESSAGE, cause.getMessage());
+
+      // To do: remove CAUSE key and replace with we're using STACK_TRACE and EXCEPTION_MESSSAGE
+      // to enable (partially) translated exception messages in Javalab.
+      // https://codedotorg.atlassian.net/browse/JAVA-439
+      // Commenting out these lines until then to minimize size of message sent to user.
+      // detail.put(ClientMessageDetailKeys.STACK_TRACE, getUserFacingStackTraceString(cause));
+      // detail.put(ClientMessageDetailKeys.EXCEPTION_MESSAGE, getExceptionMessageString(cause));
+
+      String causeMessage = cause.getMessage();
+      if (causeMessage != null) {
+        detail.put(
+            ClientMessageDetailKeys.CAUSE_MESSAGE,
+            trimMessage(causeMessage, MAX_CAUSE_MESSAGE_SIZE_CHARS));
       }
     }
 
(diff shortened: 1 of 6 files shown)


org-code-javabuilder/protocol/src/main/java/org/code/protocol/JavabuilderThrowableMessageUtils.java
```diff
@@ -31,15 +41,17 @@ public static HashMap<String, String> getExceptionDetails(
...
     String preferredFallbackMessage = null;
     if (cause != null) {
-      // To do: remove CAUSE key once we're using STACK_TRACE and EXCEPTION_TYPE
-      // in (partially) translated exception messages in Javalab.
+      // To do: remove CAUSE key and replace with stack trace and exception message individually
+      // to enable (partially) translated exception messages in Javalab.
       // https://codedotorg.atlassian.net/browse/JAVA-439
       preferredFallbackMessage = getDefaultFallbackMessageString(cause);
       detail.put(ClientMessageDetailKeys.CAUSE, preferredFallbackMessage);
-      detail.put(ClientMessageDetailKeys.STACK_TRACE, getUserFacingStackTraceString(cause));
```
bencodeorg, line 39: `STACK_TRACE` and `EXCEPTION_MESSAGE` were added to allow us to translate part of the exception message to students, and are not used by Java Lab. Since these make the payload much larger, removing them for now (until we actually want to do that translation work).

https://github.com/code-dot-org/javabuilder/pull/205

org-code-javabuilder/protocol/src/main/java/org/code/protocol/JavabuilderThrowableMessageUtils.java
```diff
@@ -7,6 +7,16 @@
 
 /* Helper for formatting exceptions that will be displayed to the user. */
 public final class JavabuilderThrowableMessageUtils {
+  // The detail payload can get extremely large (eg, in case of sending stack trace from stack
+  // overflow error).
+  // We include two potentially very long detail keys (CAUSE and FALLBACK_MESSAGE) --
+  // limit these to 50KB each, which should keep us well under 128KB payload size limit:
+  // https://docs.aws.amazon.com/apigateway/latest/developerguide/limits.html
+  private static final int MAX_DETAIL_SIZE_CHARS = 50000;
+  // CAUSE_MESSAGE is used in Javalab to access to get relatively short strings
```
jmkulwik, line 16: NIT: "to access to get" -> "to get"?

jmkulwik (approved):
LGTM!

molly-moen (approved):

org-code-javabuilder/protocol/src/main/java/org/code/protocol/JavabuilderThrowableMessageUtils.java
```diff
@@ -68,7 +80,7 @@ public static String getLoggingString(Throwable throwable) {
   private static String getDefaultFallbackMessageString(Throwable cause) {
     String loggingString = "Exception message: " + getExceptionMessageString(cause) + "\n";
     loggingString += getUserFacingStackTraceString(cause);
-    return loggingString;
+    return trimDetail(loggingString, MAX_DETAIL_SIZE_CHARS);
```
sanchitmalhotra126, line 83: nit: I think we typically have a preference to use `this`, i.e. `this.trimDetail()`?

org-code-javabuilder/protocol/src/main/java/org/code/protocol/JavabuilderThrowableMessageUtils.java
```diff
@@ -93,4 +105,12 @@ public static String getUserFacingStackTraceString(Throwable cause) {
   private static String getExceptionMessageString(Throwable cause) {
     return cause.toString();
   }
+
+  private static String trimDetail(String detailMessage, int charLimit) {
+    String trimmedMessage = detailMessage;
+    if (detailMessage.length() > charLimit) {
+      trimmedMessage = trimmedMessage.substring(0, charLimit);
+    }
+    return trimmedMessage;
```
sanchitmalhotra126, line 114: nit: could potentially simplify this to `return detailMessage.substring(0, Math.min(charLimit, detailMessage.length()))`

sanchitmalhotra126 (approved):
Nice!

org-code-javabuilder/protocol/src/main/java/org/code/protocol/JavabuilderThrowableMessageUtils.java
bencodeorg: It's a static method so there's no `this`, right?

sanchitmalhotra126: Oh whoops my bad, missed that it was static :)

PR #275, opened 2022-04-19
Scanner load test

bencodeorg:
Adds option to include custom metrics used for scanner load test when running a load test.

### DIFF
--- a/load-test/scripts/loadTest.js
+++ b/load-test/scripts/loadTest.js
@@ -14,9 +14,13 @@ import {
   TIMEOUT_MS,
   REQUEST_TIME_MS
 } from "./configuration.js";
-// Custom metrics for scanner load test.
+
+// ************** SETTINGS **************************************************************************
+
+// Custom metrics for load test (currently only implemented for scanner).
 // Expects an object with onMessage and onClose event handler properties.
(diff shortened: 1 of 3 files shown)


load-test/scripts/loadTest.js
```diff
@@ -107,6 +109,7 @@ function onSocketConnect(socket, requestStartTime, websocketStartTime, sessionId
       console.log(`EXCEPTION for session id ${sessionId} ` + parsedData.value);
       exceptionCounter.add(1);
     }
+    metricsReporter.onMessage(socket, parsedData);
```
molly-moen, line 112: I think it would be worth merging this so we have the logic if we want to do this test again. Simplest option might be to add a flag to turn on/off the metrics reporter?

bencodeorg: Ok cool, I can do that 👍

load-test/scripts/loadTest.js
```diff
@@ -15,19 +15,30 @@ import {
   REQUEST_TIME_MS
 } from "./configuration.js";
 
+// ************** SETTINGS **************************************************************************
```
molly-moen, line 18: nice!

molly-moen (approved):
🎉

PR #276, opened 2022-04-20
Prevent token reuse

sanchitmalhotra126:
Adds logic to prevent a token from being reused if it already exists in the DynamoDB table, or if it's already been marked as used by the WebSocketAuthorizer lambda. The solution implemented here is to pass along the request from the authorizer but with an error code that the next lambda uses to reject the response (as discussed in our recent infra meeting).

Note that we'll need an accompanying change in Javalab to properly handle and translate this message. For now, if this happens, the request will just fail and we'll display our standard error message of "We hit an error connecting to our server. Try again."

Testing: tested using dev lambda by manually using the same token twice, both during the upload sources step, and during the websocket connection step.


api-gateway-routes/auth_error_response_helper.rb
```diff
@@ -0,0 +1,14 @@
...
+  def get_response(authorizer_payload)
+    return nil unless authorizer_payload[AUTHORIZATION_ERROR_KEY] && authorizer_payload[AUTHORIZATION_ERROR_CODE_KEY]
+
+    body = {
+      type: AUTHORIZATION_ERROR_KEY,
+      value: authorizer_payload[AUTHORIZATION_ERROR_KEY]
+    }
+    { statusCode: authorizer_payload[AUTHORIZATION_ERROR_CODE_KEY], body: body.to_s }
+  end
+end
```
molly-moen, line 14: nit: add blank line

api-gateway-routes/auth_error_response_helper.rb
```diff
@@ -0,0 +1,14 @@
+module AuthErrorResponseHelper
+  AUTHORIZATION_ERROR_KEY = "authorization_error".freeze
```
molly-moen, line 2: I added [logic](https://github.com/code-dot-org/code-dot-org/pull/45202/files) a while back to parse authorizer messages when they eventually come in. Can we align with that format so we don't need to do excess work or is there a reason we can't?

sanchitmalhotra126: Oh good call! Totally forgot we had this already. I'll update this to match the existing format.

api-gateway-routes/auth_error_response_helper.rb
sanchitmalhotra126: Done!

api-gateway-routes/auth_error_response_helper.rb
sanchitmalhotra126: Updated this and our other token status to match Javalab!

bencodeorg:
Were you able to see whether this is happening at all currently (via cloudwatch logs)? Not sure how searchable the logging message is currently.

sanchitmalhotra126:
> Were you able to see whether this is happening at all currently (via cloudwatch logs)? Not sure how searchable the logging message is currently.

Yep! They were showing up in both authorizer lambdas with messages like: `TOKEN VALIDATION ERROR: used token_id: 84ec7ca3-7aba-4892-8871-ccdd320ea221`

bencodeorg:
> > Were you able to see whether this is happening at all currently (via cloudwatch logs)? Not sure how searchable the logging message is currently.
> 
> Yep! They were showing up in both authorizer lambdas with messages like: `TOKEN VALIDATION ERROR: used token_id: 84ec7ca3-7aba-4892-8871-ccdd320ea221`

You mean we are seeing this happen in production, or just you saw them when you were testing? If in production, how frequently?

sanchitmalhotra126:
> > > Were you able to see whether this is happening at all currently (via cloudwatch logs)? Not sure how searchable the logging message is currently.
> > 
> > 
> > Yep! They were showing up in both authorizer lambdas with messages like: `TOKEN VALIDATION ERROR: used token_id: 84ec7ca3-7aba-4892-8871-ccdd320ea221`
> 
> You mean we are seeing this happen in production, or just you saw them when you were testing? If in production, how frequently?

Ah sorry - I meant only on my dev instance while testing. In production, only see "not_vetted" appearing in the WebsocketAuthorizerLambda logs and "user_blocked" appearing in the HttpAuthorizerLambda

molly-moen (approved):

bencodeorg (approved):

bencodeorg:
> Ah sorry - I meant only on my dev instance while testing. In production, only see "not_vetted" appearing in the WebsocketAuthorizerLambda logs and "user_blocked" appearing in the HttpAuthorizerLambda

Cool -- tangential to this PR but curious about not vetted, did you see that a lot? We wouldn't expect that, right?

sanchitmalhotra126:
> Cool -- tangential to this PR but curious about not vetted, did you see that a lot? We wouldn't expect that, right?

Yep - from just briefly scanning the logs, my guess is that any tokens that are returning "user_blocked" in the http authorizer are also returning "not_vetted" in the websocket authorizer, since I believe don't mark the token as vetted if the user is blocked, but we're still passing on the request.

PR #279, opened 2022-04-22
Add theater and scanner load test sources

bencodeorg:
Add sources used for scanner and theater load tests.

The scanner load test had some custom metrics that were a bit more involved to set up -- [that additional logic is here ](https://github.com/code-dot-org/javabuilder/pull/275) if anyone wants to see it, but I think I'll probably just close that PR without merging it.


molly-moen (approved):
🎉

PR #280, opened 2022-04-22
Hide deprecation warnings

bencodeorg:
Hides deprecation warnings from students. Relevant for situations where they are taught methods that are deprecated (eg, integer constructor).

[JAVA-473](https://codedotorg.atlassian.net/browse/JAVA-473)

**Before (top set of messages) and after (bottom set) this change**

image

I looked into [using suppressing deprecation warnings via tag instead](https://docs.oracle.com/en/java/javase/11/core/enhanced-deprecation1.html#GUID-F0CD4B54-CD8A-4153-87F9-7E7481F9AAFC) (which works if you add it directly to the top of the class with the main method in the student code being compiled), but doesn't work in our reflective world.

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserCodeCompiler.java
+++ b/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserCodeCompiler.java
@@ -21,6 +21,12 @@ public class UserCodeCompiler {
   private final File tempFolder;
   private final OutputAdapter outputAdapter;
 
+  private static final String SYSTEM_PACKAGE_OVERRIDE_NAME = "org.code.lang.System";
+  private static final String DIAGNOSTIC_CODE_SINGLE_IMPORT_ERROR =
+      "compiler.err.already.defined.single.import";
+  private static final String DIAGNOSTIC_CODE_DEPRECATED_WARNING_PREFIX =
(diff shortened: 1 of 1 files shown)


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserCodeCompiler.java
```diff
@@ -44,6 +50,10 @@ public void compileProgram() throws InternalServerError, UserInitiatedException
 
     // diagnostics will include any compiler errors
     for (Diagnostic<? extends JavaFileObject> diagnostic : diagnostics.getDiagnostics()) {
+      if (diagnostic.getCode().startsWith(DIAGNOSTIC_CODE_DEPRECATED_WARNING_PREFIX)) {
```
molly-moen, line 53: nit: can you add a comment as to why we skip these?

molly-moen (approved):
Looks good, one request for an explanatory comment :)

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UserCodeCompiler.java
bencodeorg: Can do!

PR #281, opened 2022-04-25
Setup error messaging for websocket authorization errors

sanchitmalhotra126:
This adds some special case logic to the Start Session lambda to send authorizer/throttling messages to the client in a way that can be interpreted and displayed. 

The issue here is that simply sending the error message and a 403 code in the `on_connect` function will not work; when this happens, the server is indicating to the client that will not establish the websocket connection - unfortunately, the browser (client) does not actually interpret the response body, but simply closes the connection with an error. In order to actually send the error message back to the client so it can be displayed to the user, we instead need to send it as a websocket message, which can only be done after the connection has been successfully established. This follows a pattern similar to our `"connectivityTest"` case where if the auth token or incoming message is "connectivityTest", we simply return a success code and a message. In this case, if there is an authorization error, we establish the connection successfully, but do not invoke the Build and Run lambda. Instead, we wait for the client to send us a message (separate PR), and then can send back the authorization error. It definitely feels like a bit of a hacky approach because we rely on the client to send us a message in order to do anything, but the scope of this is fairly limited; we should only hit this if somehow the client made a successful HTTP request to put sources but then their token was invalidated before making the websocket connection.

Tested on dev lambda by manually reusing the same token.

### DIFF
--- a/api-gateway-routes/api_gateway_proxy_function.rb
+++ b/api-gateway-routes/api_gateway_proxy_function.rb
@@ -27,7 +27,15 @@ def on_connect(event, context)
   authorizer = request_context['authorizer']
 
   authorization_error_response = AuthErrorResponseHelper.get_response(authorizer)
-  return authorization_error_response unless authorization_error_response == nil
+  # If there is an authorization error, allow the websocket connection to open,
+  # but early return so we don't actually invoke the Build and Run lambda and try
+  # to create an SQS queue. We will send the auth error message via the websocket
+  # once the connection has been successfully made (in on_default).
+  if authorization_error_response
+    # Add some logging to make sure we know this is happening.
+    puts "AUTHORIZATION ERROR: #{authorization_error_response[:body]}"
+    return { statusCode: 200, body: authorization_error_response[:body] }
+  end
 
   # Add in some logging to make debugging easier
   puts "CONNECT REQUEST CONTEXT: #{request_context}"
@@ -82,6 +90,12 @@ def on_connect(event, context)
 end
 
 def on_disconnect(event, context)
+  authorizer = event["requestContext"]["authorizer"]
+  authorization_error_response = AuthErrorResponseHelper.get_response(authorizer)
+  # The return value is not actually interpreted by the client, but early return here so that
+  # we don't try to delete a queue that doesn't exist.
+  return authorization_error_response if authorization_error_response
+
   sqs = Aws::SQS::Client.new(region: get_region(context), retry_mode: "adaptive")
   request_context = event['requestContext']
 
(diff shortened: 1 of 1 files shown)


api-gateway-routes/api_gateway_proxy_function.rb
```diff
@@ -27,7 +27,15 @@ def on_connect(event, context)
   authorizer = request_context['authorizer']
 
   authorization_error_response = AuthErrorResponseHelper.get_response(authorizer)
-  return authorization_error_response unless authorization_error_response == nil
+  # If there is an authorization error, allow the websocket connection to open,
+  # but early return so we don't actually invoke the Build and Run lambda and try
+  # to create an SQS queue. We will send the auth error message via the websocket
+  # once the connection has been successfully made (in on_default).
```
molly-moen, line 33: clarifying q: is the reason the browser doesn't interpret the 403 because we haven't implemented that on the java lab side or because it's not possible to interpret it from the browser side?

api-gateway-routes/api_gateway_proxy_function.rb
```diff
@@ -27,7 +27,15 @@ def on_connect(event, context)
...
+  if authorization_error_response
```
molly-moen, line 34: can you add an additional comment that this is a pretty rare scenario because the auth error should be caught by the http authorizer?

api-gateway-routes/api_gateway_proxy_function.rb
sanchitmalhotra126: So from what I understand, it's because the browser just doesn't interpret it/handle it. A 403 here would result in the handshake portion of the websocket connection failing, which on the browser just appears as a websocket error.

jmkulwik: Does the browser differentiate between different error codes? i.e. I could see us returning a 403 for this case, but a different error code for generic errors.

sanchitmalhotra126: Hmm, not that I can tell for this specific part of the WebSocket connection. I believe if it's anything other than 200 here, the socket will just close with an error. For reference: https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_servers#the_websocket_handshake

That being said, we can set custom error codes for auth errors in our HTTP API. For now, I just went off the throttling doc and always set it to 403 ([here](https://github.com/code-dot-org/javabuilder/blob/c32ddb81a07ccae586cb1b62cb6d8b5bc6dd4d38/javabuilder-authorizer/jwt_helper.rb#L47)), but if it makes sense to have different error codes for different auth situations we can definitely support that.

molly-moen (approved):

PR #283, opened 2022-04-26
Add support and util packages in neighborhood and media 

molly-moen:
We have a new pattern for student-facing packages where the student-facing code goes in the main package and all other code goes into a `support` package. We have also adopted a pattern of putting utility classes in a `util` package and making those utility classes final and non-instantiable. This PR applies this pattern to the neighborhood and media packages. There was only one util in those packages, `AudioUtils` in `media`.

Jira Task: [JAVA-562](https://codedotorg.atlassian.net/browse/JAVA-562)


sanchitmalhotra126 (approved):
Nice!

org-code-javabuilder/media/src/main/java/org/code/media/SoundException.java
```diff
@@ -1,13 +1,14 @@
 package org.code.media;
 
+import org.code.media.support.SoundExceptionKeys;
 import org.code.protocol.JavabuilderRuntimeException;
 
 public class SoundException extends JavabuilderRuntimeException {
-  protected SoundException(SoundExceptionKeys key) {
+  public SoundException(SoundExceptionKeys key) {
```
bencodeorg, line 7: This doesn't go in support?

molly-moen: `SoundException` is thrown by `SoundLoader` which is currently in the documentation for students.

PR #285, opened 2022-04-26
Add rubocop to authorizer

bencodeorg:
Adds rubocop to our `javabuilder-authorizer`. Does not run as a pre-commit hook yet, wanted to see if y'all think this is a good idea or not first.

I copied over the explicit style overrides code-dot-org has in its `.rubocop.yml` file, as well as a couple of the rules that we are currently violating from a much longer list of rules that are currently disabled in code-dot-org via its `.rubocop_todo.yml` and `rubocop_hound.yml` files.

There weren't a ton of violations in our existing code, but those are also corrected here.

# Testing story

`bundle exec rubocop` passed successfully with the rules enabled/disabled in this PR. Ran a console program against deployed javabuilder with these changes successfully.

## Follow-up work

- Add commit hook to run rubocop
- Add this same config to `api-gateway-routes`, although there is less logic there


javabuilder-authorizer/jwt_helper.rb
```diff
@@ -91,23 +92,23 @@ def get_public_key(standardized_origin)
...
+      stage_name = 'test'
+    elsif standardized_origin == 'studio.code.org'
+      stage_name = 'production'
+    elsif standardized_origin.start_with?('adhoc-') && standardized_origin.end_with?('-studio.cdn-code.org')
+      stage_name = 'adhoc'
     end
     # End of temporary code
 
-    public_key = ENV["rsa_pub_#{stage_name}"]
+    public_key = ENV.fetch("rsa_pub_#{stage_name}", nil)
```
sanchitmalhotra126, line 111: Just curious, what rule is forcing us to use `fetch` here?

sanchitmalhotra126 (approved):
LGTM! Adding a commit hook sounds like a good idea to me

molly-moen (approved):
LGTM! One question about reuse but it may just not be possible.

javabuilder-authorizer/.rubocop.yml
```diff
@@ -0,0 +1,144 @@
+AllCops:
```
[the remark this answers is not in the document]
bencodeorg: I was just working in `javabuilder-authorizer` to start, and was thinking I'd move stuff around to share config between folders if possible later.

It looks like this is pretty easy though, I think I'll just update it now 👍 . I think `api-gateway-routes` is the only other ruby folder, yeah?

javabuilder-authorizer/jwt_helper.rb
bencodeorg: It's this one:

https://docs.rubocop.org/rubocop/cops_style.html#stylefetchenvvar

It does bring up a broader question though, this rule is currently "pending" -- there's a setting (`NewCops`) of whether you want to automatically apply new cops included in each release. Might be better to disable this setting, which would only apply new cops on a major release:

https://docs.rubocop.org/rubocop/versioning.html#pending-cops

javabuilder-authorizer/.rubocop.yml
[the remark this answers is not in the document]
molly-moen: yep!

javabuilder-authorizer/jwt_helper.rb
molly-moen: hmm interesting. This rule does seem good, but it would be annoying to get new rules all the time. Maybe we disable the setting?

bencodeorg: Yeah agree, I think I might disable it and then reset of the autogenerated changes (including the env ones here) and reapply rubocop's autoformatting

PR #286, opened 2022-04-26
Clear javabuilderfiles directory on localhost at the end of every session

sanchitmalhotra126:
Ensures that the local `"javabuilderfiles"` directory that's created on localhost is cleared after every session. This directory is analogous to S3 in production javabuilder; it stores sources.json, and any theater input/output. The local fileserver reads and writes from this location for GET and PUT requests respectively. I also renamed this util class and associated methods with "Local Storage" rather than "Temp Directory" to avoid confusion with the Java "tmp" directory that we use for loading class files.

JIRA: https://codedotorg.atlassian.net/browse/JAVA-469

Tested locally with theater and prompter.


molly-moen (approved):
🎉

PR #287, opened 2022-04-29
Remove FileNotFoundExceptions from Image and SoundLoader

sanchitmalhotra126:
Removes FileNotFoundExceptions from Image and SoundLoader so students don't have to always explicitly write try-catch statements to handle them. Instead, these are thrown as MediaRuntime/SoundExceptions with the key FILE_NOT_FOUND.

JIRA: https://codedotorg.atlassian.net/browse/JAVA-566

Tested locally.

### DIFF
--- a/org-code-javabuilder/media/src/main/java/org/code/media/support/MediaRuntimeExceptionKeys.java
+++ b/org-code-javabuilder/media/src/main/java/org/code/media/support/MediaRuntimeExceptionKeys.java
@@ -1,5 +1,6 @@
 package org.code.media.support;
 
 public enum MediaRuntimeExceptionKeys {
-  IMAGE_LOAD_ERROR
+  IMAGE_LOAD_ERROR,
+  FILE_NOT_FOUND
 }
--- a/org-code-javabuilder/media/src/main/java/org/code/media/Image.java
+++ b/org-code-javabuilder/media/src/main/java/org/code/media/Image.java
@@ -24,10 +24,13 @@ public class Image {
    * manager.
    *
    * @param filename the name of the image loaded into the asset manager for the project
-   * @throws FileNotFoundException if the file doesn't exist in the asset manager.
    */
-  public Image(String filename) throws FileNotFoundException {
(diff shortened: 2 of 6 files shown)


org-code-javabuilder/media/src/main/java/org/code/media/support/MediaRuntimeExceptionKeys.java
```diff
@@ -1,5 +1,6 @@
 package org.code.media.support;
 
 public enum MediaRuntimeExceptionKeys {
-  IMAGE_LOAD_ERROR
+  IMAGE_LOAD_ERROR,
+  FILE_NOT_FOUND
```
bencodeorg, line 5: Do we need to translate these in javalab, or do we already handle this?

bencodeorg (approved):
Nice 👍

molly-moen (approved):
🎉

org-code-javabuilder/media/src/main/java/org/code/media/support/MediaRuntimeExceptionKeys.java
sanchitmalhotra126: Nope thankfully! The actual value (`"FILE_NOT_FOUND"`) is already mapped.

bencodeorg: I guess generally I'm curious about how we manage when message keys are used in multiple places in Javabuilder -- is it just on us to make sure that if we use the same key in multiple places that the message that we'd want to display on the client is the same (ie, "many to one")?

sanchitmalhotra126: Yeah that's a good question - I think this is one of the only (if not the only) example where a key is defined in multiple enums. In this case it was mostly a matter of convenience since the translation already existed, but yeah I do wonder if we need some more clarity around how we define these.

PR #290, opened 2022-05-04
Replace jcenter with mavenCentral

cat5inthecradle:
jcenter has [shut down](https://blog.gradle.org/jcenter-shutdown), replacing with mavenCentral

Maven Central is [similarly supported](https://docs.gradle.org/current/userguide/declaring_repositories.html#sub:maven_central) by Gradle, so this swap should be sufficient.

It's possible that some packages may not be available on Maven Centeral. So far I have not encountered any errors.

### DIFF
--- a/cicd/3-app/javabuilder/pr-buildspec.yml
+++ b/cicd/3-app/javabuilder/pr-buildspec.yml
@@ -1,0 +1,31 @@
+version: 0.2
+phases:
+  install:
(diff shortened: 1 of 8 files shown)


jmkulwik (approved):
Thanks for the quick fix!

cat5inthecradle:
builds passed! merging

PR #292, opened 2022-05-05
Add warning when user is near hourly throttling limit

molly-moen:
Add logic to warn the user when they are within 10 runs of the hourly throttling limit. Currently the logic is all there but turned off until we fully turn on throttling (will be done as a follow-up). In order to warn the user we do the following
- In the http authorizer after we check for hourly usage above the limit, we also check if the user was within 10 of the limit. If they were, we set a warning on the token in the `token_status_table` DynamoDB table.
- In the websocket authorizer, when we query the `token_status_table` we check for a warning. If there is one, we send it over the newly created websocket so the user can see it. Right now we won't actually send the message. In order to turn it on we need to change a return value in `get_token_info` (there is a todo to do that)

## Follow up work
Next up we will turn on hourly throttling. We will need to do a bit of refactoring so that we don't return early if a user has reached their daily limit but not their hourly limit, as we won't be throttling on daily limits yet.

### DIFF
--- a/javabuilder-authorizer/token_validator.rb
+++ b/javabuilder-authorizer/token_validator.rb
@@ -26,7 +26,7 @@ class TokenValidator
     return error(TOKEN_USED) unless log_token
     return error(USER_BLOCKED) if user_blocked?
     return error(TEACHERS_BLOCKED) if teachers_blocked?
-    hourly_usage_response = user_hourly_usage
+    hourly_usage_response = user_usage(ONE_HOUR_SECONDS)
     return error(USER_OVER_HOURLY_LIMIT) if user_over_hourly_limit?(hourly_usage_response)
     return error(USER_OVER_DAILY_LIMIT) if user_over_daily_limit?
     return error(TEACHERS_OVER_HOURLY_LIMIT) if teachers_over_hourly_limit?
@@ -84,10 +84,6 @@ class TokenValidator
     end
 
     blocked
-  end
-
-  def user_hourly_usage
-    user_usage(ONE_HOUR_SECONDS)
   end
 
   def user_over_hourly_limit?(hourly_usage_response)
--- a/javabuilder-authorizer/token_validator.rb
+++ b/javabuilder-authorizer/token_validator.rb
@@ -11,6 +11,7 @@ class TokenValidator
   # TO DO: move this into env variable
   # https://codedotorg.atlassian.net/browse/JAVA-536
   TEACHER_HOURLY_LIMIT = 1000
+  NEAR_LIMIT_BUFFER = 10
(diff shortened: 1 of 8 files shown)


javabuilder-authorizer/token_validator.rb
```diff
@@ -25,11 +26,15 @@ def validate
...
-    return error(USER_OVER_HOURLY_LIMIT) if user_over_hourly_limit?
+    hourly_usage_response = user_usage(ONE_HOUR_SECONDS)
+    return error(USER_OVER_HOURLY_LIMIT) if user_over_hourly_limit?(hourly_usage_response)
     return error(USER_OVER_DAILY_LIMIT) if user_over_daily_limit?
     return error(TEACHERS_OVER_HOURLY_LIMIT) if teachers_over_hourly_limit?
+    near_limit_detail = user_near_hourly_limit?(hourly_usage_response.count)
+
     log_requests
     mark_token_as_vetted
+    set_token_warning(NEAR_LIMIT, near_limit_detail) if near_limit_detail
```
sanchitmalhotra126, line 37: I know we're not planning on warning on daily limit yet, but just curious, will we only show one warning if the user is near both their hourly and daily limit?

javabuilder-authorizer/token_validator.rb
```diff
@@ -202,8 +221,19 @@ def user_over_limit?(time_range_seconds, limit, logging_message)
     if response.last_evaluated_key
       puts "user_requests query has paginated responses. user_id #{@user_id}"
     end
+    response
+  end
+
+  def user_near_limit?(count, limit)
```
sanchitmalhotra126, line 227: nit: this reads a little weird to me because it returns a hash in the true case (rather than a boolean). Would it make more sense for this to be something like `get_near_limit_detail` and return `nil` if it's not near the limit?

sanchitmalhotra126 (approved):
Looks good!

javabuilder-authorizer/token_validator.rb
bencodeorg: +1

javabuilder-authorizer/token_validator.rb
molly-moen: yeah right now we are set up to only have one warning (we would need multiple warning fields if we wanted to show multiple warnings). In theory it's pretty unlikely someone would reach both their hourly and daily at the same time, and the messaging would be similar (wait to run your code again).

javabuilder-authorizer/token_validator.rb
molly-moen: done!

PR #293, opened 2022-05-05
Turn on hourly throttling

molly-moen:
Turn on hourly throttling for users and classrooms. This change does the following
- actually trigger errors and warnings for hourly usage
- stop querying and logging for daily usage limits. We can't log these anymore because they would end up blocking users, as the daily blocked users go to the same table as the hourly blocked users.
- Fix up the error keys sent back to align with those expected by Java Lab (we had `TEACHERS_BLOCKED` instead of `CLASSROOM_BLOCKED` and sent an invalid key the first time a user was blocked).
- move the teacher hourly limit to an environment variable.
NOTE: Clear the blocked users table before deploying this change

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -31,6 +31,8 @@ SILENCE_ALERTS=${SILENCE_ALERTS-'false'}
 # Default per-user limits to prevent javabuilder abuse.
 LIMIT_PER_HOUR=${LIMIT_PER_HOUR-'50'}
 LIMIT_PER_DAY=${LIMIT_PER_DAY-'150'}
+# Default per-classroom hourly limit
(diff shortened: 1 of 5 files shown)


bencodeorg:
Were you able to do some manual testing here?

jmkulwik (approved):
LGTM!

bencodeorg:
One thing to remember is to clear the blocked_users table manually before deploying this change.

molly-moen:
> Were you able to do some manual testing here?

@bencodeorg yes, I lowered the thresholds and checked that both users and teachers got blocked

molly-moen:
I've cleared the blocked users table and will check again before deploying this change

PR #294, opened 2022-05-06
Add and run basic integration tests

sanchitmalhotra126:
Sets up and adds a basic integration test. This creates a new folder, `integration-tests/` which contains an npm project, test code, and support code. Quick rundown of the files/folders -

- `test/` contains the tests and support code. The tests are split into different files per mini-app, but currently only `helloWorld.js` has any tests in it
  - `test/helpers` - helper functions for common test scenarios
  - `test/lib` - library/support code for running tests (ex. code for connecting to an instance of Javabuilder)
- `run-integration-test.sh` is a script for running the tests standalone locally (because tests will typically run as part of the CI/CD process, which is in a different script with commands very similar to this one)
- `standalone.config` is a config file for adjusting values used by run-integration-test.sh.

I have a few follow-up tasks in mind after this, but I particularly wanted to call out that some of this code looks pretty similar to code in `load-test/` (ex. the sources.js file, making a connection to Javabuilder in JavabuilderConnectionHelper.js). It would be ideal to share some of the common code, but one complication with this is that the integration-test project is setup as an npm project while the load-test project can't use npm. So we'll have to find a way to share the code in a way that doesn't depend on npm packages.

Example test output after running ./run-integration-tests.sh:

Screen Shot 2022-05-06 at 12 47 11 PM

### DIFF
--- a/integration-tests/test/lib/JavabuilderConnectionHelper.js
+++ b/integration-tests/test/lib/JavabuilderConnectionHelper.js
@@ -1,0 +1,99 @@
+import { v4 as uuidv4 } from "uuid";
+import fetch from "node-fetch";
+import jwt from "jsonwebtoken";
+import crypto from "crypto";
+import WebSocket from "ws";
+import {
+  JAVABUILDER_HTTP_URL,
+  JAVABUILDER_WEBSOCKET_URL,
+  PASSWORD,
+  PRIVATE_KEY,
+} from "./environment.js";
+
+// TODO: use integration test specific origin once the key pair has been setup.
+const INTEGRATION_TESTS_ORIGIN = "localhost-studio.code.org";
+const INTEGRATION_TESTS_SESSION_ID_PREFIX = "integrationTests-";
+
+/**
+ * Helper class for facilitating a connection to a Javabuilder instance
+ */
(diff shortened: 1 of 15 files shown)


integration-tests/test/lib/JavabuilderConnectionHelper.js
```diff
@@ -0,0 +1,99 @@
...
+import crypto from "crypto";
+import WebSocket from "ws";
+import {
+  JAVABUILDER_HTTP_URL,
+  JAVABUILDER_WEBSOCKET_URL,
+  PASSWORD,
+  PRIVATE_KEY,
+} from "./environment.js";
+
+// TODO: use integration test specific origin once the key pair has been setup.
```
molly-moen, line 13: do we have a task for this already?

molly-moen (approved):
LGTM!

integration-tests/test/lib/JavabuilderConnectionHelper.js
sanchitmalhotra126: Yep! https://codedotorg.atlassian.net/browse/JAVA-588

PR #298, opened 2022-05-10
Move public keys to separate files and add integration test key

sanchitmalhotra126:
This moves Javabuilder's public keys into their own files and out of environment variables, as we were running into the 4 KB lambda environment variable limit when trying to add new keys. This is meant to be a short-term solution for adding a new integration test key. A longer term solution will be to store these keys in a separate data store (like AWS Parameter Store) and configure specific instances of Javabuilder with specific keys.

See: https://codedotorg.slack.com/archives/C01EF4GJ9GE/p1652139429311099

Testing: used integration tests to manually verify each origin with each key pair and confirmed that we were able to connect to Javabuilder and run code on every origin.


molly-moen (approved):
LGTM!

PR #302, opened 2022-05-12
Use canAccessDashboardAssets flag

molly-moen:
[This PR](https://github.com/code-dot-org/code-dot-org/pull/46315) adds a new `canAccessDashboardAssets` flag on the code-dot-org side. Use this flag on the Javabuilder side to set the `canAccessDashboardAssets` property instead of inferring it based on the dashboard hostname. This PR also removed the `isIntegrationTest` property as it was only used to check if dashboard assets were accessible. It also adds the `canAccessDashboardAssets` property to the integration test configuration, which will enable theater integration tests. 

Note: this should not be merged until the PR linked above is deployed.

### DIFF
--- a/api-gateway-routes/api_gateway_proxy_function.rb
+++ b/api-gateway-routes/api_gateway_proxy_function.rb
@@ -67,7 +67,8 @@ def on_connect(event, context)
     :miniAppType => authorizer["mini_app_type"],
     :javabuilderSessionId => authorizer['sid'],
     :queueName => queue_name,
-    :executionType => authorizer['execution_type']
+    :executionType => authorizer['execution_type'],
+    :canAccessDashboardAssets => authorizer['can_access_dashboard_assets']
   }
 
   response = nil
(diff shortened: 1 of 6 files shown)


sanchitmalhotra126 (commented):
Nice refactor! Do we need to make any changes on WebSocketServer for local Javabuilder? AFAIK local Javabuilder would always be able to access dashboard assets no matter the origin. I think currently we just don't set the `canAccessDashboardAssets` in the WebSocketServer so it defaults to true, but with the client now explicitly passing the flag, does it feel a little weird to outright ignore it?

That being said, all this is AWS-specific - we only read the canAccessDashboardAssets flag in AWSContentManager, and not in LocalContentManager. So maybe we just need to call that out somewhere?

molly-moen:
@sanchitmalhotra126 no changes needed for local Javabuilder. We parse the authorization token differently on the local side, so I think it's ok that it just ignores this field.

sanchitmalhotra126 (approved):

PR #303, opened 2022-05-13
Add FatalError class

sanchitmalhotra126:
Some pre-work for refactoring our error handling. This adds a new `FatalError` class, which is a sub class of `JavabuilderError` - JavabuilderError is the `Error` analog of JavabuilderException/RuntimeException. These classes are currently unused, but they'll help our error handling code differentiate between truly unrecoverable errors (ex. out of disk space) that require shutting down the runtime, vs severe but recoverable exceptions (InternalServerException/RuntimeExceptions).

Tested locally (to make sure there's no effect on existing behavior)

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/FatalErrorKey.java
+++ b/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/FatalErrorKey.java
@@ -1,0 +1,19 @@
+package org.code.javabuilder;
+
+import static org.code.javabuilder.LambdaErrorCodes.LOW_DISK_SPACE_ERROR_CODE;
+import static org.code.javabuilder.LambdaErrorCodes.TEMP_DIRECTORY_CLEANUP_ERROR_CODE;
+
+public enum FatalErrorKey {
+  LOW_DISK_SPACE(LOW_DISK_SPACE_ERROR_CODE),
+  TEMP_DIRECTORY_CLEANUP_ERROR(TEMP_DIRECTORY_CLEANUP_ERROR_CODE);
+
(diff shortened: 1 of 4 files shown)


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/FatalErrorKey.java
```diff
@@ -0,0 +1,19 @@
+package org.code.javabuilder;
+
+import static org.code.javabuilder.LambdaErrorCodes.LOW_DISK_SPACE_ERROR_CODE;
+import static org.code.javabuilder.LambdaErrorCodes.TEMP_DIRECTORY_CLEANUP_ERROR_CODE;
+
+public enum FatalErrorKey {
+  LOW_DISK_SPACE(LOW_DISK_SPACE_ERROR_CODE),
```
bencodeorg, line 7: For my understanding, what is going on syntactically here?

bencodeorg (approved):

molly-moen (approved):

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/FatalErrorKey.java
sanchitmalhotra126: So because this enum has a constructor (below on line 12), each enum constant needs to be constructed with an int parameter. In this case, I'm just passing in integer constants from LambdaErrorCodes. This will be useful for when we need to call `System.exit(int code)` when we encounter a FatalError because we can just use the error code associated with the enum constant.

PR #305, opened 2022-05-13
Add cloudwatch put metric data permission to Lambdas

molly-moen:
We need this permission to publish custom metrics to cloudwatch from the build and run lambdas. More details in [this slack thread](https://codedotorg.slack.com/archives/C01EF4GJ9GE/p1652460222474799)

### DIFF
--- a/iam.yml
+++ b/iam.yml
@@ -57,6 +57,12 @@ Resources:
                   - "execute-api:ManageConnections"
                 Resource:
                   - !Sub "arn:aws:execute-api:*:${AWS::AccountId}:${JavabuilderApiId}/*"
+              # Build and Run Lambdas need to be able to write metrics to CloudWatch
+              - Effect: Allow
+                Action:
+                  - 'cloudwatch:PutMetricData'
+                Resource:
+                  - '*'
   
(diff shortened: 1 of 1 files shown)


iam.yml
```diff
@@ -57,6 +57,12 @@ Resources:
                   - "execute-api:ManageConnections"
                 Resource:
                   - !Sub "arn:aws:execute-api:*:${AWS::AccountId}:${JavabuilderApiId}/*"
+              # Build and Run Lambdas need to be able to write metrics to CloudWatch
+              - Effect: Allow
+                Action:
+                  - 'cloudwatch:PutMetricData'
+                Resource:
+                  - '*'
```
sureshc, line 65: Not critical, but could we scope this to `javabuilder*` and then use that namespace for any metric that Javabuilder publishes?

molly-moen: If I'm reading [the documentation](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/permissions-reference-cw.html) right, it seems like only `*` is supported here

sureshc (approved):

PR #306, opened 2022-05-17
Neighborhood integration test

bencodeorg:
Adds an integration test for painting a single square on a Neighborhood level. It'd be nice to add an integration test to move the painter as well, but I thought it'd be good to use the exact same sources as what we have in load tests so we could easily dedupe if we'd like.

## Testing story

Successfully executed test against our test instance.

### DIFF
--- a/integration-tests/test/helloWorld.js
+++ b/integration-tests/test/helloWorld.js
@@ -1,6 +1,7 @@
 import { helloWorld } from "./lib/sources.js";
 import { NEIGHBORHOOD } from "./lib/MiniAppType.js";
-import { verifyMessagesRecevied } from "./helpers/testHelpers.js";
+import { verifyMessagesReceived } from "./helpers/testHelpers.js";
+import { expect } from "chai";
 
 describe("Hello World", () => {
   it("Runs Hello World project", (done) => {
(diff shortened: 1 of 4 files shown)


integration-tests/test/helloWorld.js
```diff
@@ -12,7 +13,8 @@ describe("Hello World", () => {
       { type: "SYSTEM_OUT", value: "\n" },
       { type: "STATUS", value: "EXITED" },
     ];
+    const assertOnMessagesReceived = receivedMessages => expect(expectedMessages).to.deep.equal(receivedMessages);
```
molly-moen, line 16: nit: should this be moved into `testHelpers` since it will most likely be used by other integration tests?

molly-moen (approved):
Nice!

integration-tests/test/helloWorld.js
bencodeorg: Yeah that sounds good, I'm kind of moving stuff around as I'm adding more tests, so I'll do that in my next integration test PR.

sanchitmalhotra126 (approved):
Looks good! To your point, I can see some of this code being shared across tests and within Neighborhood tests but agree that we can refactor as we go along.

PR #307, opened 2022-05-17
Add cloudwatch metrics

molly-moen:
Add custom cloudwatch metrics for the metrics we have been deriving via logs. These metrics are:
- Severe Errors (count)
- Cold Boot Time (ms)
- Initialization Time (ms)
- Transition Time (ms)
- Cleanup Time (ms)

The only dimension I put on the metrics is `functionName` so we can split data out by lambda.

Note: [This PR](https://github.com/code-dot-org/javabuilder/pull/305) must be merged and the iam change deployed before this can be merged.

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSMetricClient.java
+++ b/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSMetricClient.java
@@ -5,6 +5,8 @@ import com.amazonaws.services.cloudwatch.AmazonCloudWatchClientBuilder;
 import com.amazonaws.services.cloudwatch.model.*;
 import org.code.protocol.MetricClient;
 
+// Metric Client which published metrics to AWS CloudWatch.
+// Requires cloudwatch:PutMetricData Permission on the Lambda.
 public class AWSMetricClient implements MetricClient {
   private final AmazonCloudWatch cloudWatchClient;
   private final Dimension functionNameDimension;
--- a/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSMetricClient.java
+++ b/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSMetricClient.java
@@ -1,15 +1,63 @@
 package org.code.javabuilder;
 
-import java.time.Instant;
+import com.amazonaws.services.cloudwatch.AmazonCloudWatch;
+import com.amazonaws.services.cloudwatch.AmazonCloudWatchClientBuilder;
+import com.amazonaws.services.cloudwatch.model.*;
 import org.code.protocol.MetricClient;
 
 public class AWSMetricClient implements MetricClient {
-  @Override
-  public void publishSevereError() {}
+  private final AmazonCloudWatch cloudWatchClient;
+  private final Dimension functionNameDimension;
+
+  private static final String NAMESPACE = "Javabuilder";
+
+  public AWSMetricClient(String functionName) {
+    this.cloudWatchClient = AmazonCloudWatchClientBuilder.defaultClient();
+    this.functionNameDimension = new Dimension().withName("functionName").withValue(functionName);
+  }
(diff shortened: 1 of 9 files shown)


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
```diff
@@ -108,6 +108,9 @@ public String handleRequest(Map<String, String> lambdaInput, Context context) {
     logger.setUseParentHandlers(false);
     Properties.setConnectionId(connectionId);
 
+    MetricClient metricClient = new AWSMetricClient(context.getFunctionName());
```
sureshc, line 111: I think the FunctionName will be something long and different for each deployment of Javabuilder? Something like `javabuilder-[subdomain]-BuildAndRunJavaConsoleProjectFun-[randomid]`

If it’s not too much extra work, it might be useful to just extract the logical name of the Function (`BuildAndRunJava`), the Project Sub-Type (`ConsoleProject`), and the Stack Name / Subdomain (`beta`, `integration`, `production`, etc.), and publish each of those as separate Dimensions of the Metric.

That may make it easier for us to track Metrics by a given instance of Javabuilder, or by a given function type.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/AWSMetricClient.java
```diff
@@ -0,0 +1,70 @@
...
+    this.publishMillisecondMetric("CleanupTime", (double) cleanupTime);
+  }
+
+  private void publishMillisecondMetric(String metricName, double milliseconds) {
+    MetricDatum metricDatum =
+        new MetricDatum()
+            .withMetricName(metricName)
+            .withUnit(StandardUnit.Milliseconds)
+            .withValue(milliseconds)
+            .withDimensions(this.functionNameDimension);
```
sureshc, line 53: See note below about possibly specifying the Javabuilder instance (subdomain/stackname), the Function type (BuildAndRunJava, Authorizer, etc.) as separate Dimensions.

org-code-javabuilder/protocol/src/main/java/org/code/protocol/MetricClientManager.java
```diff
@@ -0,0 +1,31 @@
...
+    this.metricClient = metricClient;
+  }
+
+  public static void create(MetricClient metricClient) {
+    MetricClientManager.clientManagerInstance = new MetricClientManager(metricClient);
+  }
+
+  public static MetricClientManager getInstance() {
+    if (MetricClientManager.clientManagerInstance == null) {
+      throw new InternalServerRuntimeException(InternalExceptionKey.INTERNAL_EXCEPTION);
```
sanchitmalhotra126, line 18: nit: should we include an Exception cause here so we know what happened? ex. `new Exception("MetricsClient not initialized")`

sanchitmalhotra126 (approved):
LGTM!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
molly-moen: I would rather not manually parse the function name because that is fragile if we ever decide to change the pattern. I like the idea of splitting it up, but in order to do that we would need to send a few more pieces of information to each lambda (and still probably wouldn't have the stack name unless we did that manually parsing). For what it's worth the default metrics you get on a lambda have the dimension of function name, so we are following an established pattern.

org-code-javabuilder/protocol/src/main/java/org/code/protocol/MetricClientManager.java
molly-moen: done!

sureshc (approved):

PR #308, opened 2022-05-17
Theater integration test

bencodeorg:
Adds an integration test for a basic theater project that displays an image, some text, and makes a sound.

I refactored the generic case of wanting to confirm that each message's `type` and `value` properties match expectations into a helper called `assertMessagesEqual`. The complexity in expectations around what the messages should contain (at least for theater and neighborhood) are in the `detail` object, so I added an optional function parameter `verifyDetailKey` that allows the caller to verify the `detail` object as they'd like.

### DIFF
--- a/integration-tests/test/neighborhood.js
+++ b/integration-tests/test/neighborhood.js
@@ -1,3 +1,66 @@
+import {neighborhood} from "./lib/sources.js";
+import {NEIGHBORHOOD} from "./lib/MiniAppType.js";
+import {verifyMessagesReceived} from "./helpers/testHelpers.js";
+import {expect} from "chai";
+
 describe("Neighborhood", () => {
-  // TODO
+  it("Paints single square", (done) => {
+    const expectedMessages = [
+      {type: "STATUS", value: "COMPILING"},
+      {type: "STATUS", value: "COMPILATION_SUCCESSFUL"},
+      {type: "STATUS", value: "RUNNING"},
+      {
+        type: "NEIGHBORHOOD",
+        value: "INITIALIZE_PAINTER",
+        detail: {
+          x: "0",
+          y: "0",
+          paint: "1",
+          direction: "east"
+        }
+      },
+      {
+        type: "NEIGHBORHOOD",
+        value: "PAINT",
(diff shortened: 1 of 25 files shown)


integration-tests/test/neighborhood.js
```diff
@@ -1,68 +1,54 @@
...
       {
         type: "NEIGHBORHOOD",
         value: "INITIALIZE_PAINTER",
         detail: {
           x: "0",
           y: "0",
           paint: "1",
-          direction: "east"
+          direction: "east",
+          id: "painter-1"
```
bencodeorg, line 23: The `id` fields here are not strictly necessary for this test, but I thought it was clearer to show the expected structure of the detail object even if we don't use it.

integration-tests/test/theater.js
```diff
@@ -1,3 +1,49 @@
...
+
+      assertMessagesEqual(receivedMessages, expectedMessages, verifyDetailKey);
+    }
+
+    verifyMessagesReceived(theaterImageAndText, THEATER, assertOnMessagesReceived, done);
+  }).timeout(20000);
 });
+
+const verifyMediaURL = (url, filename) => {
+  expect(url.includes("content.code.org")).to.be.true;
```
bencodeorg, line 44: Wasn't sure whether to verify the entire URL (do we want integration tests to work on any javabuilder deploy, or just test?) -- can make this more substantial if we'd like.

molly-moen (approved):

integration-tests/test/theater.js
sanchitmalhotra126: Yeah good question - doesn't seem super critical to me that we verify the full URL. If we wanted to, we could cross reference by pulling out the subdomain value from JAVABUILDER_WEBSOCKET_URL in environment.js, but that doesn't seem as important as verifying the filename.

sanchitmalhotra126 (approved):
Nice refactors!

integration-tests/test/theater.js
sanchitmalhotra126: Yep, looks good!

PR #309, opened 2022-05-18
Fix local theater

molly-moen:
We had an issue where the local version of theater no longer worked because we cleared the folder containing theater output before the client had a change to fetch it. No longer clear that folder. We don't need to worry too much about this folder getting too big, because we almost always put files with the same 3 names in there, so those files will just get overwritten. If you test with a bunch of prompter images the folder could get bigger, but that is a pretty rare scenario and can be solved by the developer manually clearing the folder.


sanchitmalhotra126 (approved):
🎉

PR #310, opened 2022-05-18
Console integration test

bencodeorg:
Add an integration test for console projects, where the program asks a user's name. Adds new logic to allow for sending of messages (or doing anything) in response to messages received from Javabuilder in integration tests.

### DIFF
--- a/integration-tests/test/console.js
+++ b/integration-tests/test/console.js
@@ -18,12 +18,11 @@ describe("Console", () => {
       {type: "SYSTEM_OUT", value: "\n"},
       EXIT_STATUS_MESSAGE
     ];
-    const assertOnMessagesReceived = receivedMessages => {console.log(receivedMessages); assertMessagesEqual(receivedMessages, expectedMessages);}
+    const assertOnMessagesReceived = receivedMessages => assertMessagesEqual(receivedMessages, expectedMessages);
 
     let hasReceivedMessage, hasReceivedNewLineAfterMessage;
     const onMessageCallback = (parsedData, socket, allMessages) => {
-      // Confirm receipt of "What's your name" and new line messages back to back
-      // This logic is tightly coupled to the scanner sources.
+      // Confirm receipt of "What's your name" and new line messages back to back.
       if (parsedData.type === "SYSTEM_OUT" && parsedData.value === "What's your name?") {
         hasReceivedMessage = true;
       } else if (hasReceivedMessage && (parsedData.type === "SYSTEM_OUT" && parsedData.value === "\n")) {
--- a/integration-tests/test/console.js
+++ b/integration-tests/test/console.js
@@ -1,3 +1,48 @@
+import { scanner } from "./lib/sources.js";
+import { CONSOLE } from "./lib/MiniAppType.js";
+import {
+  INITIAL_STATUS_MESSAGES,
+  EXIT_STATUS_MESSAGE,
+  assertMessagesEqual,
+  verifyMessages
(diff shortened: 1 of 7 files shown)


integration-tests/test/console.js
```diff
@@ -1,3 +1,47 @@
...
+      {type: "SYSTEM_OUT", value: "Hello Ben!"},
+      {type: "SYSTEM_OUT", value: "\n"},
+      EXIT_STATUS_MESSAGE
+    ];
+    const assertOnMessagesObserved = observedMessages => assertMessagesEqual(observedMessages, expectedMessages);
+
+    let hasReceivedMessage, hasReceivedNewLineAfterMessage;
+    const onMessageCallback = (parsedData, socket, allMessages) => {
+      // Confirm receipt of "What's your name" and new line messages back to back.
+      if (parsedData.type === "SYSTEM_OUT" && parsedData.value === "What's your name?") {
```
molly-moen, line 26: should we have constants for `"What's your name"` and `"Ben"`?

molly-moen (approved):
LGTM!

integration-tests/test/console.js
bencodeorg: This whole implementation is tightly coupled to the sources code provided -- I thought a little about whether we should modify sources.js to include expected messages like this one, but ended up deciding it didn't feel worth doing at the moment?

sanchitmalhotra126: Yeah I think it's fair to make assumptions given the source code we're using - we do that for example in the Hello World test where we expect a specific string ("Hello World")

integration-tests/test/helpers/testHelpers.js
```diff
@@ -2,31 +2,39 @@ import connectionHelper from "../lib/JavabuilderConnectionHelper.js";
...
- * Helper for verifying the basic case that set of messages was received
+ * Helper for verifying that a set of messages was received (and possibly sent)
  * from Javabuilder. Fails if there are any errors connecting.
  *
  * @param {*} sourcesJson sources (see sources.js)
  * @param {*} miniAppType mini-app type (see MiniAppType.js)
- * @param {*} assertOnMessagesReceived a verification callback to be called once all messages have been received
+ * @param {*} assertOnMessagesObserved a verification callback to be called once all messages have been sent/received
  * @param {*} doneCallback Mocha's 'done' callback
+ * @param {*} onMessageCallback a callback executed after each received message that can be used to respond to messages
```
sanchitmalhotra126, line 12: I'm curious, do you think it would just be easier to not use this test helper for interactive programs and instead call functions on `JavabuilderConnectionHelper` directly? We could expose a `sendMessage()` that could be called directly, so you wouldn't have to pass an extra callback in here. Just a thought - this approach also seems to work fine

bencodeorg: Totally missed this message -- your comment is kind of what I ended up having to do to test having a bad token in my [error cases PR](https://github.com/code-dot-org/javabuilder/pull/312/files), and I've been feeling like I'm piping callbacks around too much generally. So this might be just what I need to do, I'll look into it thanks!

bencodeorg: Looking at it a bit, I do feel like I still need a lot of the stuff that `verifyMessages` provides (collect all messages in `onMessage`, verify messages in `onClose`), which I think I'd need to replicate or move to call `JavabuilderConnecionHelper` directly. Maybe I'll just leave as-is?

PR #312, opened 2022-05-20
Add integration tests for error states

bencodeorg:
Adds new integration states to cover:
- compilation error
- use of an unsupported class
- use of an invalid token
- runtime error (trying to use an image that has not been uploaded)

Admittedly a lot of callbacks being passed around as I've been working through this, open to thoughts on whether there's any restructuring that might make this easier to read going forward!

### DIFF
--- a/integration-tests/test/errors.js
+++ b/integration-tests/test/errors.js
@@ -61,8 +61,7 @@ describe("Errors", () => {
     verifyMessages(blockedClassError, CONSOLE, assertOnMessagesReceived, done);
   }).timeout(20000);
 
-  it("Does not have valid token", async () => {
-    // We don't appear to get any more detail than a 500 response when an invalid token is provided.
+  it("Does not accept malformed token", async () => {
     const httpResponse = await uploadSources(blankProject, 'aBadToken');
     expect(httpResponse.status).to.equal(500);
   }).timeout(2000);
--- a/integration-tests/test/errors.js
+++ b/integration-tests/test/errors.js
@@ -8,7 +8,7 @@ import {
   EXIT_STATUS_MESSAGE,
 } from "./helpers/testHelpers.js";
 import {expect} from "chai";
-import connectionHelper from "./lib/JavabuilderConnectionHelper.js";
+import {uploadSources} from "./lib/JavabuilderConnectionHelper.js";
 
 describe("Errors", () => {
   it("Compilation Error", (done) => {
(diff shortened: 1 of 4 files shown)


integration-tests/test/errors.js
```diff
@@ -0,0 +1,82 @@
...
+        }
+      },
+      EXIT_STATUS_MESSAGE
+    ];
+
+    const assertOnMessagesReceived = receivedMessages => assertMessagesEqual(receivedMessages, expectedMessages, validateExceptionDetailKey);
+    verifyMessages(blockedClassError, CONSOLE, assertOnMessagesReceived, done);
+  }).timeout(20000);
+
+  it("Does not have valid token", (done) => {
```
bencodeorg, line 64: Realizing this test case is messed up, working on an update now.

bencodeorg: Ok this has been updated and is now ready for review!

integration-tests/test/lib/JavabuilderConnectionHelper.js
```diff
@@ -64,37 +53,49 @@ class JavabuilderConnectionHelper {
...
-      sid: sessionId,
-      can_access_dashboard_assets: false
-    };
+const getRandomId = () => uuidv4();
 
-    const key = crypto.createPrivateKey({
-      key: PRIVATE_KEY,
-      passphrase: PASSWORD,
-    });
+function generateToken(miniAppType, sessionId) {
```
sanchitmalhotra126, line 60: Why this change? just curious

integration-tests/test/errors.js
```diff
@@ -0,0 +1,80 @@
...
+    ];
+
+    const assertOnMessagesReceived = receivedMessages => assertMessagesEqual(receivedMessages, expectedMessages, validateExceptionDetailKey);
+    verifyMessages(blockedClassError, CONSOLE, assertOnMessagesReceived, done);
+  }).timeout(20000);
+
+  it("Does not have valid token", async () => {
+    // We don't appear to get any more detail than a 500 response when an invalid token is provided.
+    const httpResponse = await uploadSources(blankProject, 'aBadToken');
+    expect(httpResponse.status).to.equal(500);
```
sanchitmalhotra126, line 67: This could be a follow-up, but I believe we should get a 4xx error code if the payload is signed with the wrong key (but still well-formed) if you want to add a test for that? I think the 500 is coming from the authorizer throwing an error rather than the authorizer rejecting the token

sanchitmalhotra126 (approved):
Nice!

integration-tests/test/errors.js
bencodeorg: oh I see, like the key needs to be the right length and we'll get 4xx...I can update

integration-tests/test/lib/JavabuilderConnectionHelper.js
bencodeorg: Ah this is left over from a previous WIP iteration, I'll move back.

integration-tests/test/errors.js
bencodeorg: Are there any best practices for stubbing this sort of stuff? Different encryption types are a bit over my head -- do you know where I can see the configuration we use to generate real javabuilder keys? I see something like this, but it requires a bunch of options I don't know anything about:

https://nodejs.org/api/crypto.html#cryptogeneratekeypairtype-options-callback

Alternatively, are there any best practices for stubbing this sort of stuff?

bencodeorg: I renamed the test to "malformed token", so I might just take the easy way out and merge this version

sanchitmalhotra126: Yeah that's totally fair :) I think it'll be little involved to generate a valid but incorrect key. Maybe we can take a look it at it when we add throttling error tests

PR #316, opened 2022-05-24
Add performance alarms

molly-moen:
Add alarms when we see slow performance on the build and run lambda for
- cold boot time
- initialization time
- transition time
- cleanup time

These alarms are all set up as anomaly detection for two reasons. Mainly because different lambda configurations create different performance benchmarks (namely theater, which has a higher memory size). In addition, anomaly detection gives a more accurate representation of what our expected performance is than I could get by manually looking at the data.

The alarms are very similar but I split them up individually instead of creating them via an array because we may want to tune the alarms individually. The general configuration is:
- the metric value is set to p(95) for each minute
- the threshold is greater than the anomaly band. Each anomaly band has a width--this is the `5` in `ANOMALY_DETECTION_BAND(m1, 5)`. I gave a wider band to spikier metrics and vice versa, but these may need to be tuned.
- we alarm if we hit 15 out of 15 minutes of poor performance. If we don't see a minute of data we ignore that minute, so if we saw 13 minutes of poor performance and 2 of no data we would still alert.
- The alarms will go to #ap-csa-dev.

### DIFF
--- a/template.yml.erb
+++ b/template.yml.erb
@@ -683,7 +683,6 @@ Resources:
       AlarmActions:
         - !If [SilenceAlertsCondition, !Ref AWS::NoValue, !Sub "arn:aws:sns:${AWS::Region}:${AWS::AccountId}:javabuilder-slow-performance"]
       InsufficientDataActions: []
-      Dimensions: []
       EvaluationPeriods: 15
       DatapointsToAlarm: 15
       ComparisonOperator: GreaterThanUpperThreshold
@@ -717,7 +716,6 @@ Resources:
       AlarmActions:
         - !If [SilenceAlertsCondition, !Ref AWS::NoValue, !Sub "arn:aws:sns:${AWS::Region}:${AWS::AccountId}:javabuilder-slow-performance"]
       InsufficientDataActions: []
-      Dimensions: []
       EvaluationPeriods: 15
       DatapointsToAlarm: 15
       ComparisonOperator: GreaterThanUpperThreshold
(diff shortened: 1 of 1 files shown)


sanchitmalhotra126 (approved):
LGTM!

bencodeorg:
> * we alarm if we hit 15 out of 15 minutes of poor performance. If we don't see a minute of data we ignore that minute, so if we saw 13 minutes of poor performance and 2 of no data we would still alert.

LMK if I'm reading this correctly -- if we saw 1 minute of poor performance and 14 of no data, we alert?

molly-moen:
@bencodeorg yes that's correct. Another option would be to treat missing data as good, but then if we saw 14 minutes of poor performance and 1 minute of no data we wouldn't alert, and we often having missing minutes of data. Not sure the best configuration to use here.

bencodeorg:
Woof the missing data section in CloudWatch docs is a doozy, I'm down to put them in and see how it goes 😆 

https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/AlarmThatSendsEmail.html#alarms-and-missing-data

bencodeorg (approved):

template.yml.erb
```diff
@@ -671,6 +671,139 @@ Resources:
...
+  <%=name%>SlowCleanupTimeAlarm:
+    Type: AWS::CloudWatch::Alarm
+    Properties:
+      AlarmName: !Sub "${SubDomainName}_<%=name.downcase%>_slow_cleanup_time"
+      AlarmDescription: p(95) cleanup time in Javabuilder's <%=name%> build and run lambda exceeded
+        the normal cleanup time for 15 minutes. Investigate if there has been a performance regression.
+      ActionsEnabled: true
+      OKActions: []
+      AlarmActions:
+        - !If [SilenceAlertsCondition, !Ref AWS::NoValue, !Sub "arn:aws:sns:${AWS::Region}:${AWS::AccountId}:javabuilder-slow-performance"]
```
cat5inthecradle, line 684: Is this topic defined in code somewhere?

cat5inthecradle (approved):

template.yml.erb
molly-moen: It is not (nor are any of our other topics). How should these be defined?

PR #317, opened 2022-05-24
Add isFacing<Direction> methods to painter

sanchitmalhotra126:
Ask from curriculum - adds `isFacing<Direction>` methods to Painter to conform to conventions around using be-naming for boolean accessors. This keeps the original methods around but with `@deprecated` tags to avoid breaking existing student projects.

https://codedotorg.atlassian.net/browse/JAVA-612

Tested locally.

### DIFF
--- a/org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/Painter.java
+++ b/org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/Painter.java
@@ -175,23 +175,55 @@ public class Painter {
   }
 
   /** @return True if facing North */
-  public boolean facingNorth() {
+  public boolean isFacingNorth() {
     return this.direction.isNorth();
   }
 
   /** @return True if facing East */
-  public boolean facingEast() {
+  public boolean isFacingEast() {
     return this.direction.isEast();
   }
 
   /** @return True if facing South */
-  public boolean facingSouth() {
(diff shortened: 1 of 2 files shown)


org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/Painter.java
```diff
@@ -175,25 +175,57 @@ public String getColor() {
...
   }
 
   /** @return True if facing West */
-  public boolean facingWest() {
+  public boolean isFacingWest() {
     return this.direction.isWest();
   }
 
+  /**
+   * @deprecated use {@link Painter#isFacingNorth()}
```
bencodeorg, line 198: Does a deprecation warning show up in Javalab? I can't remember if we turned off all deprecation warnings or just for certain classes

bencodeorg (approved):

org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/Painter.java
sanchitmalhotra126: I think we intentionally hide deprecated warnings to avoid confusing students. However, if Javalab docs are still being updated from the source here, then this deprecation warning should at least show up in the docs.

PR #318, opened 2022-05-24
Add CloudWatch metrics for authorization/throttling

bencodeorg:
Adds metrics for places where we return an error (ie, do not allow access) in either the HTTP or websocket authorizers.

A couple things I don't love about what I have here:
- ~~Duplicated logging code between HTTP and websocket authorizers. The validation of the token in the websocket authorizer should be refactored to look more like the validation that happens in the HTTP authorizer, which uses a helper `TokenValidator` to do the validation. The code that takes the `event` and `context` objects and deconstructs it into function names, the AWS region, etc. is kind of all over the place. I can't decide whether I should do some prefactoring before adding these metrics.~~ I did a bit of a refactor to share logging code between the two authorizers. Still would be nice to move the DynamoDB client in websocket_authorizer somewhere else, but can do that another time.
- ~~The metrics produced here are all caps, snake cased (eg, NOT_VETTED). Metrics being produced from our Java project are upper camel cased (eg, ColdBootTime). We could update the strings of the token states we use to be something more descriptive/camel cased (eg, TokenNotVetted), but would also need to update on javalab side.~~ Added a little mapping between statuses (SNAKE_CASE) and metric names (UpperCamelCase)


javabuilder-authorizer/websocket_authorizer_function.rb
```diff
@@ -81,3 +74,22 @@ def get_token_info(context, sid)
 def get_region(context)
   context.invoked_function_arn.split(':')[3]
 end
+
+def error(client, status, function_name, sid)
+  puts "TOKEN VALIDATION ERROR: #{status} token_id: #{sid}"
+  metric_data = {
+    metric_name: status,
```
molly-moen, line 81: potential solution for the all caps vs snake case issue: could we have a helper method that translates from status to metric name?

bencodeorg: Yep that could work! Do we like UpperCamelCase?

molly-moen: That's how AWS's default metrics are formatted, so I like it for that consistency.

javabuilder-authorizer/websocket_authorizer_function.rb
```diff
@@ -81,3 +74,22 @@ def get_token_info(context, sid)
 def get_region(context)
   context.invoked_function_arn.split(':')[3]
 end
+
+def error(client, status, function_name, sid)
+  puts "TOKEN VALIDATION ERROR: #{status} token_id: #{sid}"
+  metric_data = {
```
sanchitmalhotra126, line 80: Hmm yeah, it might be nice to pull this out into a helper to share between the two authorizers? Maybe a `MetricsReporter(/Helper)` with a `report_token_error(status, function_name)`? That helper module could also own the cloudwatch client then?

bencodeorg: Yeah that makes sense, I think I was getting bogged down by the fact that I think the dynamodb stuff in websocket_authorizer should also be moved into a helper. Lemme see if I can at least share the cloudwatch code between the two.

bencodeorg:
@sanchitmalhotra126 @molly-moen updated this a bit to have less duplicate code, LMK what you think!

molly-moen (approved):

sanchitmalhotra126 (approved):
LGTM!

PR #319, opened 2022-05-25
Add high websocket connection alarm

molly-moen:
Add an alarm when we see significantly higher websocket connections than normal, using AWS anomaly detection. I made the band pretty wide since we expect a lot of spikiness in this metric.

This alarm is silent, as it will be used as part of a composite alarm soon. See [this slack message](https://codedotorg.slack.com/archives/C01EF4GJ9GE/p1653501758703379) for details.

### DIFF
--- a/template.yml.erb
+++ b/template.yml.erb
@@ -587,6 +587,37 @@ Resources:
               Period: 60
               Stat: Maximum
 <%end -%>
+
+  HighWebsocketConnectionsAlarm:
+    Type: AWS::CloudWatch::Alarm
+    Properties:
+      AlarmName: !Sub "${SubDomainName}_high_websocket_connections"
+      AlarmDescription: Significantly higher websocket connections than normal detected. Investigate if there is a DDOS.
+      ActionsEnabled: false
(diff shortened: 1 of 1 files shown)


bencodeorg (approved):
To make sure I understand this "ConnectCount" represents attempts to connect, not actual connections, right? Referencing docs here:

https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-websocket-api-logging.html

sanchitmalhotra126 (approved):
Nice!

molly-moen:
@bencodeorg good clarification, yes. Seems like there isn't a way to measure actual connections

molly-moen:
@bencodeorg I'm not sure on the 20 minute question. I think it's something we should fine-tune, because just a few minutes could be a blip, but maybe 20 is too long.
For the treat missing as good, that was intentional. My thought was if there were no connections we would see no data, so that would not be a breaching data point. For performance if most minutes were slow but there were just no connections for the other minutes we would probably want to know.

Going to merge this as-is and we can fine tune the times as we get more data.

PR #320, opened 2022-05-26
Add alarm for high HTTP requests

bencodeorg:
Adds an alarm for the default HTTP request count CloudWatch metric. Uses same configuration as the websocket alarm -- not configured to notify even if threshold is breached.

## Testing story

I build the metric manually in the AWS web console (in our production account), copied the config and made some slight updates to reference the correct API ID, etc. I was able to deploy it to my dev deploy. I just deleted the prod alarm to avoid conflicts with an existing resource.

### DIFF
--- a/template.yml.erb
+++ b/template.yml.erb
@@ -628,7 +628,6 @@ Resources:
         OKActions: []
         AlarmActions: []
         InsufficientDataActions: []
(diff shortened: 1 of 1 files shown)


molly-moen (approved):
🎉

PR #322, opened 2022-05-27
Improve performance alarms

molly-moen:
We got some false alarms last night due to a small number of data points breaching the threshold for transition time. To improve I made the following changes:
- update the missing metric behavior to be `notBreaching` instead of `ignore`
- Change the evaluation period to 20 minutes but keep the datapoints to alarm at 15. Since missing data points are now good, this means that we need to see at least 15 out of 20 minutes with data that is breaching.
- Increase the width of the band for the transition time alarm. That was the one that breached last night, and the band was too narrow for the spikiness of the data.


sanchitmalhotra126 (approved):
LGTM!

PR #324, opened 2022-05-27
Move error handling logic to the handler level

sanchitmalhotra126:
This refactors our handler and code execution logic to move error handling up to the handler level. The primary goals of this change are to 1) have a central place where all exception/errors are caught and dealt with and 2) try to ensure as much as possible that no exceptions/errors go uncaught from the handler. This shouldn't result in any behavioral changes; it's mostly moving existing logic around. 

Quick summary of the main changes to hopefully make things easier to read:
- `ExceptionHandler` is a new class that houses all the exception/error handling logic. It's mostly the same as the pre-existing error handling in `CodeBuilderRunnable` with some new additions to handle new error/exception types
- In `CodeExecutionManager`, while previously all the initialization, execution, and cleanup would happen within `execute`, now callers need to manually call `shutDown` after they call `execute`. This is because in order for exceptions to bubble up to the top level, we need to allow `execute` to throw any exceptions it encounters, and afterwards call `shutDown` once all exceptions have been dealt with. This also replaces  the `requestEarlyExit` method since the timeout thread can now just call `shutDown` directly.
- In `LambdaRequestHandler` I reorganized the code to hopefully make it easier to parse. Related chunks of logic like initialization, creating the execution manager, etc, are moved into their own methods to slim down the `handle()` method a bit.

There a few cleanup/follow-up tasks I have in mind which I opted to not include since this PR is already pretty large. These include:
- Finding a way to consolidate the duplicated cleanup code between LambdaRequestHandler and AWSSystemExitHelper
- Dealing with the fact that anytime the OutputAdapter is called outside of the code execution block, we need to wrap it in a try-catch to prevent errors from propagating. A couple things we could do are 1) store the connection state on the output adapter (similar to the flag we used to have when we were running code in a thread) and 2) move the shared try-catch block to some util

Testing: tested locally and on dev lambda. Tried to simulate as many possible error types including fatal errors, internal errors, timeouts, user errors, etc.

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeExecutionManager.java
+++ b/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeExecutionManager.java
@@ -1,8 +1,6 @@
 package org.code.javabuilder;
 
-import static org.code.javabuilder.InternalFacingExceptionTypes.CONNECTION_TERMINATED;
 import static org.code.javabuilder.LambdaErrorCodes.TEMP_DIRECTORY_CLEANUP_ERROR_CODE;
-import static org.code.protocol.LoggerNames.MAIN_LOGGER;
 
 import java.io.File;
 import java.io.IOException;
@@ -10,7 +8,7 @@ import java.io.InputStream;
 import java.io.PrintStream;
 import java.nio.file.Files;
 import java.util.List;
-import java.util.logging.Logger;
+import org.code.javabuilder.util.LambdaUtils;
 import org.code.protocol.*;
 
 /**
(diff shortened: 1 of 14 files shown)


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeExecutionManager.java
```diff
@@ -162,20 +145,29 @@ private void onPreExecute() throws InternalServerException {
...
    */
-  private void onPostExecute() throws InternalServerException {
-    if (!this.executionInProgress) {
-      Logger.getLogger(MAIN_LOGGER)
-          .warning("onPostExecute() called while execution not in progress.");
-      return;
+  private void onPostExecute() {
+    // Notify user and listeners
+    try {
+      this.outputAdapter.sendMessage(new StatusMessage(StatusMessageKey.EXITED));
```
molly-moen, line 159: this logic for catching output adapter messages is repeated in `ExceptionHandler`, would it make sense for this logic to move to `AWSOutputAdapter` instead?

molly-moen (approved):
LGTM!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeExecutionManager.java
sanchitmalhotra126: Yeah, so I went back and forth on this - unfortunately the situation with the OutputAdapter is a little weird right now (definitely open to suggestions on how to streamline this). During code execution, we do want the output adapter to throw an exception if it encounters a GoneException because that's how we terminate code execution early, and we have logic in the ExceptionHandler to catch it and handle it accordingly. However, outside of code execution (which is here, in LambdaRequestHandler, and in ExceptionHandler), we don't want any exceptions to propagate because that will result in an uncaught exception being thrown from the handler. For that reason, I added these try-catch blocks, but yep that results in this block of code being copied multiple times. 

One way we could avoid copying code might be to move this into a util? something like `ProtocolUtils.safelySendMessage(OutputAdapter, ClientMessage)`?

molly-moen: yeah refactoring to a util seems like a good step

sanchitmalhotra126: Cool, I'll make a quick util for this

molly-moen (approved):

PR #313, opened 2022-05-20
Add a single ruby test

molly-moen:
This PR sets up ruby testing for `api-gateway-routes` and adds a single test. This is just to get things set up and is not meant to be full coverage of the folder yet.

This also adds the test command to the buildspec so these tests will run on commit to a PR.

## Follow-up Work
- configure unit tests for `javabuilder-authorizer`
- add more unit tests

PR #325, opened 2022-06-01
Set up unit tests for javabuilder-authorizer

sanchitmalhotra126:
Sets up unit testing and adds a single test to `javabuilder-authorizer`. This is pretty much the same as https://github.com/code-dot-org/javabuilder/pull/313, plus a single test (jwt_helper_test).

JIRA: https://codedotorg.atlassian.net/browse/JAVA-609


molly-moen (approved):
🎉

PR #329, opened 2022-06-13
Neighborhood: add show/hide buckets

molly-moen:
Add methods to the painter to show and hide buckets for the entire neighborhood. Since students only interact with the painter object we need these methods on the painter.


hannahbergam (approved):
LGTM! Will this require a corresponding pr in the monolith?

molly-moen:
@hannahbergam yep, I'm going to follow up with a code-dot-org pr once all the maze changes are in and I update the npm package

molly-moen:
Build failure is unrelated to this change

PR #330, opened 2022-06-21
Neighborhood: Refactor show/hide buckets and output adapter

molly-moen:
When working on [this change](https://codedotorg.atlassian.net/browse/JAVA-600) and trying to refactor `World.java`, I noticed we only pass `outputAdapter` around from `World` to `GridFactory` to `Grid` so we can output the show/hide bucket methods. We moved these methods up to the painter since we only expose the painter to students. I removed the show/hide painter methods from `Grid` and instead send the output message directly from `Painter` and was able to remove all references to `outputAdapter` from `World`, `GridFactory`, and `Grid`.

With this change `painter` now gets `outputAdapter` directly from `GlobalProtocol`, similar to our other mini apps.

I also updated the gitignore to ignore a neighborhood file that is auto-generated locally that we never want to commit.

### DIFF
--- a/org-code-javabuilder/.gitignore
+++ b/org-code-javabuilder/.gitignore
@@ -17,3 +17,6 @@ app/.project
 .vscode/*
 
 *.zip
+
+# ignore neighborhood grid.txt file (auto-generated when running locally)
+lib/grid.txt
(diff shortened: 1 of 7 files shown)


org-code-javabuilder/.gitignore
```diff
@@ -17,3 +17,6 @@ app/.project
 .vscode/*
 
 *.zip
+
+# ignore neighborhood grid.txt file (auto-generated when running locally)
+lib/grid.txt
```
bencodeorg, line 22: 🙌

bencodeorg (approved):
Nice cleanup!

PR #331, opened 2022-06-22
Fix template configuration for composite alarm

molly-moen:
The new composite alarm fails to deploy if in a stack that does not already contain the alarms it depends on. Fix this by adding a `DependsOn` clause to the alarm configuration in cloudformation.


sanchitmalhotra126 (approved):

molly-moen:
build failure is unrelated to this change

PR #333, opened 2022-06-30
Part 2 of adding global context

molly-moen:
Part 2 of the work started in [this PR](https://github.com/code-dot-org/javabuilder/pull/332). This PR covers the following:
- All singletons are now registered with global context (let me know if I missed any!)
- `LifecycleListener` and `LifecycleNotifier` are removed since `JavabuilderContext` now handles all cleanup of singletons

## Links
- [jira ticket](https://codedotorg.atlassian.net/browse/JAVA-600)
- [initial slack conversation](https://codedotorg.slack.com/archives/C01EF4GJ9GE/p1652827910656259)

## Testing
tested local websocket locally, and tested on a dev deploy against local dashboard and with integration tests.

## Next Steps
This is a reasonable stopping point for this work. It is worth discussing if we want to maintain `GlobalProtocol` as a unique singleton after this. It may be worth keeping around for convenience since it now only tracks the classes we need to reference a lot, and by keeping `GlobalProtocol` we can avoid a lot of typecasting.


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeExecutionManager.java
```diff
@@ -127,10 +122,7 @@ private void onPreExecute() throws InternalServerException {
     // Create the Global Protocol instance
     GlobalProtocol protocolInstance =
         new GlobalProtocol(
-            this.outputAdapter,
-            new InputHandler(this.inputAdapter),
-            this.lifecycleNotifier,
-            this.contentManager);
+            this.outputAdapter, new InputHandler(this.inputAdapter), this.contentManager);
```
sanchitmalhotra126, line 125: I love how short this list of constructor arguments has progressively gotten 😄

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaRunner.java
```diff
@@ -93,11 +90,13 @@ private boolean run(
             classLoaderUrls, JavaRunner.class.getClassLoader(), classNames, permissionLevel);
 
     boolean runResult;
-    PerformanceTracker.getInstance().trackUserCodeStart();
+    PerformanceTracker performanceTracker =
+        (PerformanceTracker) JavabuilderContext.getInstance().get(PerformanceTracker.class);
```
sanchitmalhotra126, line 94: General comment, and this can definitely be a follow-up - with these singletons being moved into JavabuilderContext, we could also now make them constructor arguments which can be more easily mocked in unit tests (instead of having to mock static objects) - for example here `performanceTracker` can be passed in via the constructor

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
```diff
@@ -26,7 +27,12 @@ private static void playInternal(Scene... scenes) {
...
     }
-
-    TheaterPlayer.getInstance().play(allActions);
+    JavabuilderContext context = JavabuilderContext.getInstance();
+    if (!context.containsKey(TheaterPlayer.class)) {
+      context.register(TheaterPlayer.class, new TheaterPlayer());
+    }
+    TheaterPlayer theaterPlayer =
+        (TheaterPlayer) JavabuilderContext.getInstance().get(TheaterPlayer.class);
+    theaterPlayer.play(allActions);
```
sanchitmalhotra126, line 36: Just an idea, and this can certainly be a follow-up if we want to do this - but this got me thinking about if we need mini-app specific initialization code that runs on every execution, similar to our overall initialization code in our handlers. Specifically, mini-apps could use this to register any singletons they want with `JavabuilderContext` so it all happens in one place before execution starts.

sanchitmalhotra126 (approved):
LGTM! The only other singleton I could find was `PlaygroundMessageHandler` but it's probably fine to skip it as Playground will be removed soon.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaRunner.java
molly-moen: are you thinking `performanceTracker` would be a constructor argument to `JavabuilderContext`?

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
molly-moen: Would we decide to run this initialization code based on some flag that is sent to Javabuilder? Right now a given instance of javabuilder can run any mini app, and I don't think we actually send the mini app type when we run code (although we certainly could).

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaRunner.java
sanchitmalhotra126: Oh I meant `performanceTracker` could be an instance variable on `JavaRunner` passed in via constructor overloading - 
```
public JavaRunner(...) {
  this(..., JavabuilderContext.get(PerformanceTracker.class), ...);
}
```
So that we way we can pass in a `performanceTracker` mock directly in unit tests rather than having to mock `JavabuilderContext` or the static `PerformanceTracker` class (this example maybe isn't the best since I don't think we currently have a test for `JavaRunner`, but just thinking in general that JavabuilderContext can allow us to use this pattern)

org-code-javabuilder/theater/src/main/java/org/code/theater/Theater.java
sanchitmalhotra126: Yeah, not sure exactly - maybe we initialize the mini-app based on the lambda? But you're right, any instance can run any mini-app, so maybe we would just initialize all mini-app resources? I think it would depend on what the initialization ends up becoming. If it's just registering static objects then it doesn't seem too bad for all mini-apps to register their static objects on startup but if it's something like memory allocation then we'd want to probably only initialize when required. This was more of a code organization idea, so I can see it going both ways.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/JavaRunner.java
molly-moen: gotcha, yeah I think I'll leave as-is but will do a follow-up if there's classes where that pattern makes sense for tests.

PR #334, opened 2022-07-13
Remove the playground Lambda

molly-moen:
Remove the playground lambda and the routing to it. We are no longer supporting playground. A request to javabuilder with `playground` as the mini app type will now return a 400.

## Next Steps
- Remove playground logic from code-dot-org
- Remove playground logic from the build and run Lambda


sanchitmalhotra126 (approved):
LGTM! To confirm, what's the user experience on playground levels now? Will they just see the standard "We had an error connecting to our server" message?

molly-moen:
> LGTM! To confirm, what's the user experience on playground levels now? Will they just see the standard "We had an error connecting to our server" message?

@sanchitmalhotra126 yep they get the generic server error message. Since the course is deprecated now the only place they should be able to even run playground is the allthethings level (which we will remove soon).

PR #337, opened 2022-07-22
Remove Playground code

bencodeorg:
Removes the `playground` package from Javabuilder. Also deletes other tangential code directly tied to playground (I searched for the string "playground", let me know if there's other playground-specific code that I might have missed with that query).

## Links

- [JAVA-623](https://codedotorg.atlassian.net/browse/JAVA-623)

## Testing story

I tested via local Javabuilder that I was able to run console, neighborhood, theater, and prompter levels after this change.

~~I'm having issues deploying due to inability to install a newer version of Ruby, so I haven't been able to test this on a deployed instance yet (we should do this before merging -- depending on how long it takes to get Ruby working with meetings week coming up, I might ask someone pull this branch and deploy it to their dev instance).~~
I was able to run projects of each level type against Molly's dev instance with these changes deployed (thanks Molly)!


molly-moen (approved):
🎉

PR #339, opened 2022-07-26
Better error state test names

bencodeorg:
The current integration test names for error states give confusing results in CodePipeline where it looks like you have errors (but actually do not):

image

Updating test names to hopefully make it a bit more clear (not sure if the red autoformatting from AWS will go away or not though).


molly-moen (approved):
Nice fix!

PR #340, opened 2022-07-28
Notify slack upon pipeline progress

cat5inthecradle:
This PR creates the notification rule and SNS topic for pipeline status changes. A separate, manually managed resource, handles the connection to slack channels. This is configured to hit the `#infra-javabuilder` channel.

### DIFF
--- a/cicd/README.md
+++ b/cicd/README.md
@@ -53,3 +53,7 @@ Finally, all of the above need some Roles to exist in the AWS accounts before we
 
 * "deploy-setup.sh"
 * "setup.template.yml" - AWS resources for the Setup infrastructure
+
+## Opportunities
+
+- Design more granular notifications by calling Slack webhooks _within_ steps.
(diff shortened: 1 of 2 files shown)


cicd/README.md
```diff
@@ -53,3 +53,7 @@ Finally, all of the above need some Roles to exist in the AWS accounts before we
 
 * "deploy-setup.sh"
 * "setup.template.yml" - AWS resources for the Setup infrastructure
+
+## Opportunities
+
+- Design more granular notifications by calling Slack webhooks _within_ steps.
```
sanchitmalhotra126, line 59: Is this straightforward to do? (and should we log a JIRA if so?)

sanchitmalhotra126 (approved):

cicd/README.md
cat5inthecradle: I think it's Jira worthy. Straightforward, but iut's just a different mechanism. We would need to implement code, likely in Bash, that posts to the same SNS topic.

PR #344, opened 2022-08-05
Update high usage and performance alarms

molly-moen:
Our alarms are too noisy now that schools are coming back. This PR makes the following changes:
- temporarily silence the high usage composite alarm. Our usage is going to continue to increase over the next month so this alarm won't be helpful. We should revisit this once usage is stabilized.
- Update the slow performance alarms to alarm on static thresholds instead of anomaly. The anomaly detection didn't go a good job for our scenario. This is because we often have two sets of performance numbers within a single metric--the cold boot and warm boot numbers. The alarm now goes off if we see 15 out of 20 minutes with an average higher than expected for cold boot. 

I've updated the cloudwatch dashboard to show the performance thresholds. We can adjust them if we find they are too sensitive or not sensitive enough. I biased towards less sensitive numbers since these alarms are intended to tell us if there is a regression.

## Performance Graphs with new thresholds
Screen Shot 2022-08-05 at 11 10 51 AM
Screen Shot 2022-08-05 at 11 10 43 AM
Screen Shot 2022-08-05 at 11 10 34 AM
Screen Shot 2022-08-05 at 11 10 24 AM

### DIFF
--- a/cicd/3-app/javabuilder/template.yml.erb
+++ b/cicd/3-app/javabuilder/template.yml.erb
@@ -658,7 +658,9 @@ Resources:
     Properties:
         ActionsEnabled: true
         AlarmActions:
-            - !If [SilenceAlertsCondition, !Ref AWS::NoValue, !Sub "arn:aws:sns:${AWS::Region}:${AWS::AccountId}:javabuilder-high-usage"]
+            # TODO: after we have run at high usage for a while, consider re-enabling this alarm. Right now it is too noisy
+            # - !If [SilenceAlertsCondition, !Ref AWS::NoValue, !Sub "arn:aws:sns:${AWS::Region}:${AWS::AccountId}:javabuilder-high-usage"]
+            - !Ref AWS::NoValue
         AlarmDescription: Send message if abnormally high Javabuilder usage detected.
             Monitors usage across the HTTP API, WebSocket API, and all Build and Run
             Lambdas.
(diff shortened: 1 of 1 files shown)


cicd/3-app/javabuilder/template.yml.erb
```diff
@@ -658,7 +658,9 @@ Resources:
     Properties:
         ActionsEnabled: true
         AlarmActions:
-            - !If [SilenceAlertsCondition, !Ref AWS::NoValue, !Sub "arn:aws:sns:${AWS::Region}:${AWS::AccountId}:javabuilder-high-usage"]
+            # TODO: after we have run at high usage for a while, consider re-enabling this alarm. Right now it is too noisy
```
bencodeorg, line 661: More a comment for when we turn this back on, but is the problem that our usage is increasing and that is what is making the alarm too noisy, or that the anomaly detection band isn't wide enough? Do you have a Jira ticket for this where I could add this comment?

cicd/3-app/javabuilder/template.yml.erb
```diff
@@ -760,133 +762,98 @@ Resources:
...
+      Threshold: 200
+      Period: 60
   
   <%=name%>SlowColdBootTimeAlarm:    
     Type: AWS::CloudWatch::Alarm
     Properties:
       AlarmName: !Sub "${SubDomainName}_<%=name.downcase%>_slow_cold_boot_time"
-      AlarmDescription: p(95) cold boot time in Javabuilder's <%=name%> build and run lambda exceeded
-        the normal cleanup time for 15 minutes. Investigate if there has been a performance regression.
+      AlarmDescription: Aerage cold boot time in Javabuilder's <%=name%> build and run lambda was high for at
```
bencodeorg, line 789: Typo

bencodeorg (approved):
Thanks for doing this 🙏

cicd/3-app/javabuilder/template.yml.erb
bencodeorg: Maybe it'll be less spiky with more ppl using it, but feel like spikiness will probably persist, especially in lower usage times (summer/holidays/weekends/nights/etc)

molly-moen: Right now I think it's some of both. This alarm may not be useful as-is, and we may just want an alarm if we are getting close to our reserved concurrency instead. I'll make a jira ticket

molly-moen: https://codedotorg.atlassian.net/browse/JAVA-648

PR #345, opened 2022-08-09
Support adhocs and deploy to production environment.

cat5inthecradle:
- Add a step to deploy to prod
- Add a step to deploy to an adhoc
- Add some parameters and a new config file to support different options
- Update CI/CD Readme

Note that this deletes "deploy.sh". Once merged, this is the only mechanism for deploying javabuilder - and javabuilder-beta will be un-updateable (unless you restore the script).


cicd/3-app/javabuilder/config/production.config.json
```diff
@@ -0,0 +1,11 @@
+{
+  "Parameters": {
+    "BaseDomainName": "code.org",
+    "BaseDomainNameHostedZonedID": "Z2LCOI49SCXUGU",
+    "ProvisionedConcurrentExecutions": "5",
```
bencodeorg, line 5: Are these supposed to be 150/1650 per what's in our deploy doc currently? Or waiting until this is actually in use to change these?

bencodeorg (commented):
A couple questions:
1. Should we keep deploy script around while javabuilder-beta is still our "prod" environment?
2. (sorry if we've discussed this and I forget) What do "dev deploys" look like in this world? Do we have no long running instances for developers (just use an adhoc)?

cicd/README.md
```diff
@@ -31,29 +7,45 @@ There are three phases of the CI/CD configuration. They are best explained in re
...
+
+By setting the `TARGET_BRANCH` you can create a new CI/CD pipeline that watches for PR's and changes to the specified branch, deploying a Test and Production environment just like the standard pipeline.
+
+```
+TARGET_BRANCH=mybranch MODE=adhoc cicd/2-cicd/deploy-cicd.sh
+```
+
+## Deploying an Adhoc environment
+
+You can create an Adhoc environment by setting the `MODE` flag on the cicd deploy script. This will create a CI/CD pipeline that will watch for updates to your `TARGET_BRANCH`.
```
molly-moen, line 47: Will the adhocs deploy to dev or prod? And what will the stack name be for a given adhoc?

deploy.sh
```diff
@@ -1,63 +0,0 @@
-#!/bin/bash -xe
```
molly-moen, line 1: +1 to Ben's comment, I think we should not delete this until prod javabuilder is fully switched over to the pipeline version. Otherwise we can't update prod right?

bencodeorg:
Also copying over a comment Jessie added to the deploy doc a few months ago:

> When we update this to no longer be "beta," let's also use that as an opportunity to update the object names in the template.yml to use "Api" instead of "API" when camel casing and to use WebSocket consistently rather than mixing in Websocket at times.

cicd/2-cicd/cicd.template.yml
```diff
@@ -26,11 +26,19 @@ Parameters:
...
 
 Conditions:
-  TargetsMainBranch: !Equals 
-    - !Ref GitHubBranch
-    - main
+  TargetsMainBranch: !Equals [ !Ref GitHubBranch, main ]
+  DeployToAdhoc: !Equals [!Ref Mode, adhoc ]
+  # DeployToTest and DeployToProd are the same right now
+  DeployToTest: !Equals [!Ref Mode, standard ]
+  DeployToProd: !Equals [ !Ref Mode, standard ]
```
sanchitmalhotra126, line 40: To clarify - what are these conditions for? Wouldn't the pipeline always deploy to both test and prod?

cicd/2-cicd/deploy-cicd.sh
```diff
@@ -2,14 +2,20 @@
 
 echo Deploying Javabuilder CICD Pipeline
 
-# Create/Update the Javabuilder build/deploy pipeline stack. This is manually created and maintained, but should not require elevated permissions. 
+# Create/Update the Javabuilder build/deploy pipeline stack. This is manually created and maintained, but should not require elevated permissions.
+# Options include:
+# - TARGET_BRANCH: Defaults to `main`, passed as a Parameter for "cicd/2-cicd/cicd.template.yml"
+# - MODE: Can be 'standard' (default) or 'adhoc', passed as a Parameter for "cicd/2-cicd/cicd.template.yml"
+# - GITHUB_BADGE_ENABLED: defaults to true, passed as a Parameter for "cicd/2-cicd/cicd.template.yml"
+# - STACK_NAME: defaults to an automated naming convention, no current use cases for changing this.
```
sanchitmalhotra126, line 10: Are adhocs named by `TARGET_BRANCH` or `STACK_NAME`?

cicd/3-app/javabuilder/config/production.config.json
cat5inthecradle: No intentional change, I'll double check these and correct them.

cicd/README.md
cat5inthecradle: Following these instructions, they will deploy to prod. This would create a stack called 'javabuilder-mybranch' that would deploy another stack called 'javabuilder-adhoc-mybranch' [per this config](https://github.com/code-dot-org/javabuilder/blob/8664d0fc20851434f2edfa5ac037413ed4cb9507/cicd/2-cicd/cicd.template.yml#L313)

That said, I think the procedure to deploy to Dev would be:
- authenticate with Dev instead of prod
- change adhoc.config to use dev-code.org and the appropriate hosted zone ID
- ensure that the stack from "1-Setup" has been deployed to Dev.

If sticking adhocs in Dev would be the preference (makes sense to me) I can modify these instructions and make sure that works.

deploy.sh
cat5inthecradle: Correct. There is risk of configuration drift, as the pipeline replicates some of the configuration and work done in the deploy script.

I'll undelete it, and make sure they're in-sync as of this PR (taking the script as the source of truth)

cicd/2-cicd/cicd.template.yml
cat5inthecradle: Yeah, I went down a route of a scenario where you might want to deploy to Test but not Prod, but didn't see the value.

I could use two conditions [AdhocMode, StandardMode]. My dilemma is whether to name the condition based on the inputs, or based on the output the condition enables.

cicd/README.md
molly-moen: I'm not sure if that's the preference or not, I'll leave it up to you whether you think that's valuable. Only potential reason to put them in dev is we do have account-wide limits on some things, but it's pretty hard to make a dev instance hit those limits

cicd/2-cicd/deploy-cicd.sh
cat5inthecradle: I think I'm going to get rid of STACK_NAME because it would be safer to just let the script generate the stack name. STACK_NAME in this case is solely referring to the CI/CD resources stack, and not the application resources stack.

If TARGET_BRANCH is 'frog', then this script will create a stack called "javabuilder-frog-cicd" which defines a Pipeline that will create "javabuilder-frog-test" and "javabuilder-frog" stacks.

(there is some logic to handle 'main', which omits the branch from the stacks, if the branch is 'main' you just get 'javabuilder-cicd', 'javabuilder-test', and 'javabuilder' stacks)

cicd/README.md
cat5inthecradle: I copied the template and moved it back to the root. I also renamed "deploy.sh" to "deploy-beta.sh"

The template within the cicd configuration includes the API>Api changes, so from here on these templates have diverged, but we can still maintain the beta environment.

cicd/2-cicd/cicd.template.yml
cat5inthecradle: cleaned these up.

cicd/2-cicd/deploy-cicd.sh
cat5inthecradle: cleaned this up.

cat5inthecradle:
I'm destroying and completely recreating the javabuilder-cicd stack, including the test environment. This is to ensure that all of this code works from a blank slate.

cat5inthecradle:
Pipeline failed, assuming unrelated.
https://us-east-1.console.aws.amazon.com/codesuite/codepipeline/pipelines/javabuilder-cicd/view?region=us-east-1

cicd/3-app/javabuilder/config/adhoc.config.json
```diff
@@ -0,0 +1,11 @@
+{
+  "Parameters": {
+    "BaseDomainName": "code.org",
+    "BaseDomainNameHostedZonedID": "Z2LCOI49SCXUGU",
+    "ProvisionedConcurrentExecutions": "1",
+    "ReservedConcurrentExecutions": "3",
+    "LimitPerHour": "15",
```
molly-moen, line 7: can we set the hourly limit to 50 and the daily limit to 150? For dev instances we'll want > 15 usages per hour

cicd/3-app/javabuilder/config/adhoc.config.json
```diff
@@ -0,0 +1,11 @@
...
+    "LimitPerDay": "40",
+    "SilenceAlerts": "false"
```
molly-moen, line 9: also if adhocs are on prod we should set SilenceAlerts to true. Anything on the dev account gets silenced by default in the template

molly-moen (commented):
One more config comment, otherwise looks good!

cat5inthecradle:
new adhoc config is: 

```
    "ProvisionedConcurrentExecutions": "1",
    "ReservedConcurrentExecutions": "3",
    "LimitPerHour": "50",
    "LimitPerDay": "150",
    "SilenceAlerts": "true"
```

cat5inthecradle:
@molly-moen FYI for adhocs, if you edit the config file and commit it to the adhoc branch, it will use that config when it redeploys it - so if there are special cases where we want to test different config, we can.

molly-moen (approved):
🎉

PR #346, opened 2022-08-10
Fix throttling alarms

bencodeorg:
Adds new metrics (`NewClassroomBlocked`, `NewUserBlocked`) to track when a classroom or user is blocked for the first time. Sets up our alarms to use these metrics, rather than our existing metrics, which track whenever we get a request that is blocked because the user or the user's classroom is blocked. These existing metrics will continue to generate data any time we block a request, but will not be used in our alarms.

## Testing story

I deployed these changes to my dev instance. I was able to see data generated for each of these new metrics (only once per user), and saw that the old metrics generated more than one event as I repeatedly made requests with an account that had been blocked. I also observed javabuilder functioning normally when no throttling thresholds were breached.

### DIFF
--- a/javabuilder-authorizer/metrics_reporter.rb
+++ b/javabuilder-authorizer/metrics_reporter.rb
@@ -11,8 +11,13 @@ class MetricsReporter
 
   def log_token_error(status, error_message)
     puts error_message
+    metric_name = TOKEN_STATUS_METRIC_NAMES[status]
+    log(metric_name)
+  end
+
+  def log(metric_name)
     metric_data = {
-      metric_name: TOKEN_STATUS_METRIC_NAMES[status],
+      metric_name: metric_name,
       dimensions: [
         {
           name: "functionName",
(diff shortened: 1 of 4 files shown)


molly-moen (approved):

javabuilder-authorizer/metrics_reporter.rb
```diff
@@ -11,8 +11,13 @@ def initialize(context)
 
   def log_token_error(status, error_message)
     puts error_message
+    metric_name = TOKEN_STATUS_METRIC_NAMES[status]
+    log(metric_name)
+  end
+
+  def log(metric_name)
```
sanchitmalhotra126, line 18: Do we want to add a `puts` here as well (like we have in `log_token_error`) or is logging the error sufficient? Just thinking if we'd want to look this up in our logs for any reason

sanchitmalhotra126 (approved):
One question but LGTM otherwise!

javabuilder-authorizer/metrics_reporter.rb
bencodeorg: mmm I guess I figured we're logging some detail to DynamoDB at the point of blocking a user/classroom, so that serves as our log in this case. I'm not sure what our best practice is here though.

sanchitmalhotra126: Yeah, totally not required, i just mentioned it since we do have a `puts` statement in the `log_token_error` function for the error message. I'd say fine to ship as is and we can always add in more logging if it becomes necessary?

PR #348, opened 2022-08-15
Force system exit on out of memory error

molly-moen:
We have seen some issues with out of memory errors causing a lambda instance to stay in a bad state. This forces a system exit if we see an out of memory error, so only one user will feel the impact of that error.

Note: I've also left in a helper method for logger info statements. It isn't used here but I often find I need something like it while debugging on a deployed instance of Javabuilder.

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/ExceptionHandler.java
+++ b/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/ExceptionHandler.java
@@ -32,6 +32,18 @@ public class ExceptionHandler {
       return;
     }
 
+    // We should also exit on an OutOfMemoryError. This means something is wrong with our memory
+    // usage
+    // and is not likely to be recoverable.
+    if (e instanceof OutOfMemoryError) {
+      final InternalServerException error =
+          new InternalServerException(InternalExceptionKey.UNKNOWN_ERROR, e);
+      LoggerUtils.logSevereError(error);
+      LambdaUtils.safelySendMessage(this.outputAdapter, error.getExceptionMessage(), true);
+      this.systemExitHelper.exit(LambdaErrorCodes.OUT_OF_MEMORY_ERROR_CODE);
+      return;
+    }
+
     // Internal server exceptions are caused by us (essentially an HTTP 5xx error). Log and notify
     // the user.
     if (e instanceof InternalServerException || e instanceof InternalServerRuntimeException) {
--- a/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/ExceptionHandler.java
+++ b/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/ExceptionHandler.java
@@ -32,8 +32,8 @@ public class ExceptionHandler {
       return;
     }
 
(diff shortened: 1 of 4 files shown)


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/ExceptionHandler.java
```diff
@@ -32,6 +32,18 @@ public void handle(Throwable e) {
       return;
     }
 
+    // We should also exit on an OutOfMemoryError.
+    // This means something is wrong with our memory usage
+    // and is not likely to be recoverable.
+    if (e instanceof OutOfMemoryError) {
+      final InternalServerException error =
+          new InternalServerException(InternalExceptionKey.UNKNOWN_ERROR, e);
```
sanchitmalhotra126, line 40: Should this be wrapped in a `FatalError` instead since we take similar steps to shut down the runtime? Don't think it makes any difference for logging purposes, but we could also maybe consolidate this block with the block above

org-code-javabuilder/protocol/src/main/java/org/code/protocol/LoggerUtils.java
```diff
@@ -108,6 +108,10 @@ public static void logTrackingException(Throwable e) {
     Logger.getLogger(MAIN_LOGGER).warning(eventData.toString());
   }
 
+  public static void logInfo(String info) {
```
sanchitmalhotra126, line 111: Nice! Definitely seems helpful. I think we even might want to eventually replace all calls to `Logger.getLogger(...)` with `LoggerUtils` given some of the weirdness around the logger object potentially getting GC'd.

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/ExceptionHandler.java
molly-moen: If we wrap in `FatalError` this error message is sent to users: `Exception message: java.lang.OutOfMemoryError: Metaspace`. If we keep it as-is we get the "we had an error on our side" message, which I think makes more sense for this scenario. So I'm incline to leave it as-is even though it's a bit repetitive.

org-code-javabuilder/protocol/src/main/java/org/code/protocol/LoggerUtils.java
molly-moen: I agree, it would make sense now that we have the utils to only use that to talk to the logger

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/ExceptionHandler.java
sanchitmalhotra126: Cool gotcha, sounds good. That being said, I think it would probably make sense for FatalErrors to also display the "we had an error on our side" message, similar to InternalExceptions right? I think the error is being printed because we don't have handling for it client side. I can make a separate task to look into that.

molly-moen: Yeah that would be good!

PR #349, opened 2022-08-17
Update error rate alarms

molly-moen:
Update the error rate and severe error rate alarms:
- The severe error rate alarm was previously using a log filter, but we added a CloudWatch metric to track severe errors a while ago. Use this metric instead, as the log filter only applies to `javabuilderbeta`. 
- Timeouts as reported as errors by AWS. We had updated our dashboard to calculate the error rate by subtracting instances that took close to the timeout time (90 seconds) from the error count to get a more accurate percentage. Since all of our lambdas now have a timeout of 90 seconds we should use that calculation here. This calculation isn't perfect--for some reason it will sometimes report a decimal number instead of an integer. But it gets us closer to the actual error rate so I thought it was a worthwhile change.

In working on this I also tested deploying an adhoc and noticed that the script had an error in the string interpolation for the adhoc stack name, this is fixed in this PR. I also added a note to the README that you can't deploy an adhoc based on a branch with a `\` character in the name.

### DIFF
--- a/beta-template.yml.erb
+++ b/beta-template.yml.erb
@@ -700,7 +700,10 @@ Resources:
             MetricStat:
                 Metric:
                     Namespace: Javabuilder
-                    MetricName: Javabuilder <%=name%> "SEVERE" errors
+                    MetricName: SevereError
+                    Dimensions:
+                      - Name: functionName
+                        Value: !Ref BuildAndRunJava<%=name%>ProjectFunction
                 Period: 300
                 Stat: Sum
           - Id: m2
@@ -734,7 +737,7 @@ Resources:
           - Id: e1
             Label: Errors / Invocations
             ReturnData: true
-            Expression: (m1 / m2) * 100
+            Expression: ((m1 - m3) / m2) * 100
           - Id: m1
             ReturnData: false
             MetricStat:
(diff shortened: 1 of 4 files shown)


sanchitmalhotra126 (commented):
Curious - does subtracting the instances close to 90s change our error rate significantly? I know timeouts are recorded as errors in lambda but SEVERE errors are something that we report manually, and in the case of timeouts I think we only report a WARNING right?

molly-moen:
> Curious - does subtracting the instances close to 90s change our error rate significantly? I know timeouts are recorded as errors in lambda but SEVERE errors are something that we report manually, and in the case of timeouts I think we only report a WARNING right?

@sanchitmalhotra126 it depends if we have a lot of timeouts or not. I think as we get more theater projects running we will see more timeouts, which will influence our error rate alarm. Ideally we shouldn't report on expected behavior.

sanchitmalhotra126:
> @sanchitmalhotra126 it depends if we have a lot of timeouts or not. I think as we get more theater projects running we will see more timeouts, which will influence our error rate alarm. Ideally we shouldn't report on expected behavior.

Right right, I think what I'm not sure about is if timeouts ever do get reported into our SEVERE error rate graphs because we don't actually log a severe error on timeouts right? I know they get reported into the lambda error graphs because lambda treats them as an error, but our manual reporting doesn't seem to.

molly-moen:
@sanchitmalhotra126 that's right, I only made the change for our error rate alarm which uses the automatically reported lambda error metric. The severe alarm doesn't change here except to use the metric instead of the log filter.

sanchitmalhotra126:
> @sanchitmalhotra126 that's right, I only made the change for our error rate alarm which uses the automatically reported lambda error metric. The severe alarm doesn't change here except to use the metric instead of the log filter.

Ahh my bad - I thought it was part of the same alarm definition. Thanks for clarifying!

sanchitmalhotra126 (approved):
LGTM! Thanks for the clarification!

PR #350, opened 2022-08-18
Validate full main method signature

sanchitmalhotra126:
Validates the main method signature (if present) to ensure we don't hit any unexpected null pointer exceptions, or other exceptions related to visibility and static modifiers. We will now throw a user facing exception unless the method is public, static, void return type, and has only one argument of a String[].

Javalab PR to translate this message: https://github.com/code-dot-org/code-dot-org/pull/47718

https://codedotorg.atlassian.net/browse/JAVA-665

Tested locally.

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/util/ProjectLoadUtils.java
+++ b/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/util/ProjectLoadUtils.java
@@ -1,6 +1,7 @@
 package org.code.javabuilder.util;
 
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 import java.net.URLClassLoader;
 import java.util.List;
 import org.code.javabuilder.JavaProjectFile;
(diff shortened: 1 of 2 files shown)


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/util/ProjectLoadUtils.java
```diff
@@ -46,6 +44,20 @@ public static Method findMainMethod(URLClassLoader classLoader, List<JavaProject
...
+    // If we found a main method, make sure the method signature is valid
+    // (public, static, void return type, one argument which should be a String[]).
+    if (mainMethod != null) {
+      final Class<?>[] parameterTypes = mainMethod.getParameterTypes();
+      final int modifiers = mainMethod.getModifiers();
+      if (!Modifier.isPublic(modifiers)
+          || !Modifier.isStatic(modifiers)
+          || mainMethod.getGenericReturnType() != Void.TYPE
+          || parameterTypes.length != 1
+          || !parameterTypes[0].equals(String[].class)) {
```
molly-moen, line 56: nice!

molly-moen (approved):

PR #353, opened 2022-08-23
Verify api client still exists on startup

molly-moen:
We have seen the api client go away the last couple mornings. This causes subsequent invocations of a container to fail. This verifies the client exists when an instance starts, and recreates it if necessary.

Tested on a dev instance. If I manually called shutdown on the api client the check will catch the error and recreate itself, and the end user will not see any errors.

We could also do this for our other aws clients (SQS and S3), but we haven't seen those have issues yet so I've left them as-is.


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
```diff
@@ -371,4 +377,20 @@ private void cleanUpAWSResources(String connectionId, AmazonApiGatewayManagement
       Logger.getLogger(MAIN_LOGGER).removeHandler(allHandlers[i]);
     }
   }
+
+  private void verifyApiClient(String connectionId) {
+    GetConnectionRequest connectionRequest =
```
sanchitmalhotra126, line 382: Should we log a warning/info here just to see if it's still happening, since we don't exactly know the root cause?

molly-moen: good call, I'll add a warning log

sanchitmalhotra126 (approved):

PR #357, opened 2022-08-24
Neighborhood Validation Updates

molly-moen:
We wanted to be able to track the following in validation code:
- the Painter's start and end direction
- If `canMove`, `isOnBucket` and/or `isOnPaint` were called.

To track direction I've extended `Position` to also include direction and have a getter for direction.

The reason we didn't previously see events for `canMove`, etc. is they do not trigger an animation and therefore were not sent over the output adapter. We only look at output adapter methods in our neighborhood tracking. To fix this I've added output events for these methods.

I also added the result of the method call (true or false) to the output message even though it wasn't necessary for this use case. The result makes the message over the websocket less confusing, because we would send something such as `CAN_MOVE` with no more information when `canMove` was called and it was false. Adding the result makes this message less confusing and opens up the ability to add some indicator of the function call on the frontend if we chose to in the future.

### DIFF
--- a/org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/Painter.java
+++ b/org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/Painter.java
@@ -143,12 +143,20 @@ public class Painter {
 
   /** @return True if there is paint in the square where the painter is standing. */
   public boolean isOnPaint() {
-    return this.grid.getSquare(this.xLocation, this.yLocation).hasColor();
+    boolean isOnPaint = this.grid.getSquare(this.xLocation, this.yLocation).hasColor();
+    this.outputAdapter.sendMessage(
+        new NeighborhoodSignalMessage(
+            NeighborhoodSignalKey.IS_ON_PAINT, this.getSignalDetailsForBooleanMethod(isOnPaint)));
+    return isOnPaint;
   }
 
   /** @return True if there is a paint bucket in the square where the painter is standing. */
   public boolean isOnBucket() {
-    return this.grid.getSquare(this.xLocation, this.yLocation).containsPaint();
+    boolean isOnBucket = this.grid.getSquare(this.xLocation, this.yLocation).containsPaint();
+    this.outputAdapter.sendMessage(
+        new NeighborhoodSignalMessage(
+            NeighborhoodSignalKey.IS_ON_BUCKET, this.getSignalDetailsForBooleanMethod(isOnBucket)));
+    return isOnBucket;
   }
 
   /** @return True if the painter's personal bucket has paint in it. */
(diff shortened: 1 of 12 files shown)


org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/Painter.java
```diff
@@ -293,6 +305,13 @@ private HashMap<String, String> getSignalDetails() {
     return details;
   }
 
+  private HashMap<String, String> getSignalDetailsForBooleanMethod(boolean result) {
+    HashMap<String, String> details = this.getSignalDetails();
+    String resultString = result ? "true" : "false";
```
sanchitmalhotra126, line 310: nit: I think you could just do `String.valueOf(result)` here?

sanchitmalhotra126 (approved):
LGTM!

org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/Painter.java
```diff
@@ -143,12 +143,20 @@ public void takePaint() {
 
   /** @return True if there is paint in the square where the painter is standing. */
   public boolean isOnPaint() {
-    return this.grid.getSquare(this.xLocation, this.yLocation).hasColor();
+    boolean isOnPaint = this.grid.getSquare(this.xLocation, this.yLocation).hasColor();
+    this.outputAdapter.sendMessage(
+        new NeighborhoodSignalMessage(
+            NeighborhoodSignalKey.IS_ON_PAINT, this.getSignalDetailsForBooleanMethod(isOnPaint)));
```
madelynkasula, line 149: not sure how strongly i feel about this, but all of the calls to `sendMessage` look pretty similar, so we could add a helper to simplify them like

```java
public void sendBooleanMessage(NeighborhoodSignalKey key, boolean value) {
  this.outputAdapter.sendMessage(
    new NeighborhoodSignalMessage(key, this.getSignalDetailsForBooleanMethod(value));
  );
}

public boolean isOnPaint() {
  this.sendBooleanMessage(NeighborhoodSignalKey.IS_ON_PAINT, isOnPaint);
}
```

madelynkasula (approved):
left a nit comment, but it's not a merge-blocker!

org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/Painter.java
molly-moen: yep, updated!

org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/Painter.java
molly-moen: updated!

PR #359, opened 2022-08-30
Bump concurrency and limits for javabuilder-test

sanchitmalhotra126:
Bumping up the hourly/daily limits and concurrency on the javabuilder-test stack a bit so that multiple developers can develop against it without too easily running into throttling or rate limiting. I matched the daily and hourly limits to the adhoc config, but the concurrency is just a ballpark estimate based on how many developers might be working on it concurrently, so let me know if these should be adjusted.

For reference - [this change](https://github.com/code-dot-org/code-dot-org/pull/47886) makes the javabuilder-test stack the default for local development, so we would expect its usage to increase.

### DIFF
--- a/cicd/3-app/javabuilder/config/test.config.json
+++ b/cicd/3-app/javabuilder/config/test.config.json
@@ -2,10 +2,10 @@
   "Parameters": {
     "BaseDomainName": "code.org",
     "BaseDomainNameHostedZonedID": "Z2LCOI49SCXUGU",
-    "ProvisionedConcurrentExecutions": "1",
-    "ReservedConcurrentExecutions": "3",
-    "LimitPerHour": "15",
-    "LimitPerDay": "40",
+    "ProvisionedConcurrentExecutions": "5",
+    "ReservedConcurrentExecutions": "25",
+    "LimitPerHour": "50",
+    "LimitPerDay": "150",
     "SilenceAlerts": "false"
   }
 }


cicd/3-app/javabuilder/config/test.config.json
```diff
@@ -2,10 +2,10 @@
   "Parameters": {
     "BaseDomainName": "code.org",
     "BaseDomainNameHostedZonedID": "Z2LCOI49SCXUGU",
-    "ProvisionedConcurrentExecutions": "1",
-    "ReservedConcurrentExecutions": "3",
-    "LimitPerHour": "15",
-    "LimitPerDay": "40",
+    "ProvisionedConcurrentExecutions": "5",
+    "ReservedConcurrentExecutions": "25",
```
molly-moen, line 6: this seems fine to me, we've only just started to see 25 concurrent executions on prod

cicd/3-app/javabuilder/config/test.config.json
```diff
@@ -2,10 +2,10 @@
...
+    "LimitPerHour": "50",
+    "LimitPerDay": "150",
```
molly-moen, line 8: we should document somewhere how to reset yourself if you hit the limit

molly-moen (approved):
LGTM! One comment on documentation but that will need to be figured out separately.

cicd/3-app/javabuilder/config/test.config.json
sanchitmalhotra126: Good call. It should probably belong in the same doc that DOTDs can reference for Javabuilder issues I think? I'll go ahead and start one and we can continue adding to it.

sanchitmalhotra126: Started a doc here: https://docs.google.com/document/d/1bHvV6pvUcwxgZpw0YWBmxFggQL5KqYx9zwolwkZhjU8/edit#

cat5inthecradle (approved):

cat5inthecradle:
No special deploy steps required for this. Once merged the pipeline will use the new values when deploying to Test.

PR #360, opened 2022-09-01
Add validation api for checking system out messages

molly-moen:
Levelbuilders would like to be able to check what was written to the console in their tests. This adds a basic api to do that. The usage on the levelbuilder side is:
```
import org.code.validation.*;
...
List<String> messages = SystemOutTestRunner.run();
```

We do this tracking similarly to neighborhood messages. Any system out message goes through `UserTestOutputAdapter`, and instead of ignoring them we parse and track any that are not `\n`. `\n` messages indicate the user used a `println` vs a `print`.

I waffled on whether or not to create an interface for the tracker type (as we now have `NeighborhoodTracker`, `PainterTracker` and `SystemOutTracker`), but ultimately decided against it as the `getLog` method needs a different return type for each version. Therefore adding an interface would really only get us some default methods `trackEvent` and `reset` (which currently isn't useful for `PainterTracker`), which didn't seem helpful enough. I am open to suggestion if we think that adding an interface would gain us other benefits.

I've documented this api (and our other validation api) in [this document](https://docs.google.com/document/d/1t_NEkBEhya70Vl17Ra07Tq3Wk1xYdpU8TYNagGiuOFM/edit).

### DIFF
--- a/org-code-javabuilder/validation/src/main/java/org/code/validation/SystemOutTestRunner.java
+++ b/org-code-javabuilder/validation/src/main/java/org/code/validation/SystemOutTestRunner.java
@@ -5,6 +5,8 @@ import org.code.protocol.JavabuilderContext;
 import org.code.validation.support.ValidationProtocol;
 
 public class SystemOutTestRunner {
+  // Run the main method of the user's program and return a list
+  // of all system.out messages in order.
   public static List<String> run() {
     ValidationProtocol protocolInstance =
         (ValidationProtocol) JavabuilderContext.getInstance().get(ValidationProtocol.class);
--- a/org-code-javabuilder/validation/src/main/java/org/code/validation/SystemOutTestRunner.java
+++ b/org-code-javabuilder/validation/src/main/java/org/code/validation/SystemOutTestRunner.java
@@ -1,0 +1,14 @@
+package org.code.validation;
+
+import java.util.List;
+import org.code.protocol.JavabuilderContext;
+import org.code.validation.support.ValidationProtocol;
+
+public class SystemOutTestRunner {
+  public static List<String> run() {
+    ValidationProtocol protocolInstance =
+        (ValidationProtocol) JavabuilderContext.getInstance().get(ValidationProtocol.class);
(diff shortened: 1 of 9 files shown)


org-code-javabuilder/validation/src/main/java/org/code/validation/SystemOutTestRunner.java
```diff
@@ -0,0 +1,16 @@
+package org.code.validation;
+
+import java.util.List;
+import org.code.protocol.JavabuilderContext;
+import org.code.validation.support.ValidationProtocol;
+
+public class SystemOutTestRunner {
```
sanchitmalhotra126, line 7: Would it make sense to just have one `ValidationTestRunner` that runs the main method once, and has methods like `getSystemOutMessages`, `getNeighborhoodLog` etc? Do we anticipate validation writers needing to verify both system out and neighborhood messages in a single test?

sanchitmalhotra126 (commented):
Looks good! One question regarding validation test runners.

Also, I agree with your assessment on making a tracker interface. If we do end up with a lot more down the line, having an interface may be helpful just for code maintenance (for stuff like `for (Tracker tracker : this.trackers) { tracker.reset(); }`), but doesn't feel super necessary for now.

jmkulwik:
> I've documented this api (and our other validation api) in [this document](https://docs.google.com/document/d/1t_NEkBEhya70Vl17Ra07Tq3Wk1xYdpU8TYNagGiuOFM/edit).

Just hopping in to say, that doc is super useful :) Thanks for keeping the documentation up to date!

org-code-javabuilder/validation/src/main/java/org/code/validation/SystemOutTestRunner.java
molly-moen: I wanted to make the api as simple as possible (so a single call both runs & gets the main method). We also shouldn't change the neighborhood api now since it's in use.

sanchitmalhotra126 (approved):

PR #362, opened 2022-09-01
Ignore unused output messages in basic output adapters

molly-moen:
I added additional painter output messages in [this PR](https://github.com/code-dot-org/javabuilder/pull/357) that are only used in validation runs. Sending these over the websocket had a performance impact for large neighborhood projects. In this PR I added a method that sets the default send status for a client message. For every message but the new neighborhood ones that method returns true, and for the new neighborhood messages it returns false.

## Testing
Tested locally and on an adhoc. Tested that normal execution no longer receives these messages and validation on them still works.

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketOutputAdapter.java
+++ b/org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketOutputAdapter.java
@@ -20,7 +20,7 @@ public class WebSocketOutputAdapter implements OutputAdapter {
 
   @Override
   public void sendMessage(ClientMessage message) {
-    if (message.shouldSendInRunMode()) {
+    if (message.shouldAlwaysSend()) {
       try {
         endpoint.sendText(message.getFormattedMessage());
       } catch (IOException e) {
@@ -32,7 +32,7 @@ public class WebSocketOutputAdapter implements OutputAdapter {
   }
 
   public void sendDebuggingMessage(ClientMessage message) {
-    if (message.shouldSendInRunMode()) {
+    if (message.shouldAlwaysSend()) {
       try {
         endpoint.sendText(message.getFormattedMessage());
       } catch (IOException e) {
--- a/org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketOutputAdapter.java
+++ b/org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketOutputAdapter.java
@@ -20,22 +20,26 @@ public class WebSocketOutputAdapter implements OutputAdapter {
 
   @Override
   public void sendMessage(ClientMessage message) {
-    try {
-      endpoint.sendText(message.getFormattedMessage());
-    } catch (IOException e) {
-      e.printStackTrace();
-    } catch (IllegalStateException e) {
-      throw new InternalServerRuntimeException(InternalExceptionKey.CONNECTION_TERMINATED, e);
+    if (message.shouldSendInRunMode()) {
+      try {
+        endpoint.sendText(message.getFormattedMessage());
+      } catch (IOException e) {
+        e.printStackTrace();
+      } catch (IllegalStateException e) {
+        throw new InternalServerRuntimeException(InternalExceptionKey.CONNECTION_TERMINATED, e);
+      }
     }
   }
 
   public void sendDebuggingMessage(ClientMessage message) {
-    try {
-      endpoint.sendText(message.getFormattedMessage());
-    } catch (IOException e) {
-      e.printStackTrace();
-    } catch (IllegalStateException e) {
-      throw new InternalServerRuntimeException(InternalExceptionKey.CONNECTION_TERMINATED, e);
+    if (message.shouldSendInRunMode()) {
+      try {
+        endpoint.sendText(message.getFormattedMessage());
+      } catch (IOException e) {
+        e.printStackTrace();
+      } catch (IllegalStateException e) {
+        throw new InternalServerRuntimeException(InternalExceptionKey.CONNECTION_TERMINATED, e);
+      }
     }
   }
 }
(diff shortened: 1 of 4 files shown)


org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketOutputAdapter.java
```diff
@@ -20,22 +20,26 @@ public WebSocketOutputAdapter(Session session) {
 
   @Override
   public void sendMessage(ClientMessage message) {
-    try {
-      endpoint.sendText(message.getFormattedMessage());
-    } catch (IOException e) {
-      e.printStackTrace();
-    } catch (IllegalStateException e) {
-      throw new InternalServerRuntimeException(InternalExceptionKey.CONNECTION_TERMINATED, e);
+    if (message.shouldAlwaysSend()) {
```
bencodeorg, line 23: Nit: feels more readable if we reversed the logic and could then just early return like `return if message.skipSend();`? Or at least shorten name to `shouldSend()`, what does the "always" represent?

org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/support/NeighborhoodSignalMessage.java
```diff
@@ -1,11 +1,27 @@
...
 public class NeighborhoodSignalMessage extends ClientMessage {
   public NeighborhoodSignalMessage(NeighborhoodSignalKey key, HashMap<String, String> detail) {
     super(ClientMessageType.NEIGHBORHOOD, key.toString(), detail);
   }
+
+  @Override
+  public boolean shouldAlwaysSend() {
+    String signalKey = this.getValue();
+    Set<String> ignoredSignalKeys = new HashSet<>();
+    // These keys are only used for testing, don't send in run mode.
```
bencodeorg, line 18: Is this method only called when in "run mode"?

molly-moen: I switched the naming convention partway through, I'll clean this up

org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketOutputAdapter.java
molly-moen: I had some struggles with naming on this one. The challenge is the specific neighborhood messages should only send when they hit `UserTestOutputAdapter`, and all other output adapters should ignore them. So `always` was supposed to signify every output adapter should send these. `skipSend` implies every output adapter should skip to me, which isn't quite what I intended, but if that makes more sense to you I can change it.

bencodeorg: > skipSend implies every output adapter should skip to me, which isn't quite what I intended, but if that makes more sense to you I can change it.

Every output adapter should skip sending these messages (except `UserTestOutputAdapter`), right? I'm suggesting set the default skipSend() to return `false`, and update the override method to return `true` where appropriate?

org-code-javabuilder/neighborhood/src/main/java/org/code/neighborhood/support/NeighborhoodSignalMessage.java
molly-moen: updated the comment, let me know if it makes more sense

org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketOutputAdapter.java
molly-moen: so the issue is that the client message doesn't know which output adapter it is sending to, so it sends the default value every time. Then the output adapter decides to ignore that skip for some messages. Which is not a very clean pattern. I didn't want to have every output adapter need to parse every message if they did not need to.

bencodeorg (approved):
Sorry for deep diving on naming, interesting to hear how you think about this stuff. LGTM!

org-code-javabuilder/lib/src/main/java/dev/javabuilder/WebSocketOutputAdapter.java
molly-moen: yep!

molly-moen:
> Sorry for deep diving on naming, interesting to hear how you think about this stuff. LGTM!

@bencodeorg they were good questions!

org-code-javabuilder/protocol/src/main/java/org/code/protocol/ClientMessage.java
```diff
@@ -67,4 +67,12 @@ public String getFormattedMessage() {
     }
     return formattedMessage.toString();
   }
+
+  /**
+   * @return whether or not this message should always be sent by any output adapter. Some messages
+   *     are only relevant for the test output adapter
+   */
+  public boolean shouldAlwaysSend() {
```
sanchitmalhotra126, line 75: I know this is already merged (and no issues with it being merged as is because I know it was fixing a prod issue) but had a couple thoughts if we want to consider a follow up -

I think the naming of `shouldAlwaysSend` feels a little confusing because the logic specifically pertains to the UserTestOutputAdapter. If we want to stick with this approach, maybe naming it something like `sendInTestsOnly()` or something might be clearer?

Another thought that could work here, especially if we might have other cases where we skip specific output adapters: 
We could add a `AdapterType/String getType()` to the `OutputAdapter` interface, and then have a method on `ClientMessage` like `List<String/AdapterType> adaptersToSkip()` that returns which adapter types the message should not be sent for. That way, if WebSocket and AWS both return the same type, something like `MAIN` or `USER`, vs if UserTest returns `TEST`, the Painter messages could signal that they want to skip the `MAIN/USER` adapters but send to all others (and by default no adapters are skipped). The logic could also be reversed where we explicitly list which adapter types are allowed (and by default all adapter types are allowed).

Let me know what you think - happy to file a JIRA as a follow-up if either of these make sense.

PR #364, opened 2022-09-02
Run smoke tests after prod deploy

cat5inthecradle:
Run integration tests against production after each deployment.

Failures are reported to #infra-javabuilder in Slack

Preview Pipeline (Delete when merging):
https://us-east-1.console.aws.amazon.com/codesuite/codepipeline/pipelines/javabuilder-smoketest-cicd/view?region=us-east-1

### DIFF
--- a/cicd/2-cicd/cicd.template.yml
+++ b/cicd/2-cicd/cicd.template.yml
@@ -395,6 +395,28 @@ Resources:
                   Capabilities: CAPABILITY_AUTO_EXPAND
                   RoleArn: !Sub arn:aws:iam::${AWS::AccountId}:role/admin/CloudFormationService
           - !Ref AWS::NoValue
+
+        - !If
+          - DeployForProduction
+          - Name: Smoke_Test
+            Actions:
+              - Name: smoke-test
+                ActionTypeId:
+                  Category: Build
+                  Owner: AWS
+                  Provider: CodeBuild
+                  Version: 1
+                InputArtifacts:
+                  - Name: sourceCode
+                Configuration:
+                  ProjectName: !Ref IntegrationTestBuildProject
+                  EnvironmentVariables: !Sub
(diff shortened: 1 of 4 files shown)


cicd/2-cicd/deploy-cicd.sh
```diff
@@ -31,7 +31,7 @@ else
   fi
 fi
 
-ENVIRONMENT_TYPE=${ENVIRONMENT_TYPE-'standard'}
+ENVIRONMENT_TYPE=${ENVIRONMENT_TYPE-'production'}
```
molly-moen, line 34: should the default be production?

molly-moen (approved):

cicd/2-cicd/deploy-cicd.sh
molly-moen: Realized this was confusing. Wondering if "adhoc" or "test" would make more sense as the default vs production

cicd/2-cicd/cicd.template.yml
```diff
@@ -395,6 +395,28 @@ Resources:
...
+              - Name: smoke-test
+                ActionTypeId:
+                  Category: Build
+                  Owner: AWS
+                  Provider: CodeBuild
+                  Version: 1
+                InputArtifacts:
+                  - Name: sourceCode
+                Configuration:
+                  ProjectName: !Ref IntegrationTestBuildProject
```
sanchitmalhotra126, line 412: Just want to make sure I understand, how are these different from integration tests?

cicd/2-cicd/deploy-cicd.sh
cat5inthecradle: We could also require it to be specified, and not provide a default. I'd like to tackle that change in a separate PR though.

cicd/2-cicd/cicd.template.yml
cat5inthecradle: The only difference is the environment variables for the domain/subdomain.

cicd/2-cicd/deploy-cicd.sh
molly-moen: makes sense, I'm fine with this for now

PR #365, opened 2022-09-06
Add reflection access for validation

molly-moen:
Levelbuilders would like to be able to validate some student code via reflection. For example, the only way to check the value of a private field is via reflection on that field. This PR adds access to the reflection package to validation code only.

Since I saw memory issues with a previously added package (mockito), in this PR I also added a memory logging function we can use when testing a new package to ensure there are no leaks. It is not used in this PR but having the function available makes adding a new package easier.

I verified that this package does not seem to have any memory leaks with our setup.

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeExecutionManager.java
+++ b/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeExecutionManager.java
@@ -6,9 +6,6 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.PrintStream;
-import java.lang.management.ClassLoadingMXBean;
-import java.lang.management.ManagementFactory;
-import java.lang.management.MemoryPoolMXBean;
 import java.nio.file.Files;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/CodeExecutionManager.java
```diff
@@ -170,4 +176,25 @@ private void onPostExecute() {
       this.isInitialized = false;
     }
   }
+
+  private void logMemoryUsage() {
```
sanchitmalhotra126, line 180: Nice! This seems like a helpful profiling tool - could we move this into a util in case we want to use it elsewhere?

molly-moen: good idea, will update

molly-moen: I moved it to a new util and removed the constant and method call here since it made less sense to call it now that it's in a separate util.

sanchitmalhotra126 (approved):
LGTM!

PR #366, opened 2022-09-06
Add ValidationHelper class

molly-moen:
This PR adds a new `ValidationHelper` class that is accessible to levelbuilder validation code. This class has a single method as of now, `getClassNames`, which returns a list of all the class names as strings in the user's project. This will help levelbuilders write tests when there could be multiple different options for class names for a level.

I gave this api a generic name so that we can add to it if we need more validation helper methods in the future. 

This api is documented in [this doc](https://docs.google.com/document/d/1t_NEkBEhya70Vl17Ra07Tq3Wk1xYdpU8TYNagGiuOFM/edit#bookmark=id.ozesjpldtb2k)

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/ValidationRunner.java
+++ b/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/ValidationRunner.java
@@ -2,8 +2,8 @@ package org.code.javabuilder;
 
 import java.lang.reflect.Method;
 import java.net.URLClassLoader;
-import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
 import org.code.javabuilder.util.ProjectLoadUtils;
 import org.code.protocol.*;
 import org.code.validation.support.NeighborhoodTracker;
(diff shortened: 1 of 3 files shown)


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/ValidationRunner.java
```diff
@@ -47,9 +48,14 @@ public boolean run(URLClassLoader urlClassLoader)
 
   private void setUpForValidation(URLClassLoader urlClassLoader) throws UserInitiatedException {
     Method mainMethod = ProjectLoadUtils.findMainMethod(urlClassLoader, this.projectFiles);
+    List<String> classNames = new ArrayList<>();
+    for (JavaProjectFile projectFile : this.projectFiles) {
+      classNames.add(projectFile.getClassName());
+    }
```
sanchitmalhotra126, line 54: nit: could replace with `stream`/`map`

sanchitmalhotra126 (approved):

PR #367, opened 2022-09-07
Shutdown lambda if container ID in table

sanchitmalhotra126:
Adds the logic to check the unhealthy_containers DynamoDB table, and shut down the JVM if the container ID is present. We check during initialization and cleanup, and the desired shutdown trigger (start or end) is part of the DynamoDB key.

JIRA: https://codedotorg.atlassian.net/browse/JAVA-674

Tested on adhoc + unit.

### DIFF
--- a/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
+++ b/org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
@@ -10,6 +10,8 @@ import com.amazonaws.services.apigatewaymanagementapi.AmazonApiGatewayManagement
 import com.amazonaws.services.apigatewaymanagementapi.model.DeleteConnectionRequest;
 import com.amazonaws.services.apigatewaymanagementapi.model.GetConnectionRequest;
 import com.amazonaws.services.apigatewaymanagementapi.model.GoneException;
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDB;
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDBClientBuilder;
 import com.amazonaws.services.lambda.runtime.Context;
 import com.amazonaws.services.lambda.runtime.RequestHandler;
 import com.amazonaws.services.s3.AmazonS3;
@@ -26,6 +28,7 @@ import java.util.Map;
 import java.util.UUID;
 import java.util.logging.Handler;
 import java.util.logging.Logger;
+import org.code.javabuilder.UnhealthyContainerChecker.ShutdownTrigger;
 import org.code.javabuilder.util.LambdaUtils;
 import org.code.protocol.*;
 import org.code.validation.support.UserTestOutputAdapter;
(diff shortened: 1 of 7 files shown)


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/UnhealthyContainerChecker.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  public boolean shouldForceShutdownContainer(String containerId, ShutdownTrigger trigger) {
+    // The container ID value is a concatenation of the ID and the shutdown trigger type
+    final String containerIdCompositeValue = containerId + "#" + trigger.getName();
+    final Map<String, AttributeValue> key =
+        Map.of(CONTAINER_ID_KEY_NAME, new AttributeValue(containerIdCompositeValue));
+    final Map<String, AttributeValue> entry;
+    try {
+      entry = this.dynamoDBClient.getItem(this.tableName, key).getItem();
```
molly-moen, line 48: is there any risk of the dynamoDB client getting into a bad state (like the api gateway client sometimes does) and therefore throwing an exception here? I wonder if for now we should log a severe error instead of a warning so we know if this happens and can build in some checks so we don't cause user's programs to fail if the client is shut down.

sanchitmalhotra126: Hmm good question. I don't think there should be because it just queries the DB directly? But of course we have seen that new AWS-related errors do crop up from time to time so it would be good to be safe. For example I know if there is a permissions issue with the request, then this will throw. I can change this to a severe error. 
Re: "so we don't cause user's programs to fail if the client is shut down" - just to clarify, are you thinking there other checks we should make here? Currently any exception here is caught and we just return false so the container is not shut down, but will proceed normally otherwise.

molly-moen: oh my bad I thought if false we would kill the program. In that case either warning or severe is fine. We probably should check our warnings on occasion though, I can update the new dashboard to try to give us a view into them

molly-moen: I was mainly thinking of how sometimes aws was shutting down our api gateway client in terms of potential errors. The s3 client doesn't seem to have that issue though.

sanchitmalhotra126: Yeah agreed, I'd expect this to work more similarly to our S3 client, but it doesn't hurt to log the exception if we encounter it.

molly-moen (approved):
looks good!

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
```diff
@@ -401,4 +419,14 @@ private void verifyApiClient(String connectionId) {
               .build();
     }
   }
+
+  /**
+   * Checks if this container has been marked unhealthy and if so, forces a shutdown via
+   * System.exit().
+   */
+  private void checkContainerHealth(ShutdownTrigger trigger) {
```
jmkulwik, line 427: I got a little confused when I read this. The function name made me think it will be returning information about the container's health, but it's instead conditionally shutting down the container. To fix this, I'd recommend updating the name to something more action-oriented, such as `shutdownIfUnhealthy` or `conditionallyShutdownContainer`.

sanchitmalhotra126: Good call - I'll rename to `shutdownContainerIfUnhealthy`

jmkulwik (approved):
Nice work!

PR #368, opened 2022-09-08
Update hourly user limit and silence classroom limit

molly-moen:
We have seen some users hit or get close to the hourly limit and multiple classrooms hit the classroom limit. Since these limits block users completely from using Javabuilder this PR does the following:
- increase the production hourly user limit to 1000 runs per hour. I chose 1000 because if you constantly ran hello world you could theoretically run it 1440 times per hour. 1000 seems like a limit a reasonable user would not hit.
- Increase the per-classroom limit to 5000 per hour (was previously 1000) and silence that limit. We will still log if a classroom hits the limit but won't block them. We can monitor this for a while and see if it's a reasonable limit or not, or if we should be blocking whole classrooms at all.
- I also increased the daily user limit to 5000, but this is not currently used.


hannahbergam (approved):
🎉Quick turnaround:) It's exciting to think about them hitting the previous limit!

PR #369, opened 2022-09-09
Update warning logs

molly-moen:
We discussed adding a view into warnings in the javabuilder dashboard in [this PR](https://github.com/code-dot-org/javabuilder/pull/367#discussion_r965243906). When I looked into this I found our warnings are hard to parse because they are generally very long strings/stack traces. This PR makes them easier to group in logs by adding a `type` field and either a `detail` field or exception-specific fields such as `stackTrace`. This PR also moves us towards only using `LoggerUtils` to generate logs.

Tested on a dev instance.

### DIFF
--- a/org-code-javabuilder/protocol/src/main/java/org/code/protocol/LoggerUtils.java
+++ b/org-code-javabuilder/protocol/src/main/java/org/code/protocol/LoggerUtils.java
@@ -101,7 +101,7 @@ public class LoggerUtils {
    * Exceptions logged in this way are intended to be informative, but should be removed after a
    * period of time.
    */
-  public static void logTrackingException(Throwable e) {
+  public static void logTrackingExceptionAsWarning(Throwable e) {
     JSONObject eventData = new JSONObject();
     eventData.put(LoggerConstants.TYPE, e.getClass().getName());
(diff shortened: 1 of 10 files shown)


madelynkasula (approved):

org-code-javabuilder/protocol/src/main/java/org/code/protocol/LoggerUtils.java
```diff
@@ -101,17 +103,29 @@ public static void logSevereException(Throwable e) {
    */
   public static void logTrackingException(Throwable e) {
```
sanchitmalhotra126, line 104: Could we go ahead and change this method name to something warning specific? maybe something like `logExceptionAsWarning` or even just another `logWarning`?

sanchitmalhotra126 (commented):
Nice update! Just one comment on a method name

org-code-javabuilder/protocol/src/main/java/org/code/protocol/LoggerUtils.java
molly-moen: good idea, will update

PR #371, opened 2022-09-12
Re-enable classroom limits and update alerts and metrics

molly-moen:
This PR does the following:
- Re-enable the classroom hourly limit but set it much higher., to 25000 requests per hour
- Add a metric that will log the hourly request count every time we query it, so we can see what sort of hourly usage we see in general. 
- Change the classroom blocked alert to alarm every time we see a new classroom blocked, so we can investigate while we are determining a good limit. This will just post to slack.
- Also updates the prod config to be in line with our beta config, which used a lower provisioned and reserved concurrency. We have found that configuration to be sufficient.

### DIFF
--- a/cicd/3-app/javabuilder/template.yml.erb
+++ b/cicd/3-app/javabuilder/template.yml.erb
@@ -1012,12 +1012,11 @@ Resources:
         ComparisonOperator: GreaterThanThreshold
         TreatMissingData: notBreaching
 
-  HighClassroomsBlockedAlarm:
+  ClassroomBlockedAlarm:
     Type: AWS::CloudWatch::Alarm
     Properties:
-        AlarmName: !Sub "${SubDomainName}_high_classrooms_blocked"
-        AlarmDescription: Unusually high number of classrooms being newly blocked by our throttling
-            thresholds.
+        AlarmName: !Sub "${SubDomainName}_classrooms_blocked"
+        AlarmDescription: A classroom was newly blocked by our threshold. Investigate if this is classroom should be blocked.
         ActionsEnabled: true
         OKActions: []
         AlarmActions:
@@ -1032,8 +1031,8 @@ Resources:
         Period: 60
         EvaluationPeriods: 1
         DatapointsToAlarm: 1
-        Threshold: 10
-        ComparisonOperator: GreaterThanThreshold
+        Threshold: 1
+        ComparisonOperator: GreaterThanOrEqualToThreshold
         TreatMissingData: notBreaching
 
   HighUnknownTokensAlarm:
--- a/cicd/3-app/javabuilder/template.yml.erb
+++ b/cicd/3-app/javabuilder/template.yml.erb
@@ -1015,7 +1015,7 @@ Resources:
   ClassroomBlockedAlarm:
     Type: AWS::CloudWatch::Alarm
     Properties:
-        AlarmName: !Sub "${SubDomainName}_classrooms_blocked"
+        AlarmName: !Sub "${SubDomainName}_classroom_blocked"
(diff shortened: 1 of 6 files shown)


cicd/3-app/javabuilder/template.yml.erb
```diff
@@ -1012,12 +1012,11 @@ Resources:
...
         TreatMissingData: notBreaching
 
-  HighClassroomsBlockedAlarm:
+  ClassroomBlockedAlarm:
     Type: AWS::CloudWatch::Alarm
     Properties:
-        AlarmName: !Sub "${SubDomainName}_high_classrooms_blocked"
-        AlarmDescription: Unusually high number of classrooms being newly blocked by our throttling
-            thresholds.
+        AlarmName: !Sub "${SubDomainName}_classrooms_blocked"
```
madelynkasula, line 1018: super nit: "${SubDomainName}_classroom_blocked" (instead of "classrooms")

javabuilder-authorizer/token_validator.rb
```diff
@@ -130,6 +125,10 @@ def teachers_over_hourly_limit?
         puts "teacher_associated_requests query has paginated responses. user_id #{@user_id} teacher_id #{teacher_id}"
       end
 
+      # Log the count of requests per hour to a metric. We are logging this for now so we can determine if we
+      # have set a good threshold for blocking classrooms.
+      @metrics_reporter.log_count_with_value(CLASSROOM_HOURLY_REQUEST_COUNT, response.count)
```
madelynkasula, line 130: should we log any other info here, like the teacher_id and/or user_id?

molly-moen: This metric is meant for graphing in cloud watch and looking for general patterns across all users. It would be extra data we would probably never use to add user and teacher ids, and I'm not sure if it would mess with the graphs to have extra dimensions we always would want to aggregate over. If we want to dig into specific users I think we could query DynamoDB directly to try to find the problem classrooms

madelynkasula (approved):

javabuilder-authorizer/token_validator.rb
jmkulwik: Logging on every request seems like it might be a bit verbose. Would logging at a less frequent interval, say every time a classroom passes mod 1000 still give us the measure we're looking for?

cicd/3-app/javabuilder/template.yml.erb
```diff
@@ -1032,8 +1031,8 @@ Resources:
         Period: 60
         EvaluationPeriods: 1
         DatapointsToAlarm: 1
-        Threshold: 10
-        ComparisonOperator: GreaterThanThreshold
+        Threshold: 1
+        ComparisonOperator: GreaterThanOrEqualToThreshold
```
sanchitmalhotra126, line 1035: Will this re-alarm if a second classroom is blocked in the same period after the first classroom was blocked? I don't think it's very likely, but just want to make sure we don't miss unblocking the second classroom

javabuilder-authorizer/token_validator.rb
molly-moen: this is logging a metric which is meant to be verbose to my understanding. We could log less often but I thought it would be helpful to see the kind of usage we are getting, we can turn it off after a month or so. We already log multiple metrics on every run (performance times, lambda itself logs things like invocations counts).

cicd/3-app/javabuilder/template.yml.erb
molly-moen: it won't, but it would give us a number of how many were blocked during the evaluation period so we will know if more than one classroom was blocked.

sanchitmalhotra126 (approved):

PR #372, opened 2022-09-13
Update error alarms

molly-moen:
Our previous configuration for error and severe error alarms was:
- Error: alert if error rate was > 25% every 5 minutes for 20 minutes
- Severe error: alert if error rate was > 25% every 5 minutes for 20 minutes

The only minimum number of invocations for these alarms was needing to have at least one invocation every 5 minutes. This meant we occasionally got alarms when there was very low usage and therefore the rate was not significant. The error rates were also too low to page someone. This PR makes the following changes:

- All error rate alarms now have a minimum number of invocations of 100 every 5 minutes for 20 minutes. This means the error rate alarms are also now composite alarms, because AWS doesn't let you create an alarm based on two different metrics without using a composite alarm.
- There are now 2 severe error and 2 error rate alarms. The severe error rates are 10% and 90% rates, and the error rates are 25% and 90% rates.
- For now, all alarms still go to slack instead of paging. Once they've been up and running for a few days and seem like they are working correctly I will change the 90% alarms to page the dotd and send out an update to developers about this new alarm.


cicd/3-app/javabuilder/template.yml.erb
```diff
@@ -669,20 +669,21 @@ Resources:
...
-      AlarmDescription: Send page if Javabuilder severe error rate exceeds 10% for 20
-          minutes. Occasional spikes are expected, but a sustained high error rate
-          is an indication of an outage.
-      ActionsEnabled: true
-      AlarmActions:
-          - !If [SilenceAlertsCondition, !Ref AWS::NoValue, !Sub "arn:aws:sns:${AWS::Region}:${AWS::AccountId}:Javabuilder-high-error-rate"]
+      AlarmName: !Sub "${SubDomainName}_<%=name.downcase%>_<%=config[:AlarmName]%>"
+      AlarmDescription: Severe error rate in Javabuilder's <%=name%> build and run lambda (the core of
+          Javabuilder, which executes student <%=name%> code) exceeded <%=config[:Threshold]%>% for four
+          consecutive 5 minute periods.
```
sanchitmalhotra126, line 682: Could we also add a bit of clarification here (or in the composite alarm definition below) for the difference between the severe error rate and non-severe? Could just be that severe errors are thrown by our code vs non-severe errors are thrown by the lambda system (ex. timeouts)

cicd/README.md
```diff
@@ -67,3 +67,12 @@ By setting the `TARGET_BRANCH` you can create a new CI/CD pipeline that watches
 ```
 TARGET_BRANCH=mybranch cicd/2-cicd/deploy-cicd.sh
 ```
+
+## Debugging `template.yml.erb`
```
sanchitmalhotra126, line 71: Nice tip 😎

cicd/3-app/javabuilder/template.yml.erb
```diff
@@ -893,6 +899,107 @@ Resources:
...
+            - Name: FunctionName
+              Value: !Ref BuildAndRunJava<%=name%>ProjectFunction
+        Period: 300
+        EvaluationPeriods: 4
+        DatapointsToAlarm: 4
+        Threshold: 100
+        ComparisonOperator: GreaterThanOrEqualToThreshold
+        TreatMissingData: notBreaching
+  
+  <%=name%>ElevatedSevereErrorRateAlarm:
```
sanchitmalhotra126, line 923: To clarify - maybe I'm misunderstanding, but should these names be flipped? Looks like this alarms on the 10% threshold vs the one below alarms on the 90%. I figured "Elevated" would refer to the 90% alarm?

sanchitmalhotra126 (commented):
Nice! Neat workaround with the composite alarm. Just a couple questions

cicd/3-app/javabuilder/template.yml.erb
molly-moen: will do

cicd/3-app/javabuilder/template.yml.erb
molly-moen: Naming struggles here. I wanted to use "High" for the 90% alarms, but I couldn't re-use the name we had before because it doesn't let you change the type of an object in cloudformation (in this case, changing from alarm to composite alarm). Your suggestion makes sense, if we go with "Elevated" for the 90% alarm does it make sense for the 10/25% threshold to just be "error rate" or is there another adjective we could use?

molly-moen: I swapped the names around for now, I think that makes more sense.

sanchitmalhotra126 (approved):

PR #373, opened 2022-09-14
remove unnecessary policy condition

cat5inthecradle:


cat5inthecradle:
I don't think this fixes the notification issue, but it cleans it up a bit.

molly-moen (approved):

PR #375, opened 2022-09-21
Update websocket logs

molly-moen:
Two changes to our websocket logs:
- no longer do verbose logging (`DataTraceEnabled` setting)
- log when we are returning `invalid mini-app`, because that previously was not logged anywhere


api-gateway-routes/api_gateway_proxy_function.rb
```diff
@@ -80,6 +80,13 @@ def on_connect(event, context)
   elsif authorizer['mini_app_type'] == 'theater'
     function_name = ENV['BUILD_AND_RUN_THEATER_PROJECT_LAMBDA_ARN']
   else
+    # log so we know we saw an invalid mini app
+    log_data = {
+      connectionId: request_context["connectionId"], 
+      statusCode: 400, 
+      reason: "invalid mini-app"
+    }.to_json
+    puts log_data
```
madelynkasula, line 89: do we want to use `puts` here? should we use `p` instead to pretty-print the object?

api-gateway-routes/api_gateway_proxy_function.rb
```diff
@@ -80,6 +80,13 @@ def on_connect(event, context)
   elsif authorizer['mini_app_type'] == 'theater'
     function_name = ENV['BUILD_AND_RUN_THEATER_PROJECT_LAMBDA_ARN']
   else
+    # log so we know we saw an invalid mini app
+    log_data = {
+      connectionId: request_context["connectionId"],
```
sanchitmalhotra126, line 85: Maybe worth pulling this out into its own function and using it in other spots where we log (lines 40 and 45) so the log format is consistent?

molly-moen: good idea

api-gateway-routes/api_gateway_proxy_function.rb
madelynkasula: oh wait, we might have to use `puts log_data.inspect` instead. if we're running this in rails, we can use `p`, otherwise we have to use `puts x.inspect`

molly-moen: yeah I actually played around with this and `puts [some object].to_json` gives us the best output in CloudWatch logs. It looks the most like real json and CloudWatch formats it nicely when you expand the log (new lines between each key)

api-gateway-routes/api_gateway_proxy_function.rb
molly-moen: updated all the `puts` in this function to use the new formatting

madelynkasula (approved):

sanchitmalhotra126 (approved):

PR #378, opened 2022-09-22
Refactor Logger.getLogger references to use LoggerUtils

madelynkasula:
[JAVA-664](https://codedotorg.atlassian.net/browse/JAVA-664) - Replace remaining `.info`, `.warning`, and `.severe` log calls that were using `Logger.getLogger` to use `LoggerUtils` instead.


org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
```diff
@@ -113,27 +110,22 @@ public String handleRequest(Map<String, String> lambdaInput, Context context) {
     JavabuilderContext.getInstance().destroyAndReset();
     this.trackStartupPerformance();
 
-    // TODO: Because we reference the logger object throughout the codebase via
-    // Logger.getLogger(MAIN_LOGGER), we need to set it up in the same scope as code execution to
-    // prevent the instance from being garbage collected. Instead, we should store a global
-    // reference to this logger object and access that directly rather than via
-    // Logger.getLogger(MAIN_LOGGER)
```
madelynkasula, line 120: @sanchitmalhotra126 just double-checking -- am i handling this TODO appropriately with my change below?

org-code-javabuilder/validation/src/main/java/org/code/validation/support/NeighborhoodTracker.java
```diff
@@ -58,11 +54,12 @@ public void trackEvent(ClientMessage message) {
...
-          .severe(
-              String.format(
-                  "Error: received a painter event for an uninitialized painter: %s, %s\n",
-                  message.getValue(), message.getDetail()));
+      String loggingString =
+          String.format(
+              "Error: received a painter event for an uninitialized painter: %s, %s\n",
+              message.getValue(), message.getDetail());
+      LoggerUtils.logSevereError(
+          (JavabuilderThrowableMessage) message, loggingString, new Throwable());
```
madelynkasula, line 62: i think i might need some java guidance on this one... originally, our severe-level error was just logging a string error. however, our current utils for severe errors all require a `Throwable cause` in some way:

https://github.com/code-dot-org/javabuilder/blob/main/org-code-javabuilder/protocol/src/main/java/org/code/protocol/LoggerUtils.java#L66-L98

a few questions about the code i have here:
1. i'm casting `message`, which is a `ClientMessage` to `JavabuilderThrowableMessage`, which extends `ClientMessage` -- is that appropriate here? (i'm a type-casting noob 😄)
2. i kept the original `loggingString` -- should it change now that we have other data being logged in our error?
3. is it okay that the `cause` will be `null`?

alternatively, i could add another util function that logs a severe error that just accepts a string 🤷🏻‍♀️

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
sanchitmalhotra126: Ah, this is my bad sorry, I definitely should have been more explicit about this task :) we don't actually need (or want to) move any `addHandler/setUseParentHandlers/removeHandler` code into `LoggerUtils` for this task (just yet). For now, we just want to replace any logging methods (e.g. `.info`, `.warning`, `.severe`) with corresponding methods on `LoggerUtils`. So we shouldn't need to change much in LambdaRequestHandler, WebSocketServer, or AWSSystemExitHelper. Changing those logging methods to use `LoggerUtils` will help us minimize the surface area where `Logger.getLogger(MAIN_LOGGER)` is called, so we can then deal with the garbage collection issue in a separate task (details below).

Context: Logger.getLogger returns a logger object from the logging system, but has a quirk that it may be garbage collected at any point "if a strong reference to the Logger is not kept" ([docs](https://docs.oracle.com/javase/7/docs/api/java/util/logging/Logger.html)). So hypothetically, two calls to `Logger.getLogger(MAIN_LOGGER)` may return different objects if the first is garbage collected before the second call. We just _happen_ to get around this today because store a reference to the Logger object returned from Logger.getLogger(MAIN_LOGGER) in this method at line 126. This method (handleRequest) is the entrypoint into the lambda system and wraps all code execution, so by storing that reference, we keep it around for the entire session. Therefore any calls to Logger.getLogger(MAIN_LOGGER) do end up returning the same reference. However, that doesn't feel great :) it just kind of happens to magically work and can be easily broken if the code is moved. So the follow-up here that I have mind, once we've moved all the logging methods to LoggerUtils, is to move the add/remove handler logic into another unit that can also handle storing the strong reference to the logger object.

org-code-javabuilder/validation/src/main/java/org/code/validation/support/NeighborhoodTracker.java
sanchitmalhotra126: Hmm, since LoggerUtils internally converts this data to a string, I think adding a new util method that logs a string is fine? But I'll also defer to @molly-moen on what would be the most helpful format here. Another option is to wrap the message in an `Exception` and just pass that in:
```
Exception e = new Exception(loggingString);
LoggerUtils.logSevereException(e);
```
Also another question for Molly...LoggerUtils.logSevere will publish a severe error metric. Should we be publishing errors for this case?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
madelynkasula: ahh, okay -- that's no problem! easy enough 😄

madelynkasula: although, now i have a new question -- if we aren't moving any handler helpers to LoggerUtils, i don't think there's any action i can take on [JAVA-680](https://codedotorg.atlassian.net/browse/JAVA-680) because `ExceptionHandlerTest` only uses `Logger.getLogger` to call those handler helpers:

https://github.com/code-dot-org/javabuilder/blob/1a0befa265ae992f810e130bbfefb1ce17470388/org-code-javabuilder/lib/src/test/java/org/code/javabuilder/ExceptionHandlerTest.java#L36-L44

so that test refactor is blocked on creating the unit that stores the strong reference to the logger, right? is that a separate task?

sanchitmalhotra126: Right yeah, so moving Logger.getLogger methods to LoggerUtils was a prereq for deflaking this test, but looks like all usages in `ExceptionHandler` had already been converted prior to this PR :) the fix to deflake the test would just be to use static mocking to mock static methods on `LoggerUtils` (so we're not dealing with the internals of LoggerUtils calling Logger.getLogger), but that can be handled as a separate task.

org-code-javabuilder/validation/src/main/java/org/code/validation/support/NeighborhoodTracker.java
molly-moen: For the formatting questions: I think for this case it would be most helpful to add a new severe error helper that doesn't require an exception, but instead gets passed a String `cause` and a String `message`. In our log parsing for the dashboard we use `cause` to aggregate errors and see patterns, so having a short `cause` for this case would be helpful. Then we can put any additional information into `message`.

molly-moen: @sanchitmalhotra126 for this question: "LoggerUtils.logSevere will publish a severe error metric. Should we be publishing errors for this case?" Can we publish an error? My understanding is errors are an internal Lambda thing. Since this is coming from us I think it makes sense to be a severe error. This is sort of a weird one because the only way this would happen is if we had a bug in the painter code here which I don't think we do.

sanchitmalhotra126: Ah sorry my question was more around whether this particular situation warranted us publishing a severe error metric at all. Do we want to be tracking this/trigger an alarm for this? Agreed that it's highly unlikely, so it's probably fine.

molly-moen: ah gotcha. I think at this point we could just get rid of this and ignore any unexpected events here. This should not ever happen.

madelynkasula: sorry, not sure i understand -- do we still want to log an error and publish a metric? or just log the error?

org-code-javabuilder/lib/src/main/java/org/code/javabuilder/LambdaRequestHandler.java
madelynkasula: ahhhh, okay, cool -- thank you! since that's the only action for JAVA-680 i'll just open a separate PR for that

org-code-javabuilder/validation/src/main/java/org/code/validation/support/NeighborhoodTracker.java
sanchitmalhotra126: We can just get rid of this log statement entirely, and just early return (right @molly-moen?)

molly-moen: ah sorry I missed this message, yes we can just get rid of it and early return

molly-moen (approved):

PR #379, opened 2022-10-06
Fix flaky ExceptionHandlerTest

molly-moen:
We were occasionally seeing this test fail because the Logger instance was being garbage collected. The fix for this is to mock `LoggerUtils` in this test instead.


madelynkasula (approved):

sanchitmalhotra126 (approved):
Thanks for getting this!

PR #382, opened 2023-02-03
fix case of subdomain term

cat5inthecradle:
I'm nearly certain this variable rename would be a no-op on the actual template.


beta-template.yml.erb
```diff
@@ -4,13 +4,13 @@ Description: Provision an instance of the Javabuilder service. Empty the Content
...
-    Description: Base domain name.
+    Description: Base domain name (e.g. 'code.org' in 'javabuilder.code.org').
   BaseDomainNameHostedZonedID:
     Type: String
     Description: AWS Route53 Hosted Zone ID for base domain name.
-  SubDomainName:
+  SubdomainName:
     Type: String
-    Description: Sub domain name for javabuilder service.
+    Description: Sub domain name for javabuilder service (e.g. 'javabuilder' in 'javabuilder.code.org').
```
Hamms, line 13: nit:

```suggestion
    Description: Subdomain name for javabuilder service (e.g. 'javabuilder' in 'javabuilder.code.org').
```

Looks like the corresponding line in `3-app` already got updated!

Hamms (approved):
LGTM; thanks for picking this up so quickly!

As an aside, there is also an argument to be made that the phrase "Base Domain Name" is a bit confusing. Technically, something like `code.org` is just called a "base domain", and consists of two parts: the "domain name" (`code`) and the "top-level domain" (`org`). We're kind of smashing together the two different-but-related concepts of "base domain" and "domain name" to produce a phrase which I think we consistently use to refer to just one of those concepts. Either way, though, that's a discussion for another PR.

cat5inthecradle:
> As an aside, there is also an argument to be made that the phrase "Base Domain Name" is a bit confusing. Technically, something like `code.org` is just called a "base domain", and consists of two parts: the "domain name" (`code`) and the "top-level domain" (`org`). We're kind of smashing together the two different-but-related concepts of "base domain" and "domain name" to produce a phrase which I think we consistently use to refer to just one of those concepts. Either way, though, that's a discussion for another PR.

I don't know what the best term is here for domain + tld, I avoided `DomainName` to make it more clear that it wouldn't include a subdomain. We're using `BaseDomain` in three ways:

1. To identify when we're using 'dev-code.org' (these could be refactored away I think)
2. To identify the hosted zone (nameserver) for new dns records (could be refactored to use hostedzoneID, which we already require)
3. To assist with the creation of certain urls like `subdomain-http.domain.tld` and `subdomain-content.domain.tld` (these could be refactored to `service-subdomain.domain.tld` or `service.subdomain.domain.tld`

Hamms:
Best term I'd say for `domain + tld` is just `BaseDomain`. In general:

```
TopLevelDomain = 'org'
DomainName = 'code'
Subdomains = ['studio']
BaseDomain = [DomainName, TopLevelDomain].join('.')
Hostname = [*Subdomains, DomainName].join('.')
```

sureshc (approved):
nit - feature branch name is way too hilarious

cat5inthecradle:
discussion of whether we need to maintain log/alarm continuity: https://codedotorg.slack.com/archives/C03DBDN67B7/p1675465733136349

PR #388, opened 2023-04-13
Reduce concurrency for demo mode

ebeastlake:

PR #390, opened 2023-04-13
Reduce demo concurrency - step 1

cat5inthecradle:
reduce costs in demo environment
step 1 of a two step process, correctly completes #388
next: reduce reserved concurrency via #391

### DIFF
--- a/cicd/3-app/javabuilder/config/production-demo.config.json
+++ b/cicd/3-app/javabuilder/config/production-demo.config.json
@@ -4,7 +4,7 @@
     "SubdomainName": "javabuilder-demo",
     "BaseDomainNameHostedZonedID": "Z2LCOI49SCXUGU",
     "ProvisionedConcurrentExecutions": "1",
-    "ReservedConcurrentExecutions": "5",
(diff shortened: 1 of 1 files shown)


cicd/3-app/javabuilder/config/production-demo.config.json
```diff
@@ -4,7 +4,7 @@
     "SubdomainName": "javabuilder-demo",
     "BaseDomainNameHostedZonedID": "Z2LCOI49SCXUGU",
     "ProvisionedConcurrentExecutions": "1",
-    "ReservedConcurrentExecutions": "5",
+    "ReservedConcurrentExecutions": "500",
```
molly-moen, line 7: Change name of pr to match that we are changing reserved?

molly-moen (approved):

cicd/3-app/javabuilder/config/production-demo.config.json
cat5inthecradle: oh, right. I had the overall diff in my head (from what's currently deployed)

PR #391, opened 2023-04-13
Reduce demo concurrency - step 2

cat5inthecradle:
Reducing concurrency to save costs.

This step 2 of a two step change, follows #390  to correctly apply #388


ebeastlake (approved):

cat5inthecradle:
This PR wasn't targeting 'main' so it didn't trigger codebuild initially. Added an empty commit after changing the base to kick that off.

PR #392, opened 2023-04-27
Add temporary per-day user throttling

molly-moen:
As part of [Java Lab Eval Mode](https://docs.google.com/document/d/1YFFCcTooSb2t5Q9pmNCOxLtujwKTM4SXTuUtj5Oma7Q/edit#bookmark=id.hv46zl6gv16), we want to add back in daily user throttling, but have it not block the user forever, only until they are back under the limit. In this PR I:
- Converted our turned-off daily throttling to be this temporary model, and added a new key `USER_BLOCKED_TEMPORARY` that we will send if the user reaches their limit in a day.
- Updated the near limit message to include the interval and whether this is a permanent or temporary block.
- Updated our daily and hourly limits such that if the limit is 0, there is no limit. We need this because we don't want hourly limits on javabuilder-demo, and we don't want a daily limit on production javabuilder (for now).

This PR cannot be merged until the [corresponding code.org](https://github.com/code-dot-org/code-dot-org/pull/51570) PR has been merged and deployed.

### DIFF
--- a/javabuilder-authorizer/token_validator.rb
+++ b/javabuilder-authorizer/token_validator.rb
@@ -11,7 +11,7 @@ class TokenValidator
   USER_REQUEST_RECORD_TTL_SECONDS = 25 * ONE_HOUR_SECONDS
   TEACHER_ASSOCIATED_REQUEST_TTL_SECONDS = 25 * ONE_HOUR_SECONDS
   NEAR_LIMIT_BUFFER = 10
-  NO_LIMIT = 0
+  NO_LIMIT = -1
 
   def initialize(payload, origin, context)
     @token_id = payload['sid']
@@ -241,7 +241,7 @@ class TokenValidator
   end
 
   def get_user_near_limit_detail(count, limit, period, lockout_type)
-    if limit > 0 && count <= limit && count >= (limit - NEAR_LIMIT_BUFFER)
+    if limit != NO_LIMIT && count <= limit && count >= (limit - NEAR_LIMIT_BUFFER)
       return {remaining: limit - count, period: period, lockout_type: lockout_type}
     else
       return nil
--- a/javabuilder-authorizer/token_validator.rb
+++ b/javabuilder-authorizer/token_validator.rb
@@ -11,6 +11,7 @@ class TokenValidator
   USER_REQUEST_RECORD_TTL_SECONDS = 25 * ONE_HOUR_SECONDS
   TEACHER_ASSOCIATED_REQUEST_TTL_SECONDS = 25 * ONE_HOUR_SECONDS
   NEAR_LIMIT_BUFFER = 10
+  NO_LIMIT = 0
 
   def initialize(payload, origin, context)
     @token_id = payload['sid']
@@ -93,17 +94,13 @@ class TokenValidator
 
   def user_over_hourly_limit?(hourly_usage_response)
     limit_per_hour =  ENV['limit_per_hour'].to_i
-    # A limit of 0 means there is no limit.
-    if limit_per_hour > 0
-      user_over_limit?(
-        hourly_usage_response,
-        limit_per_hour,
-        USER_OVER_HOURLY_LIMIT,
-        true # Should block permanently if the limit has been reached.
-      )
-    else
-      false
-    end
+    return false if limit_per_hour == NO_LIMIT
+    user_over_limit?(
+      hourly_usage_response,
+      limit_per_hour,
+      USER_OVER_HOURLY_LIMIT,
+      true # Should block permanently if the limit has been reached.
+    )
   end
 
   def get_user_near_hourly_limit_detail(hourly_usage_count)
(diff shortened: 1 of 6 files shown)


javabuilder-authorizer/token_validator.rb
```diff
@@ -88,24 +92,41 @@ def teachers_blocked?
...
 
   def user_over_hourly_limit?(hourly_usage_response)
-    user_over_limit?(
-      hourly_usage_response,
-      ENV['limit_per_hour'].to_i,
-      USER_OVER_HOURLY_LIMIT
-    )
+    limit_per_hour =  ENV['limit_per_hour'].to_i
+    # A limit of 0 means there is no limit.
+    if limit_per_hour > 0
```
sanchitmalhotra126, line 97: nit: early return false if limit_per_hour == 0?

javabuilder-authorizer/token_validator.rb
```diff
@@ -29,10 +29,14 @@ def validate
...
     return error(USER_BLOCKED) if user_over_hourly_limit?(hourly_usage_response)
-    # return error(USER_BLOCKED) if user_over_daily_limit?
+    daily_usage_response = user_usage(ONE_DAY_SECONDS)
+    return error(USER_BLOCKED_TEMPORARY) if user_over_daily_limit?(daily_usage_response)
     return error(CLASSROOM_BLOCKED) if teachers_over_hourly_limit?
 
     near_limit_detail = get_user_near_hourly_limit_detail(hourly_usage_response.count)
+    if !near_limit_detail
+      near_limit_detail = get_user_near_daily_limit_detail(daily_usage_response.count)
+    end
```
sanchitmalhotra126, line 39: With the way our limits set up, seems like there's no way for a user to be near both an hourly limit and a daily limit - but if there was, I guess we're prioritizing the hourly limit warning? Seems fine to me, just wanted to clarify

javabuilder-authorizer/token_validator.rb
```diff
@@ -226,17 +247,17 @@ def user_usage(time_range_seconds)
     response
   end
 
-  def get_user_near_limit_detail(count, limit)
-    if count <= limit && count >= (limit - NEAR_LIMIT_BUFFER)
-      return {remaining: limit - count}
+  def get_user_near_limit_detail(count, limit, period, lockout_type)
+    if limit > 0 && count <= limit && count >= (limit - NEAR_LIMIT_BUFFER)
```
sanchitmalhotra126, line 251: minor: if we want to be super clear, maybe we can save 0 as a constant like "NO_LIMIT" or something, and check `limit != NO_LIMIT` here? The 0 is documented well enough in the template so maybe not necessary

sanchitmalhotra126 (approved):
LGTM!

javabuilder-authorizer/token_validator.rb
cat5inthecradle: Would a downside of that be a user receiving a "hourly limit" warning message, so they come back in an hour only to receive the "daily limit" message on their first attempt? Only matters if the messages are unique.

cat5inthecradle: Though, the hourly limit triggers a permanent block, so it does make sense that it should go first.

cat5inthecradle (approved):
Template and config changes look good.

javabuilder-authorizer/token_validator.rb
molly-moen: Good questions. Yes, in our prod environments only one of these will be set (demo has a daily limit, prod has an hourly). But if we want both, my presumption was the hourly would be much lower than the daily and therefore the user would hit that first. And the fact that hourly triggers the permanent block means it is more important to message. The messages to the user are unique--they say what the time frame was (hour or day) and if the block is permanent.

molly-moen: It feels like overkill for now to make some combined message given we won't show it to users, but it could be a follow-up if we think we'll enable both limits at some point.

javabuilder-authorizer/token_validator.rb
molly-moen: since this is slightly hacky, it doesn't hurt to be super clear. I'll update this

javabuilder-authorizer/token_validator.rb
```diff
@@ -11,6 +11,7 @@ class TokenValidator
   USER_REQUEST_RECORD_TTL_SECONDS = 25 * ONE_HOUR_SECONDS
   TEACHER_ASSOCIATED_REQUEST_TTL_SECONDS = 25 * ONE_HOUR_SECONDS
   NEAR_LIMIT_BUFFER = 10
+  NO_LIMIT = 0
```
jmkulwik, line 14: Are there any cases where we would set the limit to zero intending to "shut off" access? i.e. I think our guidance in a DDOS is to set reserved concurrency to zero. Any reason we'd do that here too?

molly-moen: I think if we set reserved concurrency to 0 that would cover shutting off access. If we want to severely limit access we could set limits to 1. For hourly limit that would probably be more trouble than it's worth, because we would have to then go unblock everyone.

jmkulwik: Just had a thought (not blocking). Does this take negative numbers? -1 seems like a pretty common "no limit" key that's less likely to have us run into issues in the future.

cat5inthecradle: That makes sense too (using -1), but if we used that, we'd probably want to make 0 have some predictable defined behavior (automatic rejection with a 'limit exceed' error).

molly-moen: I'll try it out, the easiest way to try it out is by committing to auto-deploy to my dev instance, so there may be a little churn on this pr in the meantime.

molly-moen: -1 works and I agree it's a bit cleaner, so let's go with that!

PR #393, opened 2023-05-24
Updates for javabuilder-demo to go live

molly-moen:
- Increase provisioned concurrency to 5
- Increase reserved concurrency to 50
- Update our "high concurrency" alarm to have a different threshold and action for demo. In demo, we will alarm if we reach 45 concurrent executions for 10 minutes. This alarm will go to `#infra-javabuilder` rather than page the DOTD. I created a new [SNS topic](https://us-east-1.console.aws.amazon.com/sns/v3/home?region=us-east-1#/topic/arn:aws:sns:us-east-1:475661607190:javabuilder-low-urgency) for this.

I wanted to update the alarm to be based on a percentage of reserved concurrency, but I couldn't figure out a way to do that in a combination of CloudWatch and CloudFormation.


.ruby-version
```diff
@@ -1 +1 @@
-2.7.7
+2.7.8
```
molly-moen, line 1: This change was needed for the ci/cd pipeline to build correctly for a dev instance, so I think it will be needed for prod as well. We've done [similar changes](https://github.com/code-dot-org/javabuilder/commit/8106ac2e6d1880a7b270f2a4b78e10f915b0b694) in the past to keep up the the AWS version of ruby.

ebeastlake (approved):
This seems reasonable to me. I defer to @cat5inthecradle for the final word on the threshold numbers!

cicd/3-app/javabuilder/template.yml.erb
```diff
@@ -63,6 +63,7 @@ Globals:
     Tracing: Active
 Conditions:
   IsDevCondition: !Equals [!Ref BaseDomainName, "dev-code.org"]
+  IsDemoCondition: !Equals [!Ref SubdomainName, "javabuilder-demo"]
```
cat5inthecradle, line 66: I see you continued the existing pattern here, but we should try to keep this template environment-agnostic, and I would much rather refactor the design to allow us to eliminate the `IsDevCondition` rather than add a new `IsEnvironmentCondition`

You're using `IsDemoCondition` in two places, so I'll make comments directly there.

cicd/3-app/javabuilder/template.yml.erb
```diff
@@ -857,18 +858,19 @@ Resources:
...
+          Alarm if javabuilder usage has high concurrent executions for 10 minutes.
+          Occasional spikes are expected, but long-running high usage is an indication
+          of an attack. If this is occuring on the demo environment, this is a non-urgent
+          issue as we expect occasional periods of high usage. If it is on production, 
+          page the student learning team for further investigation. See this doc for investigation steps
           https://docs.google.com/document/d/1bHvV6pvUcwxgZpw0YWBmxFggQL5KqYx9zwolwkZhjU8/edit#bookmark=id.xs1gcuxrw6ze
       ActionsEnabled: true
       AlarmActions:
-          - !If [SilenceAlertsCondition, !Ref AWS::NoValue, !Sub "arn:aws:sns:${AWS::Region}:${AWS::AccountId}:CDO-Urgent"]
+          - !If [SilenceAlertsCondition, !Ref AWS::NoValue, !If [IsDemoCondition, !Sub "arn:aws:sns:${AWS::Region}:${AWS::AccountId}:javabuilder-demo-high-concurrency", !Sub "arn:aws:sns:${AWS::Region}:${AWS::AccountId}:CDO-Urgent"]]
```
cat5inthecradle, line 869: To eliminate `IsDemoCondition`:

**Easy**: Parameterize the `ElevatedSevereErrorRate` SNS topic name and pass it in as a new config option (then follow with one of these better options)

**Medium**: Parameterize all of the SNS topic names, for consistency

**Hard**: Create the SNS topics within this template so you can `!Ref` them directly. You can conditionally create them based on the `SilenceAlertsCondition` and keep the `!If` for that here. Then, we could conditionally _also_ notify CDO-Urgent for some alarms in production (I know, I know, "I thought you wanted to get rid of environment checks?" - I think an IsProd check for the purpose of escalating alerts is fine, and I think this would be a cleaner way to do it)



_*SNS is about 50 cents per 1 million requests, after the first free million. I don't expect this alarm to contribute in any significance to that number, so sending to both the per-stack topic and cdo-urgent isn't a major concern._

cicd/3-app/javabuilder/template.yml.erb
```diff
@@ -857,18 +858,19 @@ Resources:
...
       EvaluationPeriods: 10
       DatapointsToAlarm: 10
       Period: 60
-      Threshold: 400
+      Threshold: !If [IsDemoCondition, 45, 400]
```
cat5inthecradle, line 873: To eliminate `IsDemoCondition`:

Parameterize `HighConcurrentExecutionsAlarmThreshold`.

cat5inthecradle (changes requested):
You can't just name a PR "Final Updates" and think it's going to go through smoothly :rofl: 

Numbers look good from my perspective, and always easy to tweak. I have some feedback on the use of conditions instead of parameters for handling different environments. We should prefer parameters wherever possible. I think the desire to use conditionals here is partly because we have to work around the tech debt of manually-provisioned SNS topics.

EDIT: to be clear, my requested change is to get rid of `IsDemoCondition` and not _add_ any environment-detecting logic to the template in this PR.

cicd/3-app/javabuilder/template.yml.erb
cat5inthecradle: PR to eliminate `IsDevCondition`, definitely a non-school-year project ;)
https://github.com/code-dot-org/javabuilder/pull/394

molly-moen:
> You can't just name a PR "Final Updates" and think it's going to go through smoothly 🤣

😆 I renamed it to avoid cursing myself further

cicd/3-app/javabuilder/config/dev.config.json
```diff
@@ -6,7 +6,9 @@
     "ReservedConcurrentExecutions": "3",
     "LimitPerHour": "50",
     "LimitPerDay": "150",
-    "SilenceAlerts": "true"
+    "SilenceAlerts": "false",
```
molly-moen, line 9: this is a temporary change so I can test the config

cicd/3-app/javabuilder/template.yml.erb
molly-moen: Thanks! I started a [slack thread](https://codedotorg.slack.com/archives/C03DBDN67B7/p1684949225430359) to figure out the appropriate follow-up tasks here. For now I'm going with the "easy" option, updated and verifying with my dev instance now.

molly-moen: tested that things work as expected, so this is ready for review again!

cat5inthecradle (approved):

PR #400, opened 2023-10-31
CICD Improvements

cat5inthecradle:


cat5inthecradle:
I've manually deployed the dependencies stack to create the new role, but before testing the rest, I'd like to get a review.

cicd/1-setup/cicd-dependencies.template.yml
```diff
@@ -73,6 +73,47 @@ Resources:
...
+      Path: /service-role/
+      PermissionsBoundary: !ImportValue IAM-DevPermissions
+      Policies:
+        # - PolicyName: CloudformationPackage
+        #   PolicyDocument:
+        #     Statement:
+        #       - Effect: Allow
+        #         Action:
+        #           - TBD
+        #         Resource: TBD
```
cat5inthecradle, line 98: We might need this permission

unlox775-code-dot-org (approved):
This looks great.  I spent some time digging into how we were using `IAM-DevPermissions` as a Permission boundary.  The part that didn't sit well with me was using a automated system that in any way referenced the role used by human users.  As a permission boundary, I don't think it harms the security here.  And everything else looks good, so I'm happing shipping this PR as-is

In future I am thinking I would prefer to make sandbox-style allow list permission boundaries.  The main thing I would propose, at least to start the discussion, is to move away from block-list style rulesets, that is using negative rule logic, `Deny`, or `NotAction` or `StringNotEquals`.

Though not specifically for this PR, I put together a POC to test some of my thoughts on blocklist style access.  See here: https://github.com/unlox775/security_poc/tree/main/passrole_star

PR #412, opened 2025-11-18
Prevent Class Name Collision

fisher-alice:
Java Lab update - see jira ticket for details and before/after updates.
Related PR: https://github.com/code-dot-org/code-dot-org/pull/69703

- jira: https://codedotorg.atlassian.net/browse/AFL-207


sanchitmalhotra126 (approved):
Nice!

bencodeorg (approved):
Nice! A couple follow-up questions:

- Will we also update our frontend filename validation to match this? I think Sanchit suggested this, seems like a good idea.
- Also curious about existing projects with file names with a period in them. I guess they will hit an error, but it seems like we have a reasonably helpful error message and they can change the file names?

fisher-alice:
> Nice! A couple follow-up questions:


> * Also curious about existing projects with file names with a period in them. I guess they will hit an error, but it seems like we have a reasonably helpful error message and they can change the file names?

Currently on `prod` if you create a java file with a period within filename, there is already an error. So I don't think this will impact users who have java lab projects with periods in file names since they're already receiving an error.

Screenshot 2025-11-25 at 8 18 52 AM

Screenshot 2025-11-25 at 8 00 49 AM

Screenshot 2025-11-25 at 8 00 24 AM

> 
> * Will we also update our frontend filename validation to match this? I think Sanchit suggested this, seems like a good idea.

Yes, we can do this! https://github.com/code-dot-org/code-dot-org/pull/69772

PR #414, opened 2026-06-12
Fix bug with session id name

molly-moen:
I noticed today that Javabuilder was frequently failing to connect, with the websocket request returning a 502. I (with help from Claude) tracked the error down to `StartSessionsAndRelayMessagesFunction`, which was logging
`"errorMessage": "The name of a FIFO queue can only include alphanumeric characters, hyphens, or underscores, must end with .fifo suffix and be 1 to 80 in length.",`. This error pointed to how we were naming our SQS queues. This hasn't changed on our end, but we make an assumption that the websocket connection ids we use to generate the queue names end with a single `=`. In the case of the failures I found, the ids ended with 2 equals signs. The equals signs aren't valid in FIFO queue names, and therefore we were failing to make the queues.

It seems like this is a recent change in how websocket connection ids are generated. I've been using Javabuilder a lot lately and only noticed this happening frequently today. The last week of logs confirm this:
Screenshot 2026-06-12 at 10 47 09 AM

To fix this, instead of only stripping the last `=` from the connection id, we instead will strip all `=`. 

## Testing
I tried to deploy a dev javabuilder instance, but the permissions are broken right now. This change seems pretty safe and I added unit tests for it, so I feel ok about pushing it forward without full testing.

### DIFF
--- a/api-gateway-routes/session_id_helper.rb
+++ b/api-gateway-routes/session_id_helper.rb
@@ -6,6 +6,8 @@ module SessionIdHelper
   # The connection ID is base64-encoded, so it can end with one or more '='
   # padding characters. We remove all of them here.
   def get_session_id(event)
+    stripped_connection_id = event["requestContext"]["connectionId"].sub(/=+$/, "")
+    puts "session_id_helper stripped connection id: #{stripped_connection_id}, original connection id: #{event["requestContext"]["connectionId"]}"
     event["requestContext"]["connectionId"].sub(/=+$/, "")
   end
 end
--- a/api-gateway-routes/session_id_helper.rb
+++ b/api-gateway-routes/session_id_helper.rb
@@ -1,0 +1,11 @@
+module SessionIdHelper
+  # The session ID is the connection ID with any trailing '=' removed, so it can
+  # be used as an SQS queue name. SQS queues can only be named with the following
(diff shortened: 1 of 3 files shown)


bencodeorg (approved):
Whoa, nice find! So AWS generates the `connectionId`? And we think that changed somehow?

molly-moen:
> So AWS generates the connectionId? And we think that changed somehow?

Yep!

api-gateway-routes/session_id_helper.rb
```diff
@@ -0,0 +1,11 @@
+module SessionIdHelper
+  # The session ID is the connection ID with any trailing '=' removed, so it can
+  # be used as an SQS queue name. SQS queues can only be named with the following
+  # characters: alphanumeric characters, hyphens (-), and underscores (_).
+  # See https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/SQS/Client.html#create_queue-instance_method
+  # The connection ID is base64-encoded, so it can end with one or more '='
+  # padding characters. We remove all of them here.
+  def get_session_id(event)
+    event["requestContext"]["connectionId"].sub(/=+$/, "")
```
stephenliang, line 9: You may want to consider 1-way hashing the connection id with something like sha256 to ensure alphanumeric to future proof this.

